PR #4, opened 2020-04-15
done with Q processing code

jmfranck:


guinneak198:
I am pretty sure we are okay with what is there and this is old

PR #11, opened 2020-05-26
Folder org

guinneak198:
creating folders for first, and second level


jmfranck:
You seem to have created two copies of the file, rather than using `git mv`.  Normally, we would be able to apply the changes onto code that has been changed in the meantime, but because it wasn't done correctly, we can't do that.  I would just create a new folder to do the reorganization.  Also -- as a convention, don't use capitals in the file names, though continue to use underscores to separate words

PR #12, opened 2020-06-02
shouldn't have these quit statements -- remove and check that it works

jmfranck:


jmfranck:
changes made elsewhere

PR #13, opened 2020-06-02
add correlation alignment as a method and try it out on ODNP data as an example

jmfranck:


align_slice.py
```diff
@@ -28,3 +28,27 @@ def align_and_slice(s, dimension='t2', convwidth=500, threshold=0.05, fl=None):
     if fl is not None:
         fl.pop_marker()
     return s[dimension:slices]
+def correlation_align(s,avg,convwidth=0,dimension='t2',fl=None):
+    assert not s.get_ft_prop(dimension), 'I want time-domain data'
+    assert not avg.get_ft_prop(dimension), 'I want time-domain data'
```
guinneak198, line 33: why are we doing this twice? one with avg and one with s?

align_slice.py
```diff
@@ -28,3 +28,27 @@ def align_and_slice(s, dimension='t2', convwidth=500, threshold=0.05, fl=None):
...
+    avg.run(conj)
+    R = pi*convwidth # Lorentzian FWHM formula -- R = 1/T2, FWHM = R/pi
+    # ALSO convolve both
+    check = avg * s * exp(-2*R*abs(s.fromaxis(dimension)))
+    assert s.get_ft_prop(dimension,['start','freq']) is not None, "should be FT'd first, so the startpoint is set"
+    check.ft('t2', pad=2**14)
+    check.reorder('t2',first=False)
+    forplot = abs(check)['t2':(-50,50)]
+    myline = forplot.C.argmax('t2')
```
guinneak198, line 42: didn't you say we don't want to use C?

align_slice.py
```diff
@@ -28,3 +28,27 @@ def align_and_slice(s, dimension='t2', convwidth=500, threshold=0.05, fl=None):
...
+    forplot.smoosh(['ph1','ph2','power'],'indirect'
```
guinneak198, line 43: so because one of these dimensions is power I am assuming this funciton is only applicable to ODNP experiments?

align_slice.py
```diff
@@ -28,3 +28,27 @@ def align_and_slice(s, dimension='t2', convwidth=500, threshold=0.05, fl=None):
     if fl is not None:
         fl.pop_marker()
     return s[dimension:slices]
+def correlation_align(s,avg,convwidth=0,dimension='t2',fl=None):
```
guinneak198, line 31: this seems to just be an averaged apodization...I am certain I'm wrong but if not apodization then what?

align_slice.py
jmfranck: yes, these are two different objects

align_slice.py
jmfranck: yes, but if you don't use C here, it will replace forplot with the result of the call to argmax -- I still want forplot

align_slice.py
jmfranck: actually, no, that's not true ... I will edit this in response

align_slice.py
jmfranck: I don't know what you mean

proc_scripts/align_slice.py
```diff
@@ -28,27 +28,32 @@ def align_and_slice(s, dimension='t2', convwidth=500, threshold=0.05, fl=None):
...
         fl.next('cross-correlation')
-        fl.image(forplot,human_units=True)
-        fl.plot(myline.C.smoosh(['ph1','ph2','power'],'indirect'
+        fl.image(forplot)
+        fl.plot(myline.smoosh(phcyc_dims+indirect_dims,'indirect'
             ).setaxis('indirect','#'
                     ).reorder('indirect',first=False)
                 , 'w', linewidth=3, alpha=0.5)
-    return s*exp(-1j*2*pi*myline*s.fromaxis('t2'))
+    return s*exp(-1j*2*pi*myline*s.fromaxis(axis))
```
jmfranck: these edits are in response to [this comment](https://github.com/jmfrancklab/proc_scripts/pull/13#discussion_r434751630)

proc_scripts/align_slice.py
```diff
@@ -28,3 +28,27 @@ def align_and_slice(s, dimension='t2', convwidth=500, threshold=0.05, fl=None):
     if fl is not None:
         fl.pop_marker()
     return s[dimension:slices]
```
jmfranck: (not pertaining to these particular lines)

note that there is a conflict below.  once these conversations are resolved, we can rebase this branch onto the most recent copy of master, which I think will make the conflicts go away

align_slice.py
guinneak198: avg is the average along the coherence channels? mean_all_but('t2') excludes t2 from the averaging?

guinneak198: also, we want or do not want the dimensions of avg and s to be the same?

align_slice.py
guinneak198: I see, this makes sense then

align_slice.py
guinneak198: for clarification purposes I guess I am asking what the purpose of this function is? what does it add to the overall processing?

proc_scripts/align_slice.py
guinneak198: okay I will wait for you to resolve then!

align_slice.py
jmfranck: I don't remember offhand, but I think their ndshapes are different b/c the second is an average.  Here "dimension" is the name of the dimension along which the correlation alignment is performed -- honestly, I should have called it "axis" instead, so I just pushed a commit where I renamed it.  See [this page](https://en.wikipedia.org/wiki/Cross-correlation) if you want a sense of what this is doing

align_slice.py
jmfranck: Since the frequency is drifting, we want to frequency shift each transient slightly so that it's lined up with the average peak.

Inspired by the recent Keller + Maly paper, I try to do this using the cross-correlation -- specifically, I find what frequency the cross-correlation is maximized at, which tells me how much I need to shift the peak over to get it aligned.

See [this page](https://en.wikipedia.org/wiki/Cross-correlation) for more info on the cross-correlation.

jmfranck: actually better, please do give [this page](https://www.mathworks.com/help/signal/ug/align-signals-using-cross-correlation.html) a careful look, as well, and see if you agree with the way that this is done.

guinneak198: makes much more sense line by line! will talk in indiv meeting at 12 on this!

proc_scripts/align_slice.py
jmfranck: I'm going to wait for your go-ahead.  I need to close this pull request (which will kill all the comments) before I can rebase.

jmfranck: (i.e. is there anything in these comments that you would like addressed that is not currently addressed?)

guinneak198: no, the only remaining issue is the enhancement curve spoof. I have fixed the reordering and so the only issue is the enhancement curve which i believe is due to it doing an enhancement curve for every run rather than the average.

guinneak198:
This works and runs, good to merge

jmfranck:
There are conflicts.  After `git checkout dnp_alignment`, run `git merge master --no-ff` and resolve the conflicts, then test that everything still works the way that it should.

Also, after the group meeting, I was under the impression that you need to read through the code in detail to figure out whether or not it's working.  In particular, it would seem like you also want to write a simple test code that applies the alignment code twice, and which shows that the second time around, the white line runs straight across (indicating that the center of the peaks are aligned) rather than bouncing around.

guinneak198:
Just pushed after fixing conflicts and testing each code. Works on my end

proc_scripts/plotting.py
```diff
@@ -48,12 +48,13 @@ def side_by_side(self,plotname,s,thisrange):
...
         sca(ax1)
+        #self.set_units('t2',None)
         forplot = s['t2':expand_limits(thisrange,s)]
-        self.image(forplot)
+        self.image(forplot,human_units=False)
         print("drawing limits",thisrange)
         draw_limits(thisrange,forplot)
         sca(ax2)
-        self.image(forplot.C.cropped_log())
+        self.image(forplot.C.cropped_log(),human_units=False)
```
jmfranck, line 57: why did you add `human_units=False` here?

proc_IR.py
```diff
@@ -55,13 +57,12 @@
...
     s.ft('t2')
-    s.convolve('t2',10)
-    fl.image(s)
+    fl.image(s.C.convolve('t2',10))
     #}}}
     #{{{select t2 axis range and 
     s.ift('t2')
     s = s['t2':(0,None)]
-    s *= -1
+    #s *= -1
```
jmfranck, line 65: what's going on here?

proc_IR.py
```diff
@@ -16,12 +16,14 @@
...
 for searchstr,exp_type,nodename, postproc in [
-        ('200303_IR_AER_6','test_equip','signal','spincore_IR'),
+        ('200212_IR_3_30dBm', 'test_equip', 'signal', 
+            'spincore_IR'),
         ]:
     s = find_file(searchstr, exp_type=exp_type,
             expno=nodename,
             postproc=postproc, lookup=postproc_dict,
             dimname='indirect')
+    s *= exp(-1j*s.fromaxis('indirect')*clock_correction)
```
jmfranck, line 26: clock correction should be part of the list of tuples above

proc_scripts/plotting.py
guinneak198: will not plot if not there.  was a quick fix

proc_IR.py
guinneak198: bologna got rid of it

proc_IR.py
guinneak198: we would have to adjust the find_file function as clock_correction is not an argument of it currently.

proc_CPMG.py
```diff
@@ -79,24 +69,27 @@ def print_fun(x, f, accepted):
     fl.image(s.imag)
     #}}}
     #{{{select echo decay fit function
-    data = s['t2':(-200,200)].sum('t2')
+    s.ift('t2')
+    s = s['t2':(0,None)]
+    #s *= -1
+    s['t2',0] *= 0.5
```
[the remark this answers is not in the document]
guinneak198: incorporated in FID slice function

proc_scripts/plotting.py
jmfranck: ok, yes, here this makes sense

proc_CPMG.py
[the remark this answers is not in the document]
jmfranck: to be more specific:

1. why are you taking an FID slice inside the CPMG?
2. if you are taking an FID slice, why not use the FID slicing function?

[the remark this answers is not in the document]
guinneak198: I misunderstood what you were asking.
1. the function of FID slice is slightly different than what is done here, you can go back to the original Slice_FID_from_echo function to see difference, this is I believe just slicing out the portion of the time domain that is applicable.
2. So my previous misunderstanding was that you wanted the function to be implemented in the proc_CPMG and so that is what is there in my latest push

proc_IR.py
guinneak198: would you rather keep it as is and have clock_correction as a variable we change upon processing or incorporate it into pyspecdata?

proc_CPMG.py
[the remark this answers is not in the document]
guinneak198: This has been changed I believe according to PR#14

proc_IR.py
jmfranck: that doesn't make sense -- I suggest a fix below in 2cbc803971b64e87881

proc_CPMG.py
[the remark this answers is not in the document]
jmfranck: I diff against `alec_fitdata_rebased` and neither addresses my question:

> Why are we slicing out half of the echo?

More specifically (I said this in [a conversation on slack with Alec](https://jmfrancklab.slack.com/archives/CKRNAS3PD/p1592923328010100?thread_ts=1592922342.007900), so it might be good to go look at that and then touch base w/ him), why not (1) apply `hermitian_function_test` to find the center of the echo then (2) run `register_axis({'t2':0})` and (3) slice out the echo as large as you can, but still symmetrically about 0 (4) double-double check that the `t2` axis is symmetric about zero (`s.getaxis('t')[0]` is the same as `...[-1]`) (5) when you FT along `t2` the signal should now be entirely red.

proc_IR.py
```diff
@@ -7,17 +7,15 @@
...
 coh_err = {'ph1':1,# coherence channels to use for error
         'ph2':r_[0,2,3]}
 # }}}
 
-for searchstr,exp_type,nodename, postproc in [
+for searchstr,exp_type,nodename, postproc, clock_correction, filter_bandwidth in [
         ('200212_IR_3_30dBm', 'test_equip', 'signal', 
-            'spincore_IR'),
+            'spincore_IR', 1.785, 5e3),
         ]:
```
jmfranck: please check this out, so you can see what I was suggesting

proc_IR.py
guinneak198: my bad I misunderstood! Got it now though!

proc_CPMG.py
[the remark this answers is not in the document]
guinneak198: So I have incorporated the hermitian function test and register axis as you said. It looks the same (tE axis extends from 0 to about 2.8). Its also blue when ft'd but please take a look, something shifts it in the frequency domain but idk what....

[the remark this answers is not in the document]
guinneak198: I have incorporated these in, its actually blue when ft'ed and shifted which is weird...

proc_CPMG.py
```diff
@@ -28,20 +28,10 @@
...
-    fl.next('time domain after hermitian test')
-    fl.image(s)
-    s.register_axis({'t2':0})
-    s = s['t2':(-15,15)]
-    print(s.getaxis('t2')[0])
-    print(s.getaxis('t2')[-1])
-    fl.next('centered')
-    s.ft('t2')
-    fl.image(s,human_units=False)
+    raise ValueError("after merging PR#13, look at the code to center the echo in proc_DOSY_CPMG.py from the DOSY_T2 branch to center an echo, move it to a second-level function, and reuse it here.")
```
jmfranck, line 31: This will address #20

jmfranck: so leave it as an error for this PR

proc_CPMG.py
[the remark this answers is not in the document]
jmfranck: I'm closing this for now, mentioned in an issue and not really part of this pull request.

jmfranck:
it's my understanding that we're still not doing what's described in the second paragraph of [this comment](https://github.com/jmfrancklab/proc_scripts/pull/13#issuecomment-644160207) --> once you have a working demo of that, we can go ahead and merge this PR

proc_scripts/align_slice.py
```diff
@@ -27,4 +27,33 @@ def align_and_slice(s, dimension='t2', convwidth=500, threshold=0.05, fl=None):
...
+    indirect_dims = set(s.dimlabels) - set([axis])
+    phcyc_dims = [j for j in indirect_dims if j.startswith('ph')]
+    phcyc_dims.sort()
+    indirect_dims = list(set(indirect_dims) - set(phcyc_dims))
+    #forplot.reorder(phcyc_dims+indirect_dims)
+    forplot.smoosh(phcyc_dims+indirect_dims,dimname='indirect',noaxis=True).setaxis('indirect','#').reorder('indirect',first=False)
+    if fl is not None:
+        fl.push_marker()
+        fl.next('cross-correlation')
+        fl.image(forplot,human_units=False)
```
guinneak198, line 52: running into issues here, 7/8/20

guinneak198: okay so I did some comparisons with previous versions on the git tree that I know produced the plot at least. proc_echo_mw now AT LEAST can produce the plot and we can begin troubleshooting to see what's wrong with it/what adjustments need to be made. Take a look and let me know any thoughts or concerns you may have!

proc_scripts/align_slice.py
```diff
@@ -1,4 +1,5 @@
 from pyspecdata import *
+import numpy as np
```
guinneak198, line 2: question: why don't we use the correlate function of either scipy or numpy?

proc_scripts/align_slice.py
```diff
@@ -35,22 +36,23 @@ def correlation_align(s,avg,convwidth=0,axis='t2',fl=None):
     R = pi*convwidth # Lorentzian FWHM formula -- R = 1/T2, FWHM = R/pi
     # ALSO convolve both
     check = avg * s * exp(-2*R*abs(s.fromaxis(axis)))
+    print(check)
     assert s.get_ft_prop(axis,['start','freq']) is not None, "should be FT'd first, so the startpoint is set"
-    check.ft(axis, pad=2**14)
-    check.reorder(axis,first=False)
-    forplot = abs(check)[axis:(-50,50)]
-    myline = forplot.C.argmax(axis)
+    #check.ft(axis, pad=2**14)
```
guinneak198, line 41: why are we fting the check?

proc_scripts/align_slice.py
```diff
@@ -35,22 +36,23 @@ def correlation_align(s,avg,convwidth=0,axis='t2',fl=None):
...
+    #check.reorder(axis,first=False)
+    #forplot = abs(check)[axis:(-50,50)]
```
guinneak198, line 43: why this range on axis? what dictates this?

proc_scripts/align_slice.py
jmfranck: so, in group meeting, why don't you present the two alignment methods -- `align_slice` and the correlation alignment, and explain the math/science ideas behind them.  As I'm saying elsewhere, I think the main bottleneck here is understanding/ideas-based, not a problem with the coding.

proc_scripts/align_slice.py
jmfranck: because then you would need to figure out the axis

proc_scripts/align_slice.py
jmfranck: the typical linewidth on our instrument

proc_scripts/align_slice.py
guinneak198: so don't we adjust the axis anyhow? why would this be a problem?

proc_scripts/align_slice.py
jmfranck: I think to determine the correlation.

proc_scripts/align_slice.py
jmfranck: If we do this with pyspecdata, the offset between the peaks is given as a frequency in Hz, rather than an index number we would have to convert. Also, there is no sense to using one of those functions, since they would only replace some 2 lines of code.

jmfranck:
I think this is obsolete.  If it's not, then master needs to be merged into the relevant branch before reopening.

Issue #20, opened 2020-06-23
I diff against `alec_fitdata_rebased` and neither addresses my question:


jmfranck:
I diff against `alec_fitdata_rebased` and neither addresses my question:

> Why are we slicing out half of the echo?

More specifically (I said this in [a conversation on slack with Alec](https://jmfrancklab.slack.com/archives/CKRNAS3PD/p1592923328010100?thread_ts=1592922342.007900), so it might be good to go look at that and then touch base w/ him), why not (1) apply `hermitian_function_test` to find the center of the echo then (2) run `register_axis({'t2':0})` and (3) slice out the echo as large as you can, but still symmetrically about 0 (4) double-double check that the `t2` axis is symmetric about zero (`s.getaxis('t')[0]` is the same as `...[-1]`) (5) when you FT along `t2` the signal should now be entirely red.

_Originally posted by @jmfranck in https://github.com/jmfrancklab/proc_scripts/pull/13_

guinneak198:
I have just pushed a version of proc_CPMG.py that incorporates the hermitian function test and register_axis. when FT'd it is entirely blue however. Not certain this is what you want but take a look

jmfranck:
so we clarified that this is on `alec_fitdata_rebased`

guinneak198:
Is this resolved with the new center_CPMG_echo function?/ can this be closed?

PR #21, opened 2020-06-24
processing DOSY-CPMG

aabeaton:


proc_DOSY_CPMG.py
```diff
@@ -74,10 +74,7 @@
...
     fl.next('request 5 (after subtracting best shift)')
     s.register_axis({'t2':0})
-    #s /= zeroth_order_ph(s['t2',0], fl=fl)
-    ph0 = s['t2',0].C.mean('t2')
-    ph0 /= abs(ph0)
-    s /= ph0
+    s /= zeroth_order_ph(s['t2':0])
     fl.image(s)
     time_bound = min(abs(s.getaxis('t2')[r_[0,-1]]))
     s = s['t2':(-time_bound,time_bound)]
```
jmfranck: this section of the code here and surrounding is what will solve #20 (note for @guinneak198 ) -- please also see the other mentions of #20 : specifically where I talk about extracting this as a second order function

jmfranck: @guinneak198 : I think that what will solve #20 is for you to make this into a second order function (it uses zeroth order), so that you can use it in other places as well --> something like center_CPMG_echo()

PR #22, opened 2020-06-25
update hermitian phasing to use zeroth order phasing

jmfranck:


jmfranck:
Please check out the changes, make sure it works with all your scripts, and let me know if there are any questions!

guinneak198:
It seems it runs smoothly! QUESTION: does this mean we no longer have to do a separate zeroth order phase correction outside of the hermitian function or we still need that?

PR #26, opened 2020-08-11
Proc 90 pulse

guinneak198:
processing for simple 90 pulse bruker experiment. will be used to compare baseline methods

### DIFF
--- a/proc_IR.py
+++ b/proc_IR.py
@@ -8,38 +8,36 @@ fl = fl_mod()
 t2 = symbols('t2')
 logger = init_logging("info")
 # {{{ input parameters
-clock_correction = 1.785
 filter_bandwidth = 5e3
 coh_sel = {'ph1':0,
-        'ph2':0}
+        'ph2':-1}
 coh_err = {'ph1':1,# coherence channels to use for error
         'ph2':r_[0,2,3]}
 # }}}
 
-for searchstr,exp_type,nodename, postproc in [
-        ('w8_200731', 'test_equip', 2, 
-            'ag_IR2H'),
-        #('w8_200309', 'test_equip',2,'ag_IR2H'),
+for searchstr,exp_type,nodename, postproc, clock_correction in [
+        #('w8_200731', 'test_equip', 2, 
+        #    'ag_IR2H'),
+        ('w8_200309', 'test_equip',2,'ag_IR2H',1.785),
         ]:
     fl.basename = searchstr
-    if postproc=='ag_IR2H':
-        s = find_file(searchstr,exp_type=exp_type,
-                expno=nodename,
-                postproc=postproc, lookup=postproc_dict,
-                dimname='indirect')
-    if postproc=='spincore_IR_v1':    
+    if clock_correction is None:
         s = find_file(searchstr, exp_type=exp_type,
             expno=nodename,
             postproc=postproc, lookup=postproc_dict,
-            clock_correction=clock_correction, dimname='indirect')
+            dimname='indirect')
+    else:
+        s = find_file(searchstr, exp_type=exp_type,
+            expno=nodename,
+            postproc=postproc, lookup=postproc_dict,
+            dimname='indirect',
+            clock_correction=clock_correction)
     #{{{filter data
-    if postproc=='spincore_IR_v1':
-        s = s['t2':(-filter_bandwidth/2,filter_bandwidth/2)]
+    s = s['t2':(-filter_bandwidth/2,filter_bandwidth/2)]
     #}}}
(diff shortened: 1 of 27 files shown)


guinneak198:
This pull request should be ready for merging. the baseline methodology is a little wonky but we are able to load and preprocess the 90 pulse data

proc_IR.py
```diff
@@ -11,27 +11,35 @@
 clock_correction = 1.785
 filter_bandwidth = 5e3
 coh_sel = {'ph1':0,
-        'ph2':1}
+        'ph2':0}
```
jmfranck, line 14: this doesn't make sense -- shouldn't `ph2` be the 90 pulse and actually have Δp=-1?

proc_IR.py
```diff
@@ -11,27 +11,35 @@
...
+    if postproc=='ag_IR2H':
+        s = find_file(searchstr,exp_type=exp_type,
+                expno=nodename,
+                postproc=postproc, lookup=postproc_dict,
+                dimname='indirect')
+    if postproc=='spincore_IR_v1':    
+        s = find_file(searchstr, exp_type=exp_type,
             expno=nodename,
             postproc=postproc, lookup=postproc_dict,
-            dimname='indirect')
```
jmfranck, line 27: this conditional should be handled by `find_file`

jmfranck: (unless I'm missing something)

proc_IR.py
```diff
@@ -11,27 +11,35 @@
...
-    s *= exp(-1j*s.fromaxis('indirect')*clock_correction)
+            clock_correction=clock_correction, dimname='indirect')
     #{{{filter data
-    s = s['t2':(-filter_bandwidth/2,filter_bandwidth/2)]
+    if postproc=='spincore_IR_v1':
+        s = s['t2':(-filter_bandwidth/2,filter_bandwidth/2)]
```
jmfranck, line 37: there should be no if statement here -- if you want a different filter bandwidth, just set it above.

proc_IR.py
```diff
@@ -11,27 +11,35 @@
...
     #}}}
     #{{{hermitian function test and apply best shift
     fl.next('frequency domain before')
     fl.image(s)
+    fl.show();quit()
```
jmfranck, line 42: no quit statements in the middle of the code!

jmfranck: I'm going to stop here -- this means it's not ready to go

proc_IR.py
guinneak198: the filter bandwidth only applies to IR acquired via spincore. It does not apply to bruker IR.

proc_IR.py
guinneak198: REMEMBER this branch is for proc_90.py. Clearly the other scripts are outdated. but this branch was created I believe back in march. So it does not have proc_T1CPMG and as we noted proc_IR was really bad as well. I don't want to necessarily waste your time with having you go through the other outdated scripts. This is just for proc_90.py and the load data that accompanies it.

jmfranck: But it doesn't work like that.  The changes shown here are the changes you are requesting.  Maybe you need to merge, rebase, or something like that.  Probably easiest to talk about this over zoom?

jmfranck: When you ask for a pull request to get reviewed, you should look at the "files changed" page, to be sure everything is in order.

proc_IR.py
jmfranck: You can define a filter bandwidth for either, and it should have the same behavior for both.

proc_IR.py
guinneak198: So what is the quickest way to update all files and add in the proc_90.py?

guinneak198: So I merged in bruker processing as this is the most up to date for the rest of the processing scripts. Hopefully this will fix the issue?

jmfranck: we'll talk about this today.  I'm guessing that you probably wanted to merge in master, not bruker processing.  Typically, it's a good idea to merge in the branch that you are trying to merge into when you think a pull request is ready.  This will both bring the code up to date and flag any conflicts, etc.  I think just leave this alone and we'll discuss today.

proc_IR.py
guinneak198: should be fixed

proc_IR.py
guinneak198: Is this taken care of?

jmfranck: we're not touching this pull request until #27 is done

jmfranck:
this has a bunch of unrelated changes -- closing out to deal with one by one

PR #27, opened 2020-09-08
Bruker processing

guinneak198:
This is good for processing T1CPMG and can be merged with your approval!


proc_CPMG.py
```diff
@@ -5,25 +5,38 @@
...
-    s =  find_file(searchstr, exp_type=exp_type,
+    if postproc=='spincore_CPMG_v1':
+        s =  find_file(searchstr, exp_type=exp_type,
             expno=nodename, postproc=postproc, lookup=postproc_dict)
-    #{{{ centering CPMG echo
+    if postproc=='ag_CPMG_strob':
+        s = find_file(searchstr, exp_type=exp_type,
+                expno=nodename, postproc=postproc,
+                lookup=postproc_dict,fl=fl)
+    #fl.show();quit()
```
jmfranck, line 30: no commented code

proc_IR.py
```diff
@@ -11,27 +11,35 @@
 clock_correction = 1.785
 filter_bandwidth = 5e3
 coh_sel = {'ph1':0,
-        'ph2':1}
+        'ph2':0}
```
jmfranck, line 14: doesn't make sense

jmfranck: shouldn't it be Δp=-1??

proc_IR.py
```diff
@@ -11,27 +11,35 @@
...
+        s = find_file(searchstr, exp_type=exp_type,
             expno=nodename,
             postproc=postproc, lookup=postproc_dict,
-            dimname='indirect')
-    s *= exp(-1j*s.fromaxis('indirect')*clock_correction)
+            clock_correction=clock_correction, dimname='indirect')
     #{{{filter data
-    s = s['t2':(-filter_bandwidth/2,filter_bandwidth/2)]
+    if postproc=='spincore_IR_v1':
+        s = s['t2':(-filter_bandwidth/2,filter_bandwidth/2)]
```
jmfranck, line 37: I also commented on this in proc 90, so maybe go look there?

proc_IR.py
```diff
@@ -11,27 +11,35 @@
...
+                expno=nodename,
+                postproc=postproc, lookup=postproc_dict,
+                dimname='indirect')
+    if postproc=='spincore_IR_v1':    
+        s = find_file(searchstr, exp_type=exp_type,
             expno=nodename,
             postproc=postproc, lookup=postproc_dict,
-            dimname='indirect')
-    s *= exp(-1j*s.fromaxis('indirect')*clock_correction)
+            clock_correction=clock_correction, dimname='indirect')
```
jmfranck, line 34: so you've added the clock correction to your preprocessing, and this gets passed to the preprocessing function?

proc_IR.py
```diff
@@ -11,27 +11,35 @@
...
     #{{{filter data
-    s = s['t2':(-filter_bandwidth/2,filter_bandwidth/2)]
+    if postproc=='spincore_IR_v1':
+        s = s['t2':(-filter_bandwidth/2,filter_bandwidth/2)]
     #}}}
     #{{{hermitian function test and apply best shift
     fl.next('frequency domain before')
     fl.image(s)
+    fl.show();quit()
```
jmfranck, line 42: no quit statements -- stopping

proc_CPMG.py
```diff
@@ -5,25 +5,38 @@
 from proc_scripts.fitting import decay
```
jmfranck, line 5: there are multiple binary files checked in above -- remove them

proc_IR.py
guinneak198: So this guy has the if statements based on if they use the clock correction. If it is just ag_IR2h there is no need for an input of clock correction for the preprocessing however, if it is spincore_IR then the clock correction needs to be one of the kwarg for preprocessing.

proc_IR.py
guinneak198: So there was coherence selection in the preprocessing and this messed up a bit of things such as the hermitian function later in the code so I did some rearranging. There was some trouble running the fitting but I fixed and it works on my end thought the guess is completely wrong but aside from the guess it's good!

proc_IR.py
guinneak198: yes. This is in preprocessing for when applicable

jmfranck: cool

proc_IR.py
jmfranck: for an inversion recovery, the coherence selection should be ph1=0, ph2=-1.  Can't mere until that's fixed. This is a conceptual error, whether or not you coincidentally get the right result.

proc_IR.py
jmfranck: I'm confused what that has to do with the current line -- this is about the filter bandwidth, not the clock correction

guinneak198: fixed

proc_IR.py
guinneak198: I understand this is what we should see however, in the frequency domain prior to coherence selection this is the figure
image
Totally not sure if you are able to see that.... however in short the signal is in ph2=0 ph1=0... which is very odd

guinneak198:
IMPORTANT NOTE: this works for fitdata when pyspecdtaa is in the guess_function branch!!! not on master

jmfranck:
then the thing to do is to make sure the guess_function branch gets merged first!

proc_IR.py
jmfranck: this is bad data

proc_CPMG.py
```diff
@@ -4,26 +4,39 @@
...
+    if postproc=='spincore_CPMG_v1':
+        s =  find_file(searchstr, exp_type=exp_type,
             expno=nodename, postproc=postproc, lookup=postproc_dict)
-    #{{{ centering CPMG echo
-    s = center_CPMG_echo(s)
+    if postproc=='ag_CPMG_strob':
+        s = find_file(searchstr, exp_type=exp_type,
+                expno=nodename, postproc=postproc,
+                lookup=postproc_dict,fl=fl)
+        #{{{ centering CPMG echo
```
jmfranck, line 30: shouldn't have postprocessing-dependent if statements.  In fact, if I compare the two commands, they are the same (except one is plotting and the other is not).

proc_Hahn_echoph.py
```diff
@@ -22,7 +23,8 @@
     s.ift('t2')
     rough_center = abs(s).convolve('t2',0.01).mean_all_but('t2').argmax('t2').item()
     s.setaxis(t2-rough_center)
-    logger.info(strm(ndshape(s)))
+    logger.debug(strm(ndshape(s)))
+    quit()
```
jmfranck, line 27: AHHH!

proc_IR.py
```diff
@@ -11,21 +11,27 @@
 clock_correction = 1.785
 filter_bandwidth = 5e3
 coh_sel = {'ph1':0,
-        'ph2':1}
+        'ph2':-1}
```
jmfranck, line 14: yes, so this is good code

jmfranck:
check that everything still works for you!

guinneak198:
so set_guess is not a definition inside the fitdata class in master. Hence why it gives trouble in master and not in guess_function branch of pyspecdata. proc_IR.py needs this adjustment in order to run since there is a guess present. Other than this everything runs smoothly and it's ready to go

jmfranck:
OK! Let's go merge the guess_function branch on pyspecdata, then, and come back here, and check against master

jmfranck:
closing out so that we can deal with more manageably-sized pull request

PR #29, opened 2020-10-16
rewrite proc_square_refl from scratch

jmfranck:


guinneak198:
not sure if this should be deleted or not since in bruker_processing branch we have together integrated the third level function and I believe are happy with it.... thoughts?

guinneak198:
not sure if this should be deleted or not since in bruker_processing branch we have together integrated the third level function and I believe are happy with it.... thoughts?

jmfranck:
so, we need to compare the two files using [this](https://stackoverflow.com/questions/8131135/git-how-to-diff-two-different-files-in-different-branches) to make sure that they are similar.

so, I do

```
git checkout bruker_processing
git diff HEAD:./proc_scripts/third_level/analyze_square_refl.py square_pulse_rewrite:./proc_square_refl.py
```

I find that the two are quite different, and I'm not sure if the one in the bruker branch reflects the most recent code?

jmfranck:
OK, so after doing that, it does seem like the bruker branch is more recent, which is good.  But, before closing this pull request, you want to create a new pull request that makes the changes that demonstrate `./proc_scripts/third_level/analyze_square_refl.py` inside `examples/proc_square_refl.py`, and which removes the square pulse script from the main directory.

jmfranck:
Ideally, this should be included in the current code in some way -- if not, then merge master into this, and reopen.

PR #30, opened 2020-10-29
rescusitate hermitian phasing code that includes exponential decay

jmfranck:
this would need to be used to make changes to the existing hermitian routine


jmfranck:
this is obsolete

PR #33, opened 2021-01-08
updated clean version of proce_echo;prob with plotting

guinneak198:
updated version of proc_echo_mw for processing of ODNP data.

### DIFF
--- a/proc_echo_mw.py
+++ b/proc_echo_mw.py
@@ -14,9 +14,9 @@ t2 = symbols('t2')
 # to use: as a rule of thumb, make the white boxes
 # about 2x as far as it looks like they should be
 # leave this as a loop, so you can load multiple files
-for searchstr,exp_type,nodename,postproc,freq_range,max_t in [
+for searchstr,exp_type,nodename,postproc,freq_range,time_range in [
         ["201124_4AT100uM_DNP_cap_probe_1", 'ODNP_NMR_comp', 'signal',
-            'spincore_ODNP_v1', (-500,500), 0.065]
+            'spincore_ODNP_v1', (-500,500), (0,None)]
         ]:
     fl.basename = searchstr
     s = find_file(searchstr, exp_type=exp_type, expno=nodename,
@@ -31,7 +31,7 @@ for searchstr,exp_type,nodename,postproc,freq_range,max_t in [
     s = slice_FID_from_echo(s,max_t=max_t,fl=fl)    # visualize time domain after filtering and phasing
     #{{{apodizing and zero fill
     fl.next('apodize and zero fill')
-    R = 5.0/(max_t) # assume full decay by end time
+    R = 5.0/(time_range[-1]) # assume full decay by end time
     s *= exp(-s.fromaxis('t2')*R)
     s.ft('t2',pad=1024)
     fl.image(s.C.setaxis('power','#').set_units('power','scan #'))
--- a/proc_echo_mw.py
+++ b/proc_echo_mw.py
@@ -14,10 +14,11 @@ t2 = symbols('t2')
 # to use: as a rule of thumb, make the white boxes
 # about 2x as far as it looks like they should be
 # leave this as a loop, so you can load multiple files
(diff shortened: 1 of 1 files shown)


proc_echo_mw.py
```diff
@@ -14,10 +14,11 @@
 # to use: as a rule of thumb, make the white boxes
 # about 2x as far as it looks like they should be
 # leave this as a loop, so you can load multiple files
-for searchstr,exp_type,nodename,postproc,freq_range,time_range in [
-        ["200306_DNP_lg_probe_w34.*", 'test_equip', 'signal',
-            'spincore_ODNP_v1', (-300,300), (None,0.05)]
+for searchstr,exp_type,nodename,postproc,freq_range,max_t in [
+        ["201124_4AT100uM_DNP_cap_probe_1", 'ODNP_NMR_comp', 'signal',
+            'spincore_ODNP_v1', (-500,500), 0.065]
```
jmfranck, line 19: The previous parameters allow for the case where there's an echo that's not short time -- where there is noise between the start of acquisition and the rise of the echo.  Relatively inexpensive to permit this.

proc_echo_mw.py
```diff
@@ -27,16 +28,13 @@
...
     logger.debug(strm(ndshape(s)))
-    s = slice_FID_from_echo(s)['t2':(None,0.05)]
-    # visualize time domain after filtering and phasing 
-    fl.side_by_side('time domain (after filtering and phasing)\n$\\rightarrow$ use to adjust time range', s, time_range)
-    s =s['t2':time_range] # slices out time range along t2 axis
+    s = slice_FID_from_echo(s,max_t=max_t,fl=fl)    # visualize time domain after filtering and phasing
     #{{{apodizing and zero fill
     fl.next('apodize and zero fill')
-    R = 5.0/(time_range[-1]) # assume full decay by end time
+    R = 5.0/(max_t) # assume full decay by end time
```
jmfranck, line 34: both instances of max_t would just be time_range[-1], following on comment above.

jmfranck (changes requested):

proc_echo_mw.py
guinneak198: Okay so my understanding of this was to go back to using the t_range instead of max_t

jmfranck: yes, but above code doesn't show as outdated

proc_echo_mw.py
jmfranck: again, this should show as outdated

proc_echo_mw.py
guinneak198: sorry realized I hadn't pushed!

guinneak198:
Having issues with the load data file so am leaving this for now until that load data PR is merged

jmfranck (approved):

jmfranck:
So, based on your comment from this morning, this doesn't run yet, until we resolve the load data issues?

guinneak198:
correct specifically I get this error:
`Traceback (most recent call last):
  File "proc_echo_mw.py", line 22, in <module>
    s = find_file(searchstr, exp_type=exp_type, expno=nodename,
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\load_files\__init__.py", line 249, in find_file
    data = postproc_lookup[postproc_type](data,**kwargs)
  File "C:\Users\Alex Guinness\gitstuff\proc_scripts\proc_scripts\load_data.py", line 134, in proc_spincore_ODNP_v1
    0,dBm2power(array(s.get_prop('meter_powers'))+20)]
NameError: name 'array' is not defined`

jmfranck:
Hopefully if you look at the error again, you know what to do to fix it. If not, I guess let’s schedule a follow up

guinneak198:
so I know I need to put a np. in front but this is going to alter the load_data.py script so should be a part of the load data PR to correct this right?

jmfranck:
again -- obsolete? if not, then merge master into it and reopen

PR #34, opened 2021-01-11
Load data

guinneak198:
This is the updated version of the preprocessing that I have been using on bruker_processing branch. I realized this was outdated when trying to update the proc_echo_mw and there were issues in the load_data file. Once this is merged I can carry on with the proc_echo_mw

### DIFF
--- a/proc_scripts/load_data.py
+++ b/proc_scripts/load_data.py
@@ -297,7 +297,7 @@ def proc_nutation_amp(s,fl=None):
     s.setaxis('ph2',r_[0.,2.]/4)
     s.setaxis('ph1',r_[0.,4.]/4)
     s.set_units('t2','s')
-    s.set_units('amp','unitless')
+    s.set_units('amp', None)
     s.ft(['ph2','ph1'])
     if fl is not None:
         fl.next('after phase cycle FT')
--- a/proc_scripts/load_data.py
+++ b/proc_scripts/load_data.py
@@ -1,10 +1,22 @@
+"""
+Preprocessing Routines
+======================
+
+Defines the preprocessing routines for various experiments.
+
+All preprocessing routines are defined such that they return data where t2 is
+Fourier Transformed to the frequency domain, and any phase cycling dimensions
+are Fourier Transformed into the coherence pathway domain.
+
+**No data is thrown out** in any of the routines, so that the raw data can
+always be reconstructed through IFT (and smooshing, if necessary).
+"""
 from pyspecdata import *
 from .Utility import dBm2power
 import os
 from sympy import symbols
 import logging
 import numpy as np
-fl=figlist_var()
 def proc_bruker_deut_IR_v1(s,fl=fl):
     logger.info(strm("preprocessing for IR acquired on bruker"))
     if fl is not None:
@@ -133,7 +145,6 @@ def proc_bruker_T1CPMG_v1(s,d12,fl=None):
     dwdel1 = s.get_prop('acq')['DE']*1e-6
     dwdel2 = (anavpt*0.05e-6)/2
     #d12 is read as 0 if taken from parameters bc its too small
-    d12 = d12     
     d11 = s.get_prop('acq')['D'][11]
     p90_s = s.get_prop('acq')['P'][1]*1e-6
     quad_pts = ndshape(s)['t2'] # note that we have not yet chunked t2
@@ -189,7 +200,6 @@ def proc_bruker_CPMG_v1(s,d12,tau_extra,fl=None):
     dwdel1 = s.get_prop('acq')['DE']*1e-6
     dwdel2 = (anavpt*0.05e-6)/2
     #d12 is read as 0 if taken from parameters bc its too small
-    d12 = d12     
     d11 = s.get_prop('acq')['D'][11]
     p90_s = s.get_prop('acq')['P'][1]*1e-6
     quad_pts = ndshape(s)['t2'] # note that we have not yet chunked t2
@@ -198,7 +208,6 @@ def proc_bruker_CPMG_v1(s,d12,tau_extra,fl=None):
     # {{{ these are hard-coded for the pulse sequence
     #     if we need to, we could pull these from the pulse sequence, as we do
     #     for anavpt above
-    tau_extra = tau_extra
     tau_pad_start = tau_extra-dwdel1-6e-6
     tau_pad_end = tau_extra-6e-6
     twice_tau = 2*p90_s + 5e-6 + tau_pad_start + 1e-6 + acq_time + tau_pad_end +1e-6
@@ -368,7 +377,6 @@ def proc_DOSY_CPMG(s,dwdel1,tau_extra):
     grad_list = array([float(j.group()) for j in re.finditer('([0-9.]+)',m.groups()[0])])
     m = re.search('([0-9.]+) G/mm', s.get_prop('gradient_calib'))
     grad_list *= float(m.groups()[0])*0.1
-    dwdel1 = dwdel1     
     # {{{ find anavpt without hard-setting
     m = re.search('"anavpt=([0-9]+)"',ppg)
     if m is None:
@@ -380,7 +388,6 @@ def proc_DOSY_CPMG(s,dwdel1,tau_extra):
     quadrature_points = TD/2
     num_points_per_echo = quadrature_points/l25
     acq_time = dwdel2*num_points_per_echo*2
-    tau_extra = tau_extra
     tau_pad = tau_extra-6e-6
     tau_pad_start = tau_extra-dwdel1-6e-6
     tau_pad_end = tau_extra-6e-6
--- a/proc_scripts/load_data.py
+++ b/proc_scripts/load_data.py
@@ -3,22 +3,62 @@ from .Utility import dBm2power
 import os
 from sympy import symbols
 import logging
-#to use type s = load_data("nameoffile")
-def proc_bruker_deut_IR_withecho_mancyc(s):
-    raise RuntimeError("this is where postprocessing would be implemented -- not implemented yet")
+import numpy as np
+fl=figlist_var()
+def proc_bruker_deut_IR_withecho_mancyc(s,fl=fl):
+    print("preprocessing for IR acquired on bruker")
+    if fl is not None:
+        fl.next('raw data')
+        fl.image(s.C.setaxis(
+'indirect','#').set_units('indirect','scan #'))
+    s.chunk('indirect',['ph2','ph1','indirect'],[4,2,-1]) #expands the indirect dimension into indirect, ph1, and ph2. 
+    #                                                      Inner most dimension is the inner most in the loop in pulse sequence, 
+    #                                                      is the one on the farthest right. Brackets with numbers are the number of 
+    #                                                      phase cycle steps in each one. The number of steps is unknown in 'indirect' 
+    #                                                      and is therefore -1.
+    s.setaxis('ph1',r_[0:2.]/4) #setting values of axis ph1 to line up
+    s.setaxis('ph2',r_[0:4.]/4) #setting values of axis ph1 to line up
+    s.setaxis('indirect', s.get_prop('vd'))
+    s.ft('t2',shift=True) #fourier transform
+    if fl is not None:
+        fl.next('IR prior to FTing ph')
+        fl.image(s.C.setaxis('indirect','#').set_units('indirect','scan #'))
+    s.ft(['ph1','ph2']) #fourier transforming from phase cycle dim to coherence dimension
+    s.reorder(['indirect','t2'], first=False)
+    if fl is not None:
+        s_forplot = s.C
+        fl.next('FT')
+        fl.image(s_forplot.C.setaxis('indirect','#').set_units('indirect','scan #'))
+    if fl is not None:    
+        fl.next('time domain (all $\\Delta p$)')
+        s_forplot.ift('t2')
+        fl.image(s_forplot.C.setaxis('indirect','#').set_units('indirect','scan #'))
+    if fl is not None:    
+        fl.next('frequency domain (all $\\Delta p$)')
+        s_forplot.ft('t2',pad=4096)
+        fl.image(s_forplot.C.setaxis('indirect','#').set_units('indirect','scan #'))
+    return s
 
 def proc_bruker_deut_IR_mancyc(s, fl=None):
-    logger.info("going through this")
-    s.chunk('indirect',['indirect','ph1','ph2'],[-1,4,2]) #expands the indirect dimension into indirect, ph1, and ph2. inner most dimension is the inner most in the loop in pulse sequence, is the one on the farthest right. Brackets with numbers are the number of phase cycle steps in each one. the number of steps is unknown in 'indirect' and is therefore -1.
-    s.setaxis('ph1',r_[0:4.]/4) #setting values of axis ph1 to line up
-    s.setaxis('ph2',r_[0:2.]/4) #setting values of axis ph1 to line up
+    print("preprocessing IR acquired with ag_IR2H on bruker")
+    if fl is not None:
+        fl.next('raw data')
+        fl.image(s)
+    s.chunk('indirect',['indirect','ph1','ph2'],[-1,2,4]) #expands the indirect dimension into indirect, ph1, and ph2. 
+    #                                                      inner most dimension is the inner most in the loop in pulse sequence, 
+    #                                                      is the one on the farthest right. Brackets with numbers are the number 
+    #                                                      of phase cycle steps in each one. The number of steps is unknown in 
+    #                                                      'indirect' and is therefore -1.
+    s.setaxis('ph1',r_[0:2.]/4) #setting values of axis ph1 to line up
+    s.setaxis('ph2',r_[0:4.]/4) #setting values of axis ph1 to line up
     s.setaxis('indirect', s.get_prop('vd'))
-#titling to coherence domain
     s.ft('t2',shift=True) #fourier transform
     s.ft(['ph1','ph2']) #fourier transforming from phase cycle dim to coherence dimension
     s.reorder(['indirect','t2'], first=False)
     if fl is not None:
         s_forplot = s.C
+        s_forplot.setaxis('indirect','#')
+        s_forplot.set_units('indirect','scan #')
         fl.next('FT + coherence domain')
         fl.image(s_forplot)
     if fl is not None:    
@@ -32,7 +72,7 @@ def proc_bruker_deut_IR_mancyc(s, fl=None):
     return s
 
 def proc_spincore_CPMG_v1(s, fl=None):
-    logger.info("loading pre-processing for CPMG preprocessing")
+    print("preprocessing of CPMG acquired on SpinCore")
     SW_kHz = s.get_prop('acq_params')['SW_kHz']
     nPoints = s.get_prop('acq_params')['nPoints']
     nEchoes = s.get_prop('acq_params')['nEchoes']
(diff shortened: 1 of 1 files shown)


proc_scripts/load_data.py
```diff
@@ -52,27 +92,139 @@ def proc_spincore_CPMG_v1(s, fl=None):
...
+    if fl is not None:
+        fl.next('raw data with indirect set')
+        fl.image(s)
+    anavpt_info = [j for j in s.get_prop('pulprog').split('\n') if 'anavpt' in j.lower()]
+    anavpt_re = re.compile(r'.*\banavpt *= *([0-9]+)')
+    anavpt_matches = (anavpt_re.match(j) for j in anavpt_info)
+    for m in anavpt_matches:
+        if m is not None:
+            anavpt = int(m.groups()[0])
+    actual_SW = 20e6/anavpt # JF: check that this is based on the manual's definition of anavpt
```
jmfranck, line 130: block comment -- determine SW from anavpt

proc_scripts/load_data.py
```diff
@@ -52,27 +92,139 @@ def proc_spincore_CPMG_v1(s, fl=None):
...
+    s.ft('t2', shift=True).ft(['ph1','ph2'])
+    s.reorder(['ph1','ph2','indirect'])
+    if fl is not None:
+        fl.next('freq domain coh domain')
+        fl.image(s)
+    s.ift('t2')
+    if fl is not None:
+        fl.next('t2 chunked', figsize=(5,20))
+        fl.image(s)
+    s.ft('t2')
```
jmfranck, line 167: so, your standard is that the preprocessing FTs along both t2 and phase cycles?  I would definitely say we want the FT along phase cycles, I don't care about t2 (FT or not), but you should just pick and make it a standard

proc_scripts/load_data.py
```diff
@@ -52,27 +92,139 @@ def proc_spincore_CPMG_v1(s, fl=None):
...
+    return s
+
+def proc_bruker_CPMG_v1(s,d12,tau_extra,fl=None):
+    print("pre-processing CPMG acquired on Bruker")
+    s.chunk('indirect',['ph1','ph2','indirect'],[4,2,-1])
+    s.setaxis('ph1',r_[0:4]/4.)
+    s.setaxis('ph2',r_[0:2]/2.)
```
jmfranck, line 174: same comments

jmfranck (changes requested):
in addition to these detailed comments, I would recommend that you make this into a subpackage.

So, this file would become `proc_scripts/load_data/__init__.py`.

Then, for every preprocessing script, I would copy the function into its own file in the `load_data` directory and you can just do `from .proc_bruker_deut_IR_withecho_mancyc_v1  import proc_bruker_deut_IR_withecho_mancyc_v1`.

Also, if you see several lines of code that are shared across files, you can group them into their own function -- e.g. I see `SW_from_anavpt` as being a common function. (Again, I would put this in the `load_data` directory)

proc_scripts/load_data.py
```diff
@@ -3,22 +3,62 @@
 import os
 from sympy import symbols
 import logging
-#to use type s = load_data("nameoffile")
-def proc_bruker_deut_IR_withecho_mancyc(s):
-    raise RuntimeError("this is where postprocessing would be implemented -- not implemented yet")
+import numpy as np
+fl=figlist_var()
+def proc_bruker_deut_IR_withecho_mancyc(s,fl=fl):
+    print("preprocessing for IR acquired on bruker")
```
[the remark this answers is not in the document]
guinneak198: done

proc_scripts/load_data.py
```diff
@@ -3,22 +3,62 @@
...
+    if fl is not None:
+        fl.next('raw data')
+        fl.image(s.C.setaxis(
+'indirect','#').set_units('indirect','scan #'))
+    s.chunk('indirect',['ph2','ph1','indirect'],[4,2,-1]) #expands the indirect dimension into indirect, ph1, and ph2. 
+    #                                                      Inner most dimension is the inner most in the loop in pulse sequence, 
+    #                                                      is the one on the farthest right. Brackets with numbers are the number of 
+    #                                                      phase cycle steps in each one. The number of steps is unknown in 'indirect' 
+    #                                                      and is therefore -1.
+    s.setaxis('ph1',r_[0:2.]/4) #setting values of axis ph1 to line up
```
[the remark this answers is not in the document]
guinneak198: fixed here and recurring setaxis that were like this

proc_scripts/load_data.py
```diff
@@ -3,22 +3,62 @@
...
+    s.reorder(['indirect','t2'], first=False)
+    if fl is not None:
+        s_forplot = s.C
+        fl.next('FT')
+        fl.image(s_forplot.C.setaxis('indirect','#').set_units('indirect','scan #'))
+    if fl is not None:    
+        fl.next('time domain (all $\\Delta p$)')
+        s_forplot.ift('t2')
+        fl.image(s_forplot.C.setaxis('indirect','#').set_units('indirect','scan #'))
+    if fl is not None:
```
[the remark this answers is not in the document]
guinneak198: done

proc_scripts/load_data.py
```diff
@@ -32,7 +72,7 @@ def proc_bruker_deut_IR_mancyc(s, fl=None):
     return s
 
 def proc_spincore_CPMG_v1(s, fl=None):
-    logger.info("loading pre-processing for CPMG preprocessing")
+    print("preprocessing of CPMG acquired on SpinCore")
```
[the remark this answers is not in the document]
guinneak198: oops

proc_scripts/load_data.py
```diff
@@ -52,27 +92,139 @@ def proc_spincore_CPMG_v1(s, fl=None):
...
+    s.setaxis('ph1',r_[0,2]/4).setaxis('ph2',r_[0:4]/4)
+    s.ft(['ph1','ph2'])
+    s.reorder(['ph1','ph2','indirect'])
+    if fl is not None:
+        fl.next('raw data(t2,coh)')
+        fl.image(s)
+    #{{{removes CP aspect
+    s.ift(['ph1','ph2'])
+    s = s['ph2',[1,3]]
+    #}}}
```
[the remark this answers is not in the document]
guinneak198: okay so I will delete this for now since it seems we got rid of this in the acquisition portion anyhow

proc_scripts/load_data.py
guinneak198: yes so all preprocessing should return in the frequency domain with the phase cycles FTed. If you catch one that isn't that def slipped by me!

proc_scripts/load_data.py
```diff
@@ -3,22 +3,60 @@
...
-def proc_bruker_deut_IR_withecho_mancyc(s):
-    raise RuntimeError("this is where postprocessing would be implemented -- not implemented yet")
-
-def proc_bruker_deut_IR_mancyc(s, fl=None):
-    logger.info("going through this")
-    s.chunk('indirect',['indirect','ph1','ph2'],[-1,4,2]) #expands the indirect dimension into indirect, ph1, and ph2. inner most dimension is the inner most in the loop in pulse sequence, is the one on the farthest right. Brackets with numbers are the number of phase cycle steps in each one. the number of steps is unknown in 'indirect' and is therefore -1.
-    s.setaxis('ph1',r_[0:4.]/4) #setting values of axis ph1 to line up
-    s.setaxis('ph2',r_[0:2.]/4) #setting values of axis ph1 to line up
+import numpy as np
+fl=figlist_var()
```
jmfranck, line 7: you cannot / should not be defining a new instance of `figlist_var` here -- this is an issue we've covered before. -- default to `fl=None` below, and generate the instance inside your main script.

proc_scripts/load_data.py
```diff
@@ -3,22 +3,60 @@
...
+def proc_bruker_deut_IR_v1(s,fl=fl):
+    logger.info(strm("preprocessing for IR acquired on bruker"))
+    if fl is not None:
+        fl.next('raw data')
+        fl.image(s.C.setaxis(
+'indirect','#').set_units('indirect','scan #'))
```
jmfranck, line 13: indentation problem -- highlight and press `-` in gvim

proc_scripts/load_data.py
jmfranck, line 7: you cannot / should not be defining a new instance of `figlist_var` here -- this is an issue we've covered before. -- default to `fl=None` below, and generate the instance inside your main script.

proc_scripts/load_data.py
jmfranck, line 13: indentation problem --  highlight and press `=` in gvim

proc_scripts/load_data.py
```diff
@@ -52,27 +90,137 @@ def proc_spincore_CPMG_v1(s, fl=None):
...
+    bruker_final_t2_value = double(s.getaxis('t2')[-1].item())
+    s.setaxis('t2',1./actual_SW*r_[0:ndshape(s)['t2']]) # reset t2 axis to true values based on anavpt
+    logger.debug(strm("the final t2 value according to the Bruker SW_h was",
+        bruker_final_t2_value, "but I determine it to be",
+        double(s.getaxis('t2')[-1].item()), "with anavpt"))
+    nEchoes = s.get_prop('acq')['L'][25]
+    dwdel1 = s.get_prop('acq')['DE']*1e-6
+    dwdel2 = (anavpt*0.05e-6)/2
+    #d12 is read as 0 if taken from parameters bc its too small
+    d12 = d12
```
jmfranck, line 136: what's going on here?

proc_scripts/load_data.py
```diff
@@ -52,27 +90,137 @@ def proc_spincore_CPMG_v1(s, fl=None):
...
+        fl.next('raw data with indirect set')
+        fl.image(s)
+    #{{{ determining SW from anavpt
+    anavpt_info = [j for j in s.get_prop('pulprog').split('\n') if 'anavpt' in j.lower()]
+    anavpt_re = re.compile(r'.*\banavpt *= *([0-9]+)')
+    anavpt_matches = (anavpt_re.match(j) for j in anavpt_info)
+    for m in anavpt_matches:
+        if m is not None:
+            anavpt = int(m.groups()[0])
+    actual_SW = 20e6/anavpt # JF: check that this is based on the manual's definition of anavpt
```
jmfranck, line 125: did you check the definition of `anavpt` in the manual?  Is the clock 20 MHz?  Equivalently, it might talk about 50 ns per sampling period.

proc_scripts/load_data.py
```diff
@@ -52,27 +90,137 @@ def proc_spincore_CPMG_v1(s, fl=None):
...
+    quad_pts = ndshape(s)['t2'] # note that we have not yet chunked t2
+    nPoints = quad_pts/nEchoes
+    acq_time = dwdel2*nPoints*2
+    # {{{ these are hard-coded for the pulse sequence
+    #     if we need to, we could pull these from the pulse sequence, as we do
+    #     for anavpt above
+    tau_extra = tau_extra
+    tau_pad_start = tau_extra-dwdel1-6e-6
+    tau_pad_end = tau_extra-6e-6
+    twice_tau = 2*p90_s + 5e-6 + tau_pad_start + 1e-6 + acq_time + tau_pad_end +1e-6
```
jmfranck, line 204: so we would not know if this number was wrong, so we want to double-check it.  You should also be able to calculate this from the acquisition length and the padding on either side of  the acquisition.  We should double-check that both are equal with an `assert` statement.  You might need to get @aabeaton 's help to interpret what I'm saying here + how to implement it.

proc_scripts/load_data.py
```diff
@@ -95,43 +243,92 @@ def proc_Hahn_echoph(s, fl=None):
         fl.image(abs(s))
     return s
 
-def proc_spincore_IR(s,fl=None):
+def proc_spincore_IR(s,clock_correction,fl=None):
+    logger.info(strm("pre-processing for IR acquired on spincore"))
```
jmfranck, line 247: do not include the clock correction as an argument here -- that's not part of the preprocessing.  This is especially true now that we have the algorithm that does the clock correction w/out any reference data.

proc_scripts/load_data.py
```diff
@@ -95,43 +243,92 @@ def proc_Hahn_echoph(s, fl=None):
...
         fl.next('frequency domain (all $\\Delta p$)')
-        fl.image(s)
+        fl.image(s.C.setaxis('indirect','#').set_units('indirect','scan #'))
+    s *= exp(-1j*s.fromaxis('indirect')*clock_correction)
     return s
 
-def proc_nutation(s):
-    logging.info("loading pre-processing for nutation")
+def proc_nutation(s,fl=None):
+    logger.info(strm("pre-processing for p90 varied nutation"))
```
jmfranck, line 267: FYI ,you only need `strm` if you're trying to print a variety of arguments, as you would in a print statement w/ commas

proc_scripts/load_data.py
```diff
@@ -95,43 +243,92 @@ def proc_Hahn_echoph(s, fl=None):
...
+
+def proc_nutation_amp(s,fl=None):
+    logger.info(strm("pre-processing for amplitude varied nutation"))
+    orig_t = s.getaxis('t')
+    s.chunk('t',['ph2','ph1','t2'],[2,4,-1])
+    s.reorder(['ph1','ph2'])
+    s.setaxis('ph2',r_[0.,2.]/4)
+    s.setaxis('ph1',r_[0.,4.]/4)
+    s.set_units('t2','s')
+    s.set_units('amp','unknown')
```
jmfranck, line 293: I would say this is just unitless? (b/c it's ratio relative to the max setting)

proc_scripts/load_data.py
```diff
@@ -152,32 +349,26 @@ def proc_spincore_ODNP_v1(s,fl=None):
...
     d11 = s.get_prop('acq')['D'][11]
     p1 = s.get_prop('acq')['P'][1]
     ppg = s.get_prop('pulprog')
-    # {{{ these are explanatory -- maybe comment them out?
     m = re.search(('.*dwdel1=.*'),ppg,flags=re.IGNORECASE)
     logging.info(strm(m.groups())) # show the line that sets dwdel1
-    # then look for de and depa
     logging.info(strm([(j,s.get_prop('acq')[j]) for j in s.get_prop('acq').keys() if 'de' in j.lower()]))
-    # I actually can't find depa
-    # }}}
```
jmfranck, line 175: don't remove these comments

proc_scripts/load_data.py
```diff
@@ -189,14 +380,11 @@ def proc_DOSY_CPMG(s):
     quadrature_points = TD/2
     num_points_per_echo = quadrature_points/l25
     acq_time = dwdel2*num_points_per_echo*2
-    # {{{ so, in principle, later, we can/should do what I did above (w/ eval),
-    # but it's getting crazy now, so I stop for now
-    tau_extra = 20e-6
+    tau_extra = tau_extra
```
jmfranck, line 383: `a = a` is a do-nothing statement.  I remove it here and elsewhere.

proc_scripts/load_data.py
```diff
@@ -210,17 +398,56 @@ def proc_DOSY_CPMG(s):
...
+            forcibly sets a specific symbol
+        set_to: double, optional
+            the specific value (int) you are assigning the symbol you included
+
+        Returns
+        -------
+        self: nddata
+            the fit function evaluated along the axis coordinates that were passed
+        """
+
```
jmfranck, line 426: what is going on here??

jmfranck (changes requested):
see comments -- I'm also pushing my own commit

proc_scripts/load_data.py
guinneak198: not sure. Took out after realizing d12 isn't even used in the preprocessing

proc_scripts/load_data.py
guinneak198: I pulled these variables straight from the pulse program. These are how they are defined there. At least up until twice_tau. 
image
this is the pulse sequence timing where we calculate twice tau

proc_scripts/load_data.py
guinneak198: I think you are referring to the units of am? so I changed unknown to unitless?

proc_scripts/load_data.py
guinneak198: lol I saw this before this comment and thought the same thing! I deleted sorry about that! I don't know how I didn't catch this!

proc_scripts/load_data.py
jmfranck: this doesn't address my point, which is that twice tau should also be equal to
`dwdel1+5e-6+tau_pad_start+1e-6+dwdel2*num_pts+tau_pad_end+1e-6+5e-6` (the time from one 180 pulse to the next) --> so we want an assertion statement that also checks that it's equal to this expression

proc_scripts/load_data.py
jmfranck: yes, units of amp -- but you can just set the units to `None`

proc_scripts/load_data.py
jmfranck: this is still outstanding until you reply

guinneak198: image
I think this answers your question that yes the clock is 20MHz

proc_scripts/load_data.py
guinneak198: I'm confused... we are asserting that twice_tau=twice_tau?

jmfranck: your expression involves the 90 time, while mine does not  -- they are not the same expression

jmfranck: more specifically -- tau should be equal to two time periods:

- the equivalent evolution time during the 90 pulse (given by Cavanagh -- 2*t90/π, from pg 266 Eq. [3.199]), plus the time from the end of the ninety up to the start of the first 180 (what I think you calculate above)
- half of the total time from the end of one 180 pulse until the start of the next

guinneak198: okay so since I am using twice_tau this statement should follow those bullet points correct?

guinneak198: or did I forget to divide 2*p_90 by pi?

jmfranck: yes, it should follow those statements & it does look like you forgot to divide by pi

jmfranck: actually, why not just define two variables -- `ninety_through_180` and `between_180s`, and then check `assert ninety_through_180 == between_180s`

(`twice_tau` is ambiguous in the current discussion)

proc_scripts/load_data.py
```diff
@@ -52,27 +102,134 @@ def proc_spincore_CPMG_v1(s, fl=None):
...
+    acq_time = dwdel2*nPoints*2
+    # {{{ these are hard-coded for the pulse sequence
+    #     if we need to, we could pull these from the pulse sequence, as we do
+    #     for anavpt above
+    tau_pad_start = tau_extra-dwdel1-6e-6
+    tau_pad_end = tau_extra-6e-6
+    twice_tau = 2*p90_s + 5e-6 + tau_pad_start + 1e-6 + acq_time + tau_pad_end +1e-6
+    #twice_tau should be the period from one 180 to another
+    # }}}
+    assert twice_tau = dwdel1 + 5e-6 +tau_pad_start + 1e-6 + dwdel2*nPoints + tau_pad_end + 1e-6 + 5e-6, "twice_tau is not equal to the time from one 180 pulse to the next"
```
jmfranck, line 215: does this even run? you would need a `==` not a `=`.  Just see my most recent comment in [this thread](https://github.com/jmfrancklab/proc_scripts/pull/34/files#r559809273), which I think should clear up any remaining confusion + get us on our way!

jmfranck (changes requested):

proc_scripts/load_data.py
guinneak198: works on my end now

guinneak198:
Am I missing something? this should be ready to merge.... I want to hold off on making the sub package until after all the processing scripts are updated so we can go step by step and right now in bruker_processing we know this works. Then once master is updated to be working again I can do the subpackage?

PR #35, opened 2021-01-18
Jf tweak fit

jmfranck:


proc_linewidth.py
```diff
@@ -9,57 +9,71 @@
...
+            False,
+            1e-3
+            )
         ]:
     d = find_file(searchstr + '.DSC', exp_type=exp_type, postproc=postproc,
                   lookup=postproc_dict)
-    fl.next('linewidth for 10mM 4AT')
-    fl.plot(d)
+    plt.figure()
+    plt.title('linewidth for 10mM 4AT')
```
guinneak198, line 48: why not use fl.plot? why use the matplotlib stuff?

proc_linewidth.py
```diff
@@ -9,57 +9,71 @@
...
             print("no pickle file found -- generating")
             z = ((B-B_center) + s.I*R)/sigma/s.sqrt(2)
             faddeeva = s.simplify(s.exp(-z**2) * s.erfc(-s.I*z))
             voigt = A*s.re(faddeeva)/sigma/s.sqrt(2*s.pi)
+            voigt *= sigma * R # so adjusting linewidth doesn't change amplitude
             voigt = voigt.simplify()
-            dVoigt = voigt.diff(B).simplify()
+            # add real below b/c lambdify was giving complex answer
+            dVoigt = s.re(s.re(voigt.diff(B)).simplify())
             pickle.dump(dVoigt,fp)
```
guinneak198, line 66: so will this generate a pickle file each time I run it hence I will need to delete the pickle file every time before I run it?

proc_linewidth.py
```diff
@@ -72,23 +86,32 @@
...
-    fl.plot(guess_nddata,':', label='guess')
-    fl.next('guess')
+    plot(d, label='data')
+    plot(guess_nddata,':', label='guess')
     model = s.Model({y_var:dVoigt})
-    guess = InteractiveGuess(model, y=d.data.real, B=d.getaxis('$B_0$'), n_points=500)
-    guess.execute()
-    print(guess)
+    if interactive:
+        guess = InteractiveGuess(model, y=d.data.real, B=d.getaxis('$B_0$'), n_points=500)
```
guinneak198, line 93: what is this function InteractiveGuess? did we define it or is it some numpy thing?

proc_linewidth.py
```diff
@@ -72,23 +86,32 @@
...
-fl.show()
+    conc_list.append(concentration)
+    R_list.append(fit_result.params['R'])
+    sigma_list.append(fit_result.params['sigma'])
+plt.figure()
+plot(conc_list,R_list,'x',label='R')
+plot(conc_list,sigma_list,'x',label=r'$\sigma$')
+plt.legend(**dict(bbox_to_anchor=(1.05,1), loc=2, borderaxespad=0.))
+plt.ylabel('concentration')
+plt.show()
```
guinneak198, line 117: And so for this PR you want me to try to integrate in the global fit?

guinneak198 (approved):

proc_linewidth.py
jmfranck: Because the interactive plot is as already directly accessing matplotlib features

proc_linewidth.py
jmfranck: No!  You only need to regenerate the pickle if you mess with the expression. The final expression takes a while to calculate, which is why I cache it in a pickle file

proc_linewidth.py
jmfranck: Symfit. You can tell this based on how it’s imported

proc_linewidth.py
jmfranck: No, here I am making sure that you understand everything that we did, since this was from the earlier meeting. The global fit should go in a separate pull request.

proc_linewidth.py
guinneak198: so is it a type of fit? similar to how global fit will fit a bunch of lines or how there are different minimizers/parameters for different fits?

proc_linewidth.py
guinneak198: okay so the only thing that is a little wonky to me is the interactive guess but I can clarify this by going to the documentation. Other than that this should be good to merge!

proc_linewidth.py
jmfranck: No -- this opens the window that allows you to adjust the parameters.  It's documented [here](https://symfit.readthedocs.io/en/stable/examples/ex_interactive_guesses_vector_2D.html)

PR #40, opened 2021-03-03
Merge all "bruker processing" stuff

jmfranck:


align_slice.py
```diff
@@ -1,30 +0,0 @@
-from pyspecdata import *
-def align_and_slice(s, dimension='t2', convwidth=500, threshold=0.05, fl=None):
-    r"""Align the peak frequencies, and automatically slice them -- passing fl assumes you are debugging and want diagnostic plots"""
-    if fl is not None:
-        fl.push_marker()
```
jmfranck, line 5: why is this no good + why are we deleting it?

align_slice.py
```diff
@@ -1,30 +0,0 @@
...
-    center_frq = abs(s_conv).argmax(dimension)
-    s *= exp(-1j*2*pi*center_frq*s.fromaxis(dimension))
-    s_conv = s_conv_save * exp(-1j*2*pi*center_frq*s.fromaxis(dimension))
-    s.ft(dimension)
-    s_conv.ft(dimension)
-    s_conv.mean_all_but([dimension])
-    slices = s_conv.contiguous(lambda x:
-            abs(x)>threshold*abs(s_conv.data).max())
-    if fl is None:
-        assert slices.shape[0] == 1, "found more than one peak: "+str(slices)
```
jmfranck, line 25: specifically -- this guy is using the convolved lineshape to determine a good slicing range --> can you point to where that is done currently?

examples/proc_square_refl.py
```diff
@@ -0,0 +1,107 @@
...
+class fl_ext(figlist_var):
+    def next(self, *arg, **kwargs):
+        kwargs.update({"figsize": (9, 5.56), "legend": True})
+        super().next(*arg, **kwargs)
+
+    def complex_plot(fl, d, label="", show_phase=False, show_real=True,alpha=0.5,linestyle=None,linewidth=3,color='k'):
+        colors = []
+        for j in range(ndshape(d)["ch"]):
+            chlabel = d.getaxis("ch")[j]
+            if j==0:
```
jmfranck, line 15: why isn't this done in a separate module?

fitdata_test.py
```diff
@@ -1,73 +0,0 @@
-from pyspecdata import *
-from proc_scripts import hermitian_function_test,zeroth_order_ph
-init_logging(level='debug')
-
-#{{{ code bit from proc_IR
-from scipy.optimize import leastsq,minimize
-#from hermitian_function_test import hermitian_function_test, zeroth_order_ph
-from sympy import symbols
-fl = figlist_var()
-t2 = symbols('t2')
```
jmfranck, line 10: so this is a copy of a file in pyspecdata?

globalfit_linewidth.py
```diff
@@ -0,0 +1,133 @@
...
+from proc_scripts import postproc_dict
+import symfit as s
+import pickle,os
+from pylab import ndarray
+from symfit import Parameter,Variable, Fit, Model
+from symfit.core.minimizers import MINPACK
+import numpy as np
+import sympy as sp
+from itertools import cycle
+B_name = '$B_0$'
```
jmfranck, line 12: so this should be an example, and not in the main directory

globalfit_lmfit_linewidth.py
```diff
@@ -0,0 +1,92 @@
...
+import symfit as s
+import pickle,os
+from pylab import ndarray
+from symfit import Parameter,Variable, Fit, Model
+from symfit.core.minimizers import MINPACK
+import numpy as np
+import sympy as sp
+from itertools import cycle
+from lmfit import Parameters, minimize,Minimizer
+B_name = '$B_0$'
```
jmfranck, line 13: why not in examples directory?

jmfranck (commented):
during individual meeting

proc_GDS.py
```diff
@@ -1,18 +0,0 @@
...
-    s = nddata_hdf5(filename+'/'+nodename,
-            directory = getDATADIR(
-                exp_type = 'test_equip'))
-    s.set_units('t','s')
-    s = s['ch',0].C
-    print(ndshape(s))
-    fl.next('idk')
-    for i,x in enumerate(s.getaxis('capture')):
-        fl.plot(s['capture',i],alpha=0.2)
-    fl.show();quit()
```
guinneak198, line 18: we have replaced this file with specific processing of things captured on the gds for example proc_triwave, proc_capture_nutation, proc_chirp. This is a general just plot what I captured. Not sure how specific/general you want these scripts to be... Could possibly be made into a second level function if those other scripts do the exact same thing?

proc_Hahn_echo_phc3.py
```diff
@@ -1,85 +0,0 @@
-from pyspecdata import *
```
guinneak198, line 1: see proc_Hahn_echo_phc1

proc_Hahn_echo_phc4.py
```diff
@@ -1,89 +0,0 @@
-from pyspecdata import *
```
guinneak198, line 1: see proc_Hahn_echo_phc1

proc_IR_noecho.py
```diff
@@ -1,103 +0,0 @@
-from pyspecdata import *
```
guinneak198, line 1: I have modified proc_IR to include arguments that make this script obsolete./ proc_IR.py does what this script does but better/more up to date

proc_MWE.py
```diff
@@ -1,27 +0,0 @@
-from pyspecdata import *
```
guinneak198, line 1: this file is literally just plotting raw data and doing some waterfall function. I'm guessing MWE is just microwave enhancement but what specifically it is enhancing I do not know. This is probably one of the files originally pulled form the OLD proc_scripts when Alec first came in

proc_calibrate_clock.py
```diff
@@ -1,48 +0,0 @@
-from pyspecdata import *
```
guinneak198, line 1: I believe now we have an automated clock correction implemented into the scripts which makes this obsolete?

proc_gradient.py
```diff
@@ -1,164 +0,0 @@
-from pyspecdata import *
```
guinneak198, line 1: So I wasn't sure if you wanted to keep scripts which I did not create in this repo. This here is Alec's processing. I have never performed anything with gradients. I can add this back in but I definitely am not up to speed on what's going on in this script since I haven't performed this type of experiment before.

proc_scope.py
```diff
@@ -1,32 +0,0 @@
-from pyspecdata import *
```
guinneak198, line 1: see proc_GDS

proc_square_refl.py
```diff
@@ -1,201 +0,0 @@
-from pyspecdata import *
-from scipy.optimize import minimize,leastsq
```
guinneak198, line 2: we moved this to the examples directory

guinneak198 (commented):
went through to give explanation to "deleted" files.

jmfranck (commented):
Made comments and stopped partway through. Please do read through all changes on the code changes page and anticipate further comments

proc_CPMG.py
```diff
@@ -3,29 +3,76 @@
...
-    s =  find_file(searchstr, exp_type=exp_type,
-            expno=nodename, postproc=postproc, lookup=postproc_dict)
+    s = find_file(searchstr, exp_type=exp_type,
+            expno=nodename, postproc=postproc, lookup=postproc_dict, fl=fl)
+    s.ift('t2')
+    if spincore:
+        s.reorder('nScans',first=True)
+        s = s['ph1',1]
+        s.mean('nScans')
+        s.reorder('t2',first=True)
```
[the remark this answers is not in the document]
jmfranck: why are you selecting different coherence pathways in the two cases?  (add as comment)

proc_CPMG.py
```diff
@@ -3,29 +3,76 @@
...
+    else:
+        s.ft('t2')
+        s = s['ph2',-2]['ph1',1]
+        s.ift('t2')
     #{{{ centering CPMG echo
-    s = center_CPMG_echo(s)
+    center = find_echo_center(s)
```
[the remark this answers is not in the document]
jmfranck: what does it do that's "a step further?"  Reading this, I don't know what it does.

actually I looked up the code, and find_echo_center literally just calls the hermitian phasing.

proc_IR.py
```diff
@@ -4,45 +4,53 @@
...
 t2 = symbols('t2')
 logger = init_logging("info")
 # {{{ input parameters
-clock_correction = 1.785
-filter_bandwidth = 5e3
+filter_bandwidth = 15e3
 coh_sel = {'ph1':0,
-        'ph2':1}
+        'ph2':-1}
 coh_err = {'ph1':1,# coherence channels to use for error
```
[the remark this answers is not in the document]
jmfranck: you want to add a docstring at the very top of the file (just a text explanation in triple quotes of what this file does) to explain, so we don't keep asking this question

jmfranck (commented):
worked through several files

align_slice.py
jmfranck: ok, good

proc_GDS.py
jmfranck: if you have the files that do what you want, for GDS, probably OK to delete this

proc_calibrate_clock.py
jmfranck: no, this is still an experiment that we want -- please leave it in

proc_CPMG.py
```diff
@@ -3,29 +3,77 @@
 from proc_scripts import *
 from proc_scripts import postproc_dict
 from proc_scripts.fitting import decay
+from sympy import symbols
+import sympy as sympy
```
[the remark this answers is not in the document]
jmfranck: maybe address w/ @aabeaton -- this has a conditional "if spincore" --> but that's exactly the purpose of having preprocessing.

proc_T1CPMG.py
```diff
@@ -1,137 +1,144 @@
...
-    title(r'$T_{1} - T_{2} distribution$')
-    image(x)
-    xlabel(r'$log(T_2/$s$)$')
-    ylabel(r'$log(T_1/$s$)$')
-    np.savez('proc_'+date+'_'+id_string+'_1',
-            data = x.data,
-            logT1 = x.getaxis('T1'),
-            logT2 = x.getaxis('T2'))
-fl.show();quit()
+    if write_h5:
```
jmfranck, line 30: why is everything just embedded in this one conditional?

proc_capture_nutation.py
```diff
@@ -0,0 +1,72 @@
...
+        fl.plot(abs(d['p90',0]),alpha=0.5, linewidth=1,label = "p90 = 0.1 us")
+        fl.plot(abs(d['p90',1]),alpha=0.5,linewidth=1,label = "p90=1.76 us")
+        fl.plot(abs(d['p90',2]),alpha=0.5,linewidth=1,label = "p90=3.41 us")
+        fl.plot(abs(d['p90',3]),alpha=0.5,linewidth=1,label = "p90=5.07 us")
+        fl.plot(abs(d['p90',4]),alpha=0.5,linewidth=1,label = "p90=6.72 us")
+        fl.plot(abs(d['p90',5]),alpha=0.5,linewidth=1,label = "p90=8.38 us")
+        fl.plot(abs(d['p90',6]),alpha=0.5,linewidth=1,label = "p90=10.03 us")
+        fl.plot(abs(d['p90',7]),alpha=0.5,linewidth=1,label = "p90=11.69 us")
+        fl.plot(abs(d['p90',8]),alpha=0.5,linewidth=1,label = "p90=13.34 us")
+        fl.plot(abs(d['p90',9]),alpha=0.5,linewidth=1,label = "p90=15 us")
```
jmfranck, line 37: this needs to be a loop

proc_chirp.py
```diff
@@ -1,108 +1,42 @@
...
+for searchstr,exp_type,nodename,postproc,corrected_volt,y_lim in [
+        ('201113_triwave_control_2','ODNP_NMR_comp','capture1','chirp',True,1.25)
+        ]:
+    a = find_file(searchstr, exp_type=exp_type, expno=nodename,
+            postproc=postproc, lookup=postproc_dict)
+fl.next('Raw signal')
+fl.plot(a['ch',0], alpha=0.5, label='control')    
+fl.plot(a['ch',1], alpha=0.5, label='reflection')
 #}}}
-with figlist_var(filename='chirp.pdf') as fl:
```
jmfranck, line 7: I don't understand why you removed the "with figlist_var" block --> why not just go back to the with block

proc_scripts/correlation_alignment.py
```diff
@@ -0,0 +1,149 @@
...
+                    name for the figures generated
+    tol:            float
+                    sets the tolerance limit for the alignment procedure
+    ph1_selection:  int
+                    index position of the coherence pathway in phase program 1
+    ph2_selection:  int
+                    index position of the coherence pathway in phase program 2
+    sigma:          int
+                    sigma value for the gaussian fitting. Related to the linewidth
+                    of the given data.
```
jmfranck, line 40: this should accept an fl argument (the figure list it uses to generate the figures) and a mask argument

jmfranck: (the mask is for the masked norm)

proc_scripts/correlation_alignment.py
```diff
@@ -0,0 +1,149 @@
...
+                fl.image(s_copy.C.setaxis('vd','#').set_units('vd','scan #'))
+        logger.info(strm('signal energy per transient (recalc to check that it stays the same):',(abs(s_copy**2).data.sum().item() / N)))
+
+        this_E = (abs(s_copy.C.sum(indirect_dim))**2).data.sum().item() / N**2
+        energy_vals.append(this_E / sig_energy)
+        logger.info(strm('averaged signal energy (per transient):', this_E))
+        if last_E is not None:
+            energy_diff = (this_E - last_E)/sig_energy
+            logger.info(strm(energy_diff))
+            if abs(energy_diff) < tol:
```
jmfranck, line 138: also require that at least 4 iterations have been run

proc_scripts/load_data.py
```diff
@@ -214,13 +409,37 @@ def proc_DOSY_CPMG(s):
     # }}}
     return s
 
+def proc_ESR_linewidth(s):
```
jmfranck, line 412: why is this called "linewidth"?  Isn't it appropriate for other cases?

proc_scripts/phasing.py
```diff
@@ -2,6 +2,10 @@
 from pyspecdata import *
 from matplotlib.patches import Ellipse
 from scipy.optimize import minimize
+import numpy as np
+import pyspecdata as pysp
+from scipy import linalg
+logger=init_logging('info')
```
jmfranck, line 8: don't init logging here -- init logging in the main script only

proc_scripts/phasing.py
```diff
@@ -53,7 +57,7 @@ def zeroth_order_ph(d, fl=None):
     # that we only want to rotate when the distribution
     # is assymetric, so include only the excess of the
     # larger eval over the smaller
-    assymetry_mag = sqrt(eigenValues[0])-sqrt(eigenValues[1])
+    assymetry_mag = float(sqrt(eigenValues[0])-sqrt(eigenValues[1]))
```
jmfranck, line 60: why??

proc_scripts/phasing.py
```diff
@@ -63,8 +67,8 @@ def zeroth_order_ph(d, fl=None):
         rotation_vector = mean_vec + assymetry_mag*eigenVectors[:,0]
     else:
         rotation_vector = mean_vec - assymetry_mag*eigenVectors[:,0]
-    ph0 = arctan2(rotation_vector[1],rotation_vector[0])
-    if fl:
+    ph0 = np.arctan2(rotation_vector[1],rotation_vector[0])
+    if fl is not None:
```
jmfranck, line 71: why is "if fl" not sufficient?

proc_scripts/phasing.py
```diff
@@ -127,44 +131,49 @@ def ph1_real_Abs(s,dw):
...
         ph0 = arg.C.sum('t2')
         ph0 /= abs(ph0)
         arg /= ph0
         return arg
     def costfun(ph1):
-        if type(ph1) is ndarray:
-            ph1 = ph1.item()
+        logger.info(strm("PH1 IS THIS:",ph1))
+        if type(ph1) is np.ndarray:
+            ph1 = ph1[0].item()
```
jmfranck, line 164: 164 is different in meaning than what it's replacing -- it means pull the first and use that -- it will not error out if there is more than one the way that the first statement will.

proc_scripts/phasing.py
```diff
@@ -127,44 +131,49 @@ def ph1_real_Abs(s,dw):
...
         temp = s.C
-        retval = applyphase(temp,ph1).run(real).run(abs).sum('t2').item()
+        retval = applyphase(temp,ph1).run(np.real).run(abs).sum('t2').item()
         return retval
-    print("rough opt cost function is",costfun(ph1_opt))
-    r = minimize(costfun,ph1_opt,
-            bounds=[(ph1_opt-dx,ph1_opt+dx)])
+    r = minimize(costfun,
+            ph1_opt)
```
jmfranck, line 169: why did you remove the bounds?

proc_scripts/plotting.py
```diff
@@ -1,5 +1,8 @@
 from pyspecdata import *
 from sympy import symbols
+import matplotlib.pyplot as plt
+import numpy as np
+logger=init_logging('info')
```
jmfranck, line 5: again, init logging in the main script.  here, use "logging.info" or "logging.debug"

jmfranck: import logging if needed

proc_triwave.py
```diff
@@ -0,0 +1,41 @@
...
+from pyspecdata import *
+from proc_scripts import postproc_dict
+from scipy import signal
+import matplotlib.pyplot as plt
+
+fl = figlist_var()
+ # {{{ load data, set units, show raw data
+for searchstr,exp_type,nodename,postproc,corrected_volt in [
+        ('201202_triwave_RMprobe_20ms','ODNP_NMR_comp','capture1','chirp',True)
+        ]:
```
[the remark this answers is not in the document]
jmfranck: this would go into examples

proc_var_tau.py
```diff
@@ -0,0 +1,37 @@
+from pylab import *
+from pyspecdata import *
+from proc_scripts.load_data import postproc_dict
+import h5py as h5
+
+with figlist_var() as fl:
+    for filename, expno, exp_type, postproc, frequency, f_range in [
+            ("201209_Ni_sol_probe_var_tau",'var_tau','var_tau','spincore_var_tau_v1',
```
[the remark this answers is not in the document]
jmfranck: examples

q
```diff
@@ -0,0 +1,17 @@
+proc_IR.py has set guess function but not usable yet
+# Please enter the commit message for your changes. Lines starting
+# with '#' will be ignored, and an empty message aborts the commit.
+#
+# On branch alec_fitdata_rebased
+# Your branch is up to date with 'origin/alec_fitdata_rebased'.
+#
+# Changes to be committed:
```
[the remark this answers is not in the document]
jmfranck: delete it

proc_scope.py
jmfranck: ???

proc_CPMG.py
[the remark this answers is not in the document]
jmfranck: this isn't spincore vs. not, it's whether or not the second pulse has been phase cycled.  You should be able to do something like `if 'ph2' in s.dimlabels` which would be the non-spincore option and `else` which would be the spincore option.  Similarly, `if 'nScans' in s.dimlabels: s.mean('nScans')`, rather than making it part of this "spincore" conditional

proc_CPMG.py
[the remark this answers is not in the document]
jmfranck: yes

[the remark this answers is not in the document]
jmfranck: there should be no function `find_echo_center` -- it literally just calls hermitian phasing

proc_CPMG.py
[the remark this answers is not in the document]
jmfranck: you don't need to address w/ @aabeaton -- see my other comments first, which I think will resolve this issue

proc_echo_mw.py
```diff
@@ -14,29 +17,29 @@
 # to use: as a rule of thumb, make the white boxes
 # about 2x as far as it looks like they should be
 # leave this as a loop, so you can load multiple files
-for searchstr,exp_type,nodename,postproc,freq_range,time_range in [
-        ["200306_DNP_lg_probe_w34.*", 'test_equip', 'signal',
-            'spincore_ODNP_v1', (-300,300), (None,0.05)]
+for searchstr,exp_type,nodename,postproc,freq_range,max_t in [
+        ["210310_TEMPOL500uM_DNP_cap_probe_1", 'ODNP_NMR_comp', 'signal',
+            'spincore_ODNP_v1', (-2000,2000),0.06]
```
jmfranck, line 22: this is a roll back -- you should be giving a time range, not just a max time.

proc_nutation.py
```diff
@@ -1,53 +1,84 @@
+from pylab import *
 from pyspecdata import *
 from scipy.optimize import minimize
-from proc_scripts import hermitian_function_test,zeroth_order_ph,postproc_dict,fl_mod,center_CPMG_echo
+from proc_scripts import hermitian_function_test,zeroth_order_ph,postproc_dict,fl_mod
```
jmfranck, line 4: is center CPMG echo just calling the hermitian function test?  Make sure that you are using these imports, and delete any not used.

proc_nutation.py
```diff
@@ -1,53 +1,84 @@
...
+    #['210302_Ni_cap_probe_nutation_1','nutation','nutation',
+    #    'spincore_nutation_v1',(-4e3,1.2e4),True]
     ]:
     s = find_file(searchstr,exp_type=exp_type,expno=nodename,postproc=postproc,
-            lookup=postproc_dict) 
- 
+            lookup=postproc_dict)#,fl=fl) 
+    if not p90_varied:
+        plen = s.get_prop('acq_params')['p90_us']
+        plen *= 10**-6
```
jmfranck, line 21: this should just be one line

proc_nutation.py
```diff
@@ -1,53 +1,84 @@
...
+for searchstr,exp_type,nodename,postproc,freq_slice,p90_varied in [
+    ['201211_Ni_sol_probe_nutation_1','nutation','nutation',
+        'spincore_nutation_v1',(-5000,13000),True],
+    #['210302_Ni_cap_probe_nutation_1','nutation','nutation',
+    #    'spincore_nutation_v1',(-4e3,1.2e4),True]
     ]:
     s = find_file(searchstr,exp_type=exp_type,expno=nodename,postproc=postproc,
-            lookup=postproc_dict) 
- 
+            lookup=postproc_dict)#,fl=fl)
```
jmfranck, line 18: don't add commented out code

proc_nutation.py
```diff
@@ -1,53 +1,84 @@
...
+        plen = s.get_prop('acq_params')['p90_us']
+        plen *= 10**-6
     # {{{ do the rough centering before anything else!
     # in particular -- if you don't do this before convolution, the
     # convolution doesn't work properly!
-    s.ft(['ph2','ph1'])
-    # }}}
-    
+    s.ift('t2')
+    s.setaxis('t2', lambda x: x-abs(s['ph1',1]['ph2',-2]).mean_all_but('t2').argmax('t2').item())
```
jmfranck, line 26: this is just a coarse correction -- why not just use Hermitian function test?

proc_nutation.py
```diff
@@ -1,53 +1,84 @@
...
-    s.setaxis('t2', lambda x: x-best_shift).register_axis({'t2':0})
+    if p90_varied:
+        best_shift = hermitian_function_test(s['ph2',0]['ph1',1])
+        logger.info(strm("best shift is",best_shift))
+        s.setaxis('t2', lambda x: x-best_shift).register_axis({'t2':0})
+        fl.next('with time shift')
+        fl.image(s)
+        fl.next('freq domain after time correction')
+        s.ft('t2')
+        fl.image(s)
```
jmfranck, line 43: I don't understand why you are only running the hermitian function test if you have a variable p90?

proc_nutation.py
```diff
@@ -1,53 +1,84 @@
...
+from proc_scripts import hermitian_function_test,zeroth_order_ph,postproc_dict,fl_mod
 from sympy import symbols
 from numpy import *
 fl = fl_mod()
 t2 = symbols('t2')
 logger = init_logging("info")
-for searchstr,exp_type,nodename,postproc in [
-    ['200219_nutation_alex_probe','test_equip','nutation','spincore_nutation_v1']
+max_kHz = 200
+for searchstr,exp_type,nodename,postproc,freq_slice,p90_varied in [
```
jmfranck, line 11: you should be able to detect whether or not p90 is varied based on the naming of the dimensions, so you shouldn't need to specify it here.

proc_nutation.py
```diff
@@ -1,53 +1,84 @@
...
+    fl.show();quit()
     #}}}
-    
+    if 'amp' in s.dimlabels:
+        s.setaxis('amp',lambda x:x*plen)
+        s.set_units('amp','s')
+        ind_dim = '\\tau_p a'
+        s.rename('amp',ind_dim)
+    if'p_90' in s.dimlabels:
+        ind_dim = 'p_90'
```
jmfranck, line 64: this is related to my comment above -- that's what I'm doing here.

proc_nutation.py
```diff
@@ -1,53 +1,84 @@
...
     #{{{ phasing with zeroth order correction
-    s.ift('t2')
     fl.next('final time domain')
     ph0 = zeroth_order_ph(s['t2':0], fl=None)
     s /= ph0
     fl.image(s)
     fl.next('phased')
-    s.ft('t2',pad=4096)
+    #s.ft('t2',pad=4096)
```
jmfranck, line 71: ahhh

proc_scope.py
jmfranck: never mind -- there's nothing of significance here -- I'm cool with a delete

proc_scripts/CPMG_phasing.py
```diff
@@ -2,7 +2,9 @@
 from sympy import symbols
 from proc_scripts import *
 import math
-def center_CPMG_echo(s, axis='t2',fl=None):
+import numpy as np
+logger=init_logging('info')
+def find_echo_center(s, axis='t2',fl=None):
```
jmfranck, line 7: so this function offers nothing new, and should not exist.

you might want to copy the documentation over to hermitian function test, if it doesn't have any.

proc_scripts/CPMG_phasing.py
```diff
@@ -18,13 +20,22 @@ def center_CPMG_echo(s, axis='t2',fl=None):
     """
     echo_center = hermitian_function_test(s, fl=fl)
     logger.info(strm("echo center is",echo_center))
+    return echo_center 
+def center_echo(s, echo_center, axis='t2',fl=None):
```
jmfranck, line 24: this guy needs documentation.  I'm not going to read the code until I know what it's supposed to do.

proc_scripts/STFT.py
```diff
@@ -0,0 +1,74 @@
...
+
+    t: time array
+    fm: modulation frequency
+    fd: doppler frequency shift
+    B: bandwidth
+    '''
+    f = B*0.5*(signal.sawtooth(2 * np.pi * fm * t,
+        width=width) +1)
+    f += fd
+    return f
```
jmfranck, line 74: these functions are generating fake data, and so should be in the example script that demonstrates the other code that's in this module.

proc_scripts/__init__.py
```diff
@@ -1,16 +1,19 @@
 from .phasing import zeroth_order_ph,hermitian_function_test,ph1_real_Abs
-from .align_slice import align_and_slice
+#from .align_slice import align_and_slice
```
jmfranck, line 2: ahhhhhhhh

proc_scripts/__init__.py
```diff
@@ -1,16 +1,19 @@
...
+from .CPMG_phasing import find_echo_center, center_echo
+from .fitting import recovery, decay
+from .STFT import stft, isftf, simu_waves, simu_freq_sawtooth 
+from .integrate_limits import integrate_limits
+from .correlation_alignment import correl_align
 __all__ = ['zeroth_order_ph',
         'ph1_real_Abs',
         'hermitian_function_test',
-        'align_and_slice',
+        #'align_and_slice',
```
jmfranck, line 16: ahhhh

proc_scripts/__init__.py
```diff
@@ -19,6 +22,15 @@
...
+        'find_echo_center',
+        'center_echo',
+        'recovery',
+        'decay',
+        'stft',
+        'isftf',
+        'simu_waves',
+        'simu_freq_sawtooth',
+        'integrate_limits',
+        'correl_align'
```
jmfranck, line 34: I would sort this whole list alphabetically.

shift-V to highlight then `:sort`

proc_scripts/baseline.py
```diff
@@ -1,8 +1,10 @@
+from pyspecdata import *
+logger=init_logging('info')
```
jmfranck, line 2: as elsewhere -- `import logging` here, then `logging.info(`, etc, in the code.

this command would initiate logging, which should only be done in the main example script.

proc_scripts/baseline.py
```diff
@@ -53,10 +56,10 @@ def costfun(ini_vec):
         d_test = apply_corr(ini_vec)
         return abs(d_test.real).sum('t2').data.item()
     max_val = abs(this_d_tdom.data).max()
-    print(max_val)
+    logger.info(strm(max_val))
     mybounds = r_[-max_val,max_val][newaxis,:]*ones(npts*2)[:,newaxis]
-    print(shape(mybounds))
-    print(mybounds)
+    logger.info(strm(shape(mybounds)))
+    logger.info(strm(mybounds))
```
jmfranck, line 62: `logging.info` should work (see comment above)

jmfranck: I would run `grep -rie 'init_logging' *` inside the `proc_scripts` subdirectory to find any files where this logging thing is an issue, and then fix it.

proc_scripts/load_data.py
```diff
@@ -3,22 +3,60 @@
...
+logger=init_logging('info')
 #to use type s = load_data("nameoffile")
-def proc_bruker_deut_IR_withecho_mancyc(s):
-    raise RuntimeError("this is where postprocessing would be implemented -- not implemented yet")
+def proc_bruker_deut_IR_withecho_mancyc(s,fl=None):
+    logger.info(strm("this is the 90 time"))
+    if fl is not None:
+        fl.next('raw data')
+        fl.image(s.C.setaxis(
+'indirect','#').set_units('indirect','scan #'))
```
jmfranck, line 14: weird indentation -- highlight, then press `=`

proc_scripts/load_data.py
```diff
@@ -3,22 +3,60 @@
...
+    s.reorder(['indirect','t2'], first=False)
+    if fl is not None:
+        s_forplot = s.C
+        fl.next('FT')
+        fl.image(s_forplot.C.setaxis('indirect','#').set_units('indirect','scan #'))
+    if fl is not None:    
+        fl.next('time domain (all $\\Delta p$)')
+        s_forplot.ift('t2')
+        fl.image(s_forplot.C.setaxis('indirect','#').set_units('indirect','scan #'))
+    if fl is not None:
```
jmfranck, line 33: you have three blocks w/ the same conditional -- just combine.

proc_scripts/load_data.py
```diff
@@ -3,22 +3,60 @@
...
+    s.setaxis('ph2',r_[0:4.]/4) #setting values of axis ph1 to line up
     s.setaxis('indirect', s.get_prop('vd'))
 #titling to coherence domain
     s.ft('t2',shift=True) #fourier transform
     s.ft(['ph1','ph2']) #fourier transforming from phase cycle dim to coherence dimension
     s.reorder(['indirect','t2'], first=False)
     if fl is not None:
         s_forplot = s.C
+        s_forplot.setaxis('indirect','#')
+        s_forplot.set_units('indirect','scan #')
```
jmfranck, line 59: I would string the methods into a single command line -- I say this in case you don't know that you can do that.  Otherwise, this could be a matter of personal style

`s.C.setaxis(....).set_units(...)`

proc_scripts/load_data.py
```diff
@@ -49,26 +87,143 @@ def proc_spincore_CPMG_v1(s, fl=None):
...
-    t2_axis = linspace(0,acq_time_s,nPoints)
+    t2_axis = np.linspace(0,acq_time_s,nPoints)
+    tE_axis = r_[1:nEchoes+1]*twice_tau
     s.setaxis('nScans',r_[0:nScans])
     s.chunk('t',['ph1','tE','t2'],[nPhaseSteps,nEchoes,-1])
-    # OK, so I made some changes and then realized that we need to assume that
-    # the pulse sequence correctly balances the evolution between 2*p90_s/pi
-    # (cavanagh chpt 3 this is the evolution during the 90 -- I'm not positive
-    # if my expression is correct or not -- please do check/change, and leave
-    # this comment in some form) and the center of the 180 pulse appropriately
```
jmfranck, line 59: don't delete explanatory comments like this -- also read the comment.

proc_scripts/load_data.py
```diff
@@ -214,13 +409,37 @@ def proc_DOSY_CPMG(s):
...
+        'spincore_nutation_v2':proc_nutation_amp,        
         'spincore_ODNP_v1':proc_spincore_ODNP_v1,
-        'square_wave_capture_v1':proc_square_wave_capture,
-        'DOSY_CPMG_v1':proc_DOSY_CPMG}
+        'spincore_var_tau_v1':proc_var_tau,
+        'square_wave_capture_v1':proc_capture,
+        'DOSY_CPMG_v1':proc_DOSY_CPMG,
+        'ESR_linewidth':proc_ESR_linewidth,
+
+}
```
jmfranck, line 444: add a docstring at the beginning of the file that describes your convention (everything should be in the Δp domain rather than the φ domain -- is it in the t2 domain or the ν2 domain -- also, anyone who modifies this isn't allowed to do anything that's not invertible (anything that throws out information, like slicing, filtering, etc.))

jmfranck (commented):
I have some 2 files left to look at, so if you go through and address all my comments, I think we're doing pretty good!

proc_echo_mw.py
```diff
@@ -14,29 +17,29 @@
...
     s.ift('t2') # inverse fourier transform into time domain
     logger.debug(strm("THIS IS THE SHAPE"))
     logger.debug(strm(ndshape(s)))
-    s = slice_FID_from_echo(s)['t2':(None,0.05)]
-    # visualize time domain after filtering and phasing 
-    fl.side_by_side('time domain (after filtering and phasing)\n$\\rightarrow$ use to adjust time range', s, time_range)
-    s =s['t2':time_range] # slices out time range along t2 axis
+    fl.next('time domain')
+    fl.image(s.C.setaxis('power','#').set_units('power','scan #'))
+    s = slice_FID_from_echo(s,max_t=max_t,fl=None)    # visualize time domain after filtering and phasing
```
jmfranck, line 36: there is no doc string for `slice_FID_from_echo`, so I don't know what it does

proc_echo_mw.py
```diff
@@ -14,29 +17,29 @@
...
     #{{{apodizing and zero fill
     fl.next('apodize and zero fill')
-    R = 5.0/(time_range[-1]) # assume full decay by end time
-    s *= exp(-s.fromaxis('t2')*R)
+    R = 5.0/(max_t) # assume full decay by end time
```
jmfranck, line 39: part of roll-back noted above

proc_scripts/slice_FID_from_echo.py
```diff
@@ -2,8 +2,8 @@
 from pyspecdata import *
 from sympy import symbols
 
-def slice_FID_from_echo(s, ph1=1, ph2=-2):
-    fl = fl_mod() 
+
+def slice_FID_from_echo(s, max_t=None, ph1=1, ph2=-2, fl=None):
```
jmfranck, line 6: I would actually say `pathway={'ph1':1, 'ph2':-2}` then below you can do something like this to select the appropriate coherence pathway (obviously replacing later occurrences of `s` with `myslice`
```
myslice = s
for k,v in pathway.items():
    myslice = s[k,v]
```

proc_spincore_IR.py
```diff
@@ -0,0 +1,225 @@
+from pyspecdata import *
```
jmfranck, line 1: two problems here -- this seems to be added to the root directory.

It's also implying there are different files to process the inversion recovery for the spincore vs. bruker, which is not what we want.

jmfranck: you want one file that you test on both bruker and spincore data

jmfranck (commented):
OK! If you deal with the comments, we should be good to go.

proc_spincore_IR.py
guinneak198: done

q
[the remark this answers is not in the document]
guinneak198: hopefully this worked....

proc_T1CPMG.py
jmfranck: just add a comment in the code to explain this?

proc_CPMG.py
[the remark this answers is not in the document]
jmfranck: yes, but rather than specifying this explicitly, you can check to see if there are 1 or 2 phase cycling dimensions, whether or not there is an nScans dimension, and then respond accordingly

proc_scripts/correlation_alignment.py
jmfranck: not quite -- I added a line to prevent the break if a minimum number of iterations hasn't been reached

proc_scripts/phasing.py
jmfranck: yes, but the point should be that it should be a single number -- what you're describing should fail

proc_nutation.py
jmfranck: my best guess is that you have more dimensions than it expects, and you end up running the zeroth order phasing for many many points

proc_nutation.py
jmfranck: which one?

proc_nutation.py
jmfranck: nice!

proc_scripts/CPMG_phasing.py
jmfranck: this has a very different purpose -- the hermitian function test is the *find* the center of the echo.  This routine is essentially an alternative to and FID slicing -- here we are trying to slice a symmetric echo (same amount of negative time as positive time).  Resolve this once you read it.

proc_scripts/load_data.py
jmfranck: you missed this one -- I just fixed it

proc_scripts/load_data.py
jmfranck: missed -- I resolved

examples/proc_var_tau.py
```diff
@@ -0,0 +1,37 @@
...
+        NV = 250e-6*55.4*2*N_A # 400 μL, 55.4 M water molecs, 2 spins/molec
+        nu0 = frequency 
+        LambdaNMR = 1.55e-4 # 1 G/√W
+        I = 0.5
+        Vsignal = LambdaNMR * NV * (gammabar_H*2*pi) * I * (I+1) * (hbar*2*pi*nu0)**2 * sqrt(50)
+        Vsignal /= 3 * k_B * (273+20)
+        axhline(y=Vsignal,alpha=0.2)
+        logger.info(strm("Vsignal expected",Vsignal))
+        fl.plot(Vsignal, 'x', label='theoretical signal at 0')
+        fl.show();quit()
```
jmfranck, line 37: `fl.show()` not needed if you use a `with` block. (and in general, we shouldn't have `quit` in anything)

proc_CPMG.py
```diff
@@ -3,29 +3,75 @@
...
+            expno=nodename, postproc=postproc, lookup=postproc_dict, fl=fl)
+    s.ift('t2')
+#{{{Spincore data requires some reordering of dimensions while bruker data does not.
+    #For this reason we make spincore an argument above for the proper processing
+    if spincore:
+        s = s['ph1',1] #the spincore version only has one phasing dimension 'nPhaseSteps'
+        s.mean('nScans')
+        s.reorder('t2',first=True)
+    else:
+        s = s['ph2',-2]['ph1',1]
```
jmfranck, line 39: still a problem with the spincore here -- as noted in other comments, just inspect the phase to see what phase cycling dimensions are present.

proc_capture_nutation.py
```diff
@@ -0,0 +1,66 @@
+from pyspecdata import *
+from pylab import *
+from sympy import symbols, latex, Symbol
```
jmfranck, line 3: shouldn't this be in the examples folder?

proc_chirp.py
```diff
@@ -1,108 +1,42 @@
...
-        fl.next('$S_{11}$ : phase')
-        fl.plot((ratio).angle/pi, '.', c='blue', **plot_params)
-        expno += 1 
-    fl.next('$S_{11}$ : phase')
-    gridandtick(gca())
-    xlim(10,20)
-    ylim(-1.0,1.0)
-    axvline(14.4289,linestyle=':', color='black')
-    #savefig('../scriv/first_year/fig/phase.png')
-    fl.next('$S_{11}$ : analytic amplitude')
```
jmfranck, line 103: so, there still should be a plot that's labeled $S_{11}$, b/c that's the goal here.

proc_nutation.py
```diff
@@ -1,53 +1,86 @@
...
+    s.ft('t2')
+    fl.image(s)
+    s.ift('t2')
     #{{{ selecting coherence and convolving
     s = s['ph2',0]['ph1',1]
-    fl.next('select $\\Delta p$ and convolve')
-    s.convolve('t2',50)
+    fl.next('select $\\Delta p$')
+    if 'p90' in s.dimlabels:
+        s.convolve('t2',50)
```
jmfranck, line 53: it's weird that this convolution depends on whether it's variable amplitude vs. variable pulse length

proc_nutation.py
```diff
@@ -1,53 +1,86 @@
...
     fl.next('sliced')
     fl.image(s)
     #}}}
-    
+    if 'amp' in s.dimlabels:
+        s.setaxis('amp',lambda x:x*plen)
+        s.set_units('amp','s')
+        ind_dim = '\\tau_p a'
+        s.rename('amp',ind_dim)
+    if 'p_90' in s.dimlabels:
```
jmfranck, line 66: elif

jmfranck (commented):
after working through comments some, went backed and looked at code that got touched.

proc_scripts/load_data.py
guinneak198: attempted

proc_nutation.py
guinneak198: no more convolution

proc_spincore_IR.py
jmfranck: so what is supposed to process the spincore data now?  I would have guessed `proc_IR.py`, but it doesn't seem to have any flexibility in terms of which phase cycling dimensions are present.

proc_CPMG.py
```diff
@@ -53,10 +53,12 @@
 'tE','#').set_units('tE','scan #'))
     s = s['t2':f_range]
     s = s.C.sum('t2')
+    # {{{ (FROM REVIEW) what is going on here???
     if spincore:
         CPMG=s
     else:
         CPMG = s['indirect',-1]
+    # }}}
```
jmfranck: what's going on here?

proc_nutation.py
```diff
@@ -25,8 +25,8 @@
     # in particular -- if you don't do this before convolution, the
     # convolution doesn't work properly!
     s.ift('t2')
-    # {{{ centering of data using hermitian function test
-    else:
+    if 'amp' not in s.dimlabels:
+        # {{{ centering of data using hermitian function test
```
jmfranck: make sure you've checked this program -- the hanging else you had would have given a syntax error.

proc_chirp.py
```diff
@@ -1,108 +1,42 @@
...
+from pylab import *
+do_slice = False # slice frequencies and downsample -- in my hands, seems to decrease the quality of the fit 
+standard_cost = False # use the abs real to determine t=0 for the blip -- this actually doesn't seem to work, so just use the max
+show_transfer_func = False # show the transfer function -- will be especially useful for processing non-square shapes
+logger = init_logging('info')
+# 2to3 JF 1/31
+
+fl = figlist_var()
+# {{{ load data, set units, show raw data
+with fl as figlist_var():
```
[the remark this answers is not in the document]
jmfranck: its `with classname() as myinstance:` -- the other way around.  If you are using a with block, then you don't use line 14 here.

proc_linewidth.py
```diff
@@ -0,0 +1,152 @@
+from pyspecdata import *
+from proc_scripts import *
+from proc_scripts import postproc_dict
+import symfit as s
+import pickle,os
+from pylab import ndarray
+from symfit import Parameter, Variable, parameters, variables, Fit, Model
+from symfit.core.minimizers import MINPACK
+from symfit.contrib.interactive_guess import InteractiveGuess
```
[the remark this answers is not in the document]
jmfranck: that's fine -- the point is just that in a python package, the only thing in the root directory should be the README file, the license file, and the setup file (possibly other things that I'm missing).  Eventually, we will want to make these better examples, but that's the location that they fit best.  Just thumbs up and I will resolve.

proc_nutation.py
jmfranck: not resolved, but will start a new comment

proc_scripts/CPMG_phasing.py
jmfranck: I edited that in 8f2de892

proc_T1CPMG.py
```diff
@@ -50,12 +50,16 @@
         s.ift('t2')
         if flat_echo:
            s['t2',16]=0
+           # {{{ (FROM REVIEW) find echo center no longer exists!
            avg_center=find_echo_center(s,fl=fl)
+           # }}}
```
jmfranck: probably want to replace this with hermitian function?

proc_T1CPMG.py
```diff
@@ -50,12 +50,16 @@
...
         else:    
             centers = []
             for j in range(ndshape(s)['indirect']):
                 s_slice = s['indirect',j]
+                # {{{ (FROM REVIEW) find echo center no longer exists!
                 this_center = find_echo_center(s_slice,fl=fl)
+                # }}}
```
jmfranck: same here?

proc_nutation.py
```diff
@@ -27,15 +27,18 @@
     s.ift('t2')
     if 'amp' not in s.dimlabels:
         # {{{ centering of data using hermitian function test
+        # {{{ (FROM REVIEW) I still don't understand why this is only done for
+        #                   an amplitude-varied only
         best_shift = hermitian_function_test(s['ph2',0]['ph1',1])
         logger.info(strm("best shift is",best_shift))
         s.setaxis('t2', lambda x: x-best_shift).register_axis({'t2':0})
+        # }}}
```
jmfranck: see comment

proc_nutation.py
guinneak198: when I attempt to apply the hermitian function test to the amp varied data (I also reworded your comment to correct that it is the amp varied where hermitian is not applied) I get this error
`Traceback (most recent call last):
  File "proc_nutation.py", line 26, in <module>
    best_shift = hermitian_function_test(s['ph1',1]['ph2',0])
  File "C:\Users\Alex Guinness\gitstuff\proc_scripts\proc_scripts\phasing.py", line 223, in hermitian_fu
nction_test
    s.ift('t2')
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\fourier\ift.py", line 179, in ift
    self._ft_shift(thisaxis,p2_pre)
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\fourier\ft_shift.py", line 77, in _ft_shif
t
    newdata = np.empty_like(self.data)
  File "<__array_function__ internals>", line 5, in empty_like
MemoryError: Unable to allocate 14.6 GiB for an array with shape (199, 4096, 1200) and data type complex
128` 
I don't understand why unless there is just a boat load of more data in the amp varied compared to the p90 varied? I have left my trouble shooting code in place so you can get an idea of how I went about it and maybe I am missing something or going about it the wrong way?

proc_scripts/CPMG_phasing.py
guinneak198: edited to be more accurate

proc_spincore_IR.py
guinneak198: proc_IR.py does both the main reason for the two files was the preprocessing was different and loading in the files was different (back when I first started this adventure and made a million scripts instead of compiling). shouldn't the coherence pathway be the same for any IR experiment? I'm a bit confused on your question. As far as I'm aware they should be processed the same (slice the freq range, apply phase corrections/clock correction if needed, align data, integrate and fit)....am I missing something that is different between the two?

proc_CPMG.py
guinneak198: so for my bruker data for some reason there still exists a dimension ['indirect',1]. Now this might be a result of us futzing around trying to find a good pp and maybe I am processing bad bruker data... When I do this however (CPMG=s['indirect',-1]) it does produce a okay decay curve and is able to fit.

proc_nutation.py
guinneak198: I've run now and it does work! good catch!

guinneak198 (commented):
okay went through 3/24/21 and hopefully this resolves some of the issues! everything I commented on/thumbs upped runs on my end!

proc_spincore_IR.py
jmfranck: OK -- could be I'm mixing things up, then.  I thought that on the bruker you were doing more or less phase cycling than on the spincore.  For the coherence selection to be the same, the phase cycling needs to be the same.

proc_Hahn_echo.py
```diff
@@ -2,31 +2,26 @@
...
-    fl.plot(s.real,alpha=0.4)
-    #fl.plot(s.imag,alpha=0.4)
-    #fl.plot(abs(s),':',c='k',alpha=0.4)
+    fl.next('raw data-real')
+    fl.plot(s.real,alpha=0.4,label='real')
+    fl.plot(s.imag,alpha=0.4,label='imaginary')
     s.ft('t',shift=True)
+    s.mean_all_but('t')
     s.convolve('t',10)
-    fl.next('Absolute value -- FT')
```
jmfranck, line 24: If you're going to do a convolution, you should mention it in the plots.  I'm also not a fan of doing a convolution like this -- rather choose your apodization carefully in the time domain.  We sometimes convolve right when plotting, but that's a convenience thing that has to do w/ the display, and isn't intended as a permanent code alteration, like here.

proc_IR.py
```diff
@@ -1,77 +1,273 @@
...
+    if 'indirect' in s.dimlabels:
+        s.rename('indirect','vd')
+    fl.next('time domain cropped log')
+    fl.image(s.C.setaxis('vd','#').set_units('vd','scan #'))
+    #{{{centering, hermitian function test and zeroth order phasing
+    rough_center = abs(s).convolve('t2',10).mean_all_but('t2').argmax('t2').item()
+    s.setaxis(t2-rough_center)
+    fl.next('rough centering')
+    fl.image(s.C.setaxis('vd','#').set_units('vd','scan #'))
+    #}}}
```
jmfranck: why do we need to do rough centering?

proc_IR.py
```diff
@@ -1,77 +1,273 @@
...
+        fl.next('clock correction')
+        fl.plot(s_clock,'.',alpha=0.7)
+        clock_corr = s_clock.argmax('clock_corr').item()
+        pyplot.axvline(x=clock_corr, alpha=0.5, color='r')
+        s *= np.exp(-1j*clock_corr*s.fromaxis('vd'))
+        fl.next('after auto-clock correction')
+        s.ft(['ph1','ph2'])
+        fl.image(s.C.setaxis('vd','#'))
+        s.ift('t2')
+        #}}}
```
jmfranck: so, is this *all* clock-correction related?

proc_IR.py
```diff
@@ -1,77 +1,273 @@
...
+    'vd','#').set_units('vd','scan #'),ax=ax2)
+        #}}}
+        #{{{ pre-alignment
+        s_final = s.C
+        s_before = s['vd',:2]
+        s_after = s['vd',3:]
+        s_after.ift(['ph2','ph1'])
+        s_after.ift('t2')
+        s_after.ft('t2')
+        fl.next('before alignment - FT')
```
jmfranck: I don't understand why this whole block is spincore-related.  The *only* thing that I've seen that should be spincore specific is the clock correction.

proc_IR.py
```diff
@@ -1,77 +1,273 @@
...
+        fl.basename = None
+        s_before.ift('t2')
+        s_before *= np.exp(-1j*2*pi*opt_shift_before*s_before.fromaxis('t2'))
+        fl.next('out of correl alignment before')
+        fl.image(s_before)
+        s_after.ift('t2')
+        s_after *= np.exp(-1j*2*pi*opt_shift_after*s_after.fromaxis('t2'))
+        fl.next('out of correl alignment after')
+        fl.image(s_after.C.setaxis(
+    'vd','#').set_units('vd','scan #'))
```
jmfranck: this is not the proper way to indent a continued line, and I actually don't understand how this happens.  If your using gvim, vscode, etc, it would not naturally do this.

If using gvim, fix by highlighting and pressing `=`.  I am fixing this in several places rather than commenting on each one.

proc_T1CPMG.py
```diff
@@ -1,137 +1,144 @@
...
 rcParams["savefig.transparent"] = True
 # {{{ input parameters
-#clock_correction = 1.785
-clock_correction = 0
-nodename = 'signal'
 filter_bandwidth = 5e3
 t2 = symbols('t2')
+test_for_flat_echo = False# test for flat echo and exit
+write_h5 = False#writes the hdf5 file
+read_h5 = True #reads the completed hdf5 file
```
jmfranck: with comments, but a space both before the `#` if it's a continuation like this, as well as and always put a space between the `#` and the comment

proc_nutation.py
```diff
@@ -1,53 +1,86 @@
...
-    s.ft(['ph2','ph1'])
-    # }}}
-    
+    s.ift('t2')
     # {{{ centering of data using hermitian function test
-    best_shift = hermitian_function_test(s['ph2',0]['ph1',1])
-    logger.info(strm("best shift is",best_shift))
-    s.setaxis('t2', lambda x: x-best_shift).register_axis({'t2':0})
+    if 'p90' in s.dimlabels:
+        best_shift = hermitian_function_test(s['ph2',0]['ph1',1])
```
jmfranck, line 44: this line shouldn't be dependent on whether it's amp or pulse length varied!!

proc_scripts/CPMG_phasing.py
```diff
@@ -2,29 +2,33 @@
...
     s.register_axis({axis:0})
-    s /= zeroth_order_ph(s[axis:0])
+    logging.debug(strm("after register axis, t2 axis is", s.getaxis(axis)))
+    s /= zeroth_order_ph(s[axis:0],fl=fl)
     time_bound = min(abs(s.getaxis(axis)[r_[0,-1]]))
+    logging.debug(strm("time bound is",time_bound))
+    axis_before = ndshape(s)[axis]
     s = s[axis:(-time_bound,time_bound)]
-    assert isclose(s.getaxis(axis)[0],-s.getaxis(axis)[-1]),"echo is not symmetric! you are using the wrong code!! (first point is %g, last point %g, dwell time %g, and time_bound=%g"%(s.getaxis(axis)[0],s.getaxis(axis)[-1],diff(s.getaxis(axis)[r_[0,1]]).item(),time_bound)
-
```
jmfranck, line 27: please don't ever remove assertions!!!!!!!!!!!!!!!!!!!!

jmfranck: OK, I see this just got moved, sorry!

proc_scripts/fitting.py
```diff
@@ -1,22 +1,25 @@
...
     if not s.get_ft_prop(direct):
         raise ValueError("Your data should be in the frequency domain!")
-    curve = s[direct:f_range].sum(direct).real
+    curve = s.real
     return fitdata(curve)
 def _fitcurve_final(f,whichrate,guess):
-    logger.info(strm("Functional form", f.functional_form))
-    logger.info(strm("Functional form", f.functional_form))
+    logging.info(strm("Functional form", f.functional_form))
+    logging.info(strm("Functional form", f.functional_form))
```
jmfranck, line 13: I'm a bit confused why there is a frequency range here that gets deleted -- what's up w/ that?

proc_chirp.py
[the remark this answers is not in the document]
jmfranck: this is not changed for some reason?

proc_CPMG.py
jmfranck: yes, but then the solution is
`if 'indirect' in s.dimlabels and ndshape(s)['indirect']==1: ....`

proc_T1CPMG.py
jmfranck: in general, when I write the comment as "FROM REVIEW" feel free to delete it when you edit the code.

proc_nutation.py
jmfranck: you are right about your correction to my question here "p90-varied only" but the question still stands -- why are you only doing this for a p90-varied, and not for amp-varied?!

examples/proc_var_tau.py
```diff
@@ -1,3 +1,4 @@
+"""PLEASE PROVIDE A DOCSTRING HERE TO DESCRIBE WHAT THIS FILE DOES"""
 from pylab import *
```
jmfranck: please do this

proc_nutation.py
```diff
@@ -37,7 +37,8 @@
     # {{{ do the centering before anything else!
     # in particular -- if you don't do this before convolution, the
     # convolution doesn't work properly!
-    else:
+    s.ift('t2')
+    if 'amp' not in s.dimlabels:
```
jmfranck: this is what I was referring to as the "hanging else" -- it's the same level of indentation with the previous line -- it's not paired with an if statement.

It actually *is* a syntax error, but you don't see it b/c you quit previous to it!

proc_nutation.py
```diff
@@ -33,7 +33,6 @@
         fl.image(s)
         s.ift('t2') # make sure everything is in the same domain
         #}}}
-    fl.show();quit()
```
jmfranck: this guy was a really big problem -- it allowed for problematic code that came after it.

proc_CPMG.py
```diff
@@ -3,29 +3,78 @@
...
+    logger.info(strm("Functional Form", fit_CPMG.functional_form))
+    logger.info(strm("Functional Form", fit_CPMG.functional_form))
+    fit_CPMG.fit()
+    logger.info(strm("output:",fit_CPMG.output()))
+    logger.info(strm("latex:",fit_CPMG.latex()))
+    T2 = 1./fit_CPMG.output('R_2')
+    fl.next('CPMG fit of T1CPMG for free D2O')
+    fl.plot(CPMG,'o',label='data')
+    fl.plot(fit_CPMG.eval(100),label='fit')
+    logger.info(strm("T2 IS:",T2))
```
jmfranck, line 77: wasn't the point of the `decay` function to do a bunch of this?  I remember there being some discussion around that,  but can't remember what happened.

proc_T1CPMG.py
```diff
@@ -1,137 +1,153 @@
...
+        #}}}
+        #}}}
+        #{{{save to hdf5 file
+        s.name(searchstr) # use searchstr as the node name withing the HDF5 file
+        s.hdf5_write(h5_name, directory=getDATADIR(h5_dir))
+        logger.info("saving as hdf5 file with shape:",strm(ndshape(s)))
+        #}}}
+    #if read_h5 is True there must already be a written h5 file for the file that 
+    #is to be plotted/imaged that already exists. 
+    if read_h5:
```
jmfranck, line 128: what isn't this just an if/else?

proc_noise.py
```diff
@@ -69,13 +68,10 @@
     fl.next('noise -- semilog', legend=True)
     fl.plot(noise['t2':frq_slice], plottype='semilogy',alpha=0.05, color=c,
             human_units=False)
-    #noise.convolve('t2',500) # convolution is just to provide a running
-    # average, to make it clear where the average noise baseline is at (remember
-    # we are doing this *after* taking the abs)
```
jmfranck, line 74: so, this seems bad -- why did this get commented out in the first place?

proc_noise.py
```diff
@@ -57,8 +58,6 @@
     signal = s['t2':(-1000,1000)]['ph1',1]['ph2',-2]
     c = next(new_colors)
     signal.ift('t2')
-    #fl.plot(signal.real, alpha=0.5, color=c, label=label_str)
-    #fl.plot(signal.imag, ':', alpha=0.5, color=c, label=label_str)
     fl.plot(abs(signal)['t2':(None,0.2)], linewidth=3, color=c, alpha=0.25,
```
jmfranck, line 61: why not uncomment these as well?

proc_noise.py
jmfranck: I should elaborate -- with the noise processing, I would argue that taking the convolution after the abs is actually essential to getting a good noise PSD within a reasonable number of scans.

we are deliberately sacrificing frequency resolution (don't care about the frequency resolution of our noise very much!) for a good SNR curve (odd to use that expression, but there is an SNR to the noise density measurement itself).

jmfranck:
just merge bruker processing, since it's become the active branch

PR #43, opened 2021-06-01
suggestions for temporary.py

jmfranck:


proc_scripts/phasing.py
```diff
@@ -201,12 +201,12 @@ def hermitian_function_test(s, down_from_max=0.5, rel_shift=1.5,shift_points=120
     data_for_peak /= max_val
     pairs = data_for_peak.contiguous(lambda x: abs(x) >
             down_from_max)
-    peak_location = pairs[0,:] #gives the longest pair
-    peak_center = peak_location.mean()
+    peak_bounds = pairs[0,:] #gives the longest pair
+    peak_center = data_for_peak['t2':(0.1e-3,None)].argmax('t2').item()
```
guinneak198, line 205: this makes sense seeing it but definitely was having trouble visualizing it as one line.

proc_scripts/phasing.py
```diff
@@ -201,12 +201,12 @@ def hermitian_function_test(s, down_from_max=0.5, rel_shift=1.5,shift_points=120
...
     s.setaxis('t2',lambda x: x-peak_center)
     s.register_axis({'t2':0})
     logging.debug(strm("closest to 0", s.getaxis('t2')[np.argmin(abs(s.getaxis('t2')-0))]))
-    max_shift = np.diff(peak_location).item()/2
+    max_shift = np.min(abs(peak_center-peak_bounds))
```
guinneak198, line 209: wouldn't this return two values? or does the min take care of that?

proc_scripts/phasing.py
```diff
@@ -304,15 +304,19 @@ def hermitian_function_test(s, down_from_max=0.5, rel_shift=1.5,shift_points=120
         fig, (ax1,ax2,ax3) = plt.subplots(3,1)
         fl.next('mirror tests', fig=fig)
         def real_imag_mirror(forplot, ax):
-            l = fl.plot(forplot.real, alpha=0.5, ax=ax, label='real')
-            fl.plot(forplot.C.setaxis('t2', lambda x: -x), color=l[-1].get_color(),alpha=0.2,ax=ax)
+            l = fl.plot(forplot.real, alpha=0.5, ax=ax, label='real',
+                    human_units=False)
```
guinneak198, line 308: when is a proper time to use human_units=False vs not having that addition? I know there are times where we do and don't want this but what are those times?

proc_scripts/phasing.py
```diff
@@ -304,15 +304,19 @@ def hermitian_function_test(s, down_from_max=0.5, rel_shift=1.5,shift_points=120
...
             l=fl.plot(forplot.imag,alpha=0.5,ax=ax)
             fl.plot(-forplot.imag.C.setaxis('t2',lambda x: -x),color=l[-1].get_color(),
-                    alpha=0.2,ax=ax)
+                    alpha=0.2,ax=ax,
+                    human_units=False)
             residual = forplot['t2':(-max_shift,max_shift)]
             if ndshape(residual)['t2'] % 2 == 0:
                 residual = residual['t2',:-1]
-            fl.plot(abs(residual-residual['t2',::-1].runcopy(np.conj))*20, ax=ax, label='residual x 20')
+            fl.plot(abs(residual-residual['t2',::-1].runcopy(np.conj))*20, ax=ax, label='residual x 20',
```
guinneak198, line 318: why times 20?

temporary.py
```diff
@@ -1,3 +1,6 @@
+generate_image_files = False
+if generate_image_files:
+    import os; os.environ['pyspecdata_figures']='latex'
```
guinneak198, line 3: why?

temporary.py
```diff
@@ -13,21 +16,23 @@
...
 fl.next("time")
 fl.image(d.C.setaxis("power", "#").set_units("power", "scan #"))
 d.register_axis({'t2':0})
 #d.ft('t2')
 #d = d['t2':(-3e3,3e3)]
 #d.ift('t2')
-best_shift, window_size = hermitian_function_test(d["ph1", 1]["ph2", 0], fl=fl)
+best_shift, window_size = hermitian_function_test(d["ph1", 1]["ph2", 0])
+logger.info(strm("hermitian best shift",best_shift))
 best_shift = 0.33e-3
```
guinneak198, line 28: doesn't this just reset the best_shift so you aren't using the one the hermitian function test returns?

temporary.py
```diff
@@ -37,17 +42,27 @@
...
 d *= np.exp(-1j*2*pi*opt_shift*d.fromaxis('t2'))
 d.ft('t2')
 fl.basename = None
-fl.next(r'after correlation alignment, $\varphi$ domain')
-fl.image(d)
-fl.show()
+fl.image(d['t2':(-500,500)].C.setaxis('power','#').set_units('power','scan #'))
+fl.next(r'after masked correlation alignment, $\varphi$ domain')
+opt_shift,sigma = correl_align(d, indirect_dim='power',
+        ph1_selection=0,ph2_selection=0,sigma=30e3/2.35)
```
guinneak198, line 60: where did the choice of the sigma selection come from? I know the divided by 2.35 has to do with the linewidth and such but what is the 30e3?

guinneak198 (commented):
went through and added questions at parts I didn't understand

proc_scripts/phasing.py
jmfranck: the latter

proc_scripts/phasing.py
jmfranck: we shouldn't really have to use them ever -- it's a bit of a bug that hasn't been fixed.  As is, the only reasonable time is, like here, when you have two sets of axis labels that are very different sizes -- so that it decides the units it would apply to each are different

proc_scripts/phasing.py
jmfranck: to make it bigger, because it should be very very small

temporary.py
jmfranck: I'm not using it, b/c even with the changes, the result is not great, and while I think my changes are in the right direction, I didn't have time to debug further

temporary.py
jmfranck: I realized this is a typo -- I mean it to be 30 Hz -- it's interesting that the result here is better I'm playing with it now.

jmfranck: [see here](https://jmfrancklab.slack.com/archives/CLMMYDD98/p1622645958002900)

proc_scripts/phasing.py
guinneak198: so to be clear. we are saying take the minimum of the two values that are returned when we take the difference between the peak bounds and the center?

proc_scripts/phasing.py
guinneak198: "it" being the residual which is equal to forplot['t2':(-max_shift,max_shift)]? and 20 is just an arbitrary choice?

guinneak198:
I'm leaving this open as I still have some questions but once answered should I merge or leave that to you?

proc_scripts/phasing.py
jmfranck: yes

proc_scripts/phasing.py
jmfranck: yes

temporary.py
jmfranck: setting this to true will generate image files for posting to slack -- you need to create the `auto_figures` directory inside the current directory first

temporary.py
jmfranck: by eye, that's what it looked like to me

guinneak198:
good to merge?

PR #47, opened 2021-06-16
Sam fixes

bett8500:
Fixing imports (swap `from proc_scripts import *` for `from pyspecProcScripts import *` ) for process_enhancement.py and process_IR.py so that they work. Putting the freq_range and t_range back into the loop for proc_echo_mw.py to set separate ranges for different files.


proc_IR.py
```diff
@@ -8,14 +8,9 @@
...
-           'FIR_34dBm','spincore_IR_v1',
-           (0.1e3,0.345e3),(None,50e-3),True,False,False),
-        # ('210322_water_control_FIR_noPower','inv_rec','signal','spincore_IR_v1',
-        #    (-0.09e3,-0.06e3),(None,83e-3),False,False),
-        #('210517_4OHTempo_TempControl_probe_FIR_34dBm.','odnp_nmr_comp/inv_rec','signal','spincore_IR_v1',
-        #    (-0.4e3,0.4e3),(None,20e-3),False,False),
-        #('w3_201111','test_equip',2,'ag_IR2H',(-600,600),(0,None),True)
+       ('210615_S175R1a_pR_DDM_ODNP','odnp',
+           'FIR_0dBm','spincore_IR_v1',
+           (-0.5e3,0.5e3),(None,50e-3),True,False,False),
```
jmfranck, line 13: this file should be an example, and this is pointing it to a different file, so it's not really a fix of the underlying code.

proc_echo_mw.py
```diff
@@ -9,7 +9,8 @@
 #from pylab import *
 #from sympy import exp as s_exp
 #from scipy.optimize import leastsq,minimize,basinhopping,nnls
-
+import os
+os.chdir('C:/Users/saman/proc_scripts/')
```
jmfranck, line 13: this would only be necessary if you didn't do `python setup.py develop`

proc_echo_mw.py
```diff
@@ -19,13 +20,15 @@
...
-for filename,nodename,file_location,postproc in [
-        ('210609_TEMPOL_100mM_cap_probe_DNP','enhancement',
-            'ODNP_NMR_comp/test_equipment','spincore_ODNP_v1'),
+for filename,nodename,file_location,postproc,freq_range,t_range in [
+        ('210616_F195R1a_pR_DDM_ODNP','enhancement',
+            'odnp','spincore_ODNP_v1',
+            (-0.4e3,0.4e3),(0,50e-3)),
+#        ('210615_S175R1a_pR_DDM_ODNP','enhancement',
+#            'odnp','spincore_ODNP_v1'),
+#            (-5e3,5e3)),
```
jmfranck, line 31: seems to be retooling for a different set of data

pyspecProcScripts/third_level/process_IR.py
```diff
@@ -5,7 +5,7 @@
 import numpy as np
 import matplotlib.pyplot as plt
 from sympy import symbols, latex, Symbol
-from proc_scripts import *
+from pyspecProcScripts import *
 t2 = symbols('t2')
```
jmfranck, line 9: this is OK, but it would be better not to have an import *

pyspecProcScripts/third_level/process_enhancement.py
```diff
@@ -1,8 +1,8 @@
 from pyspecdata import *
 from scipy.optimize import leastsq,minimize,basinhopping,nnls
-from proc_scripts import *
-from proc_scripts import postproc_dict
-from proc_scripts.correlation_alignment_ODNP import correl_align
+from pyspecProcScripts import *
+from pyspecProcScripts import postproc_dict
+from pyspecProcScripts.correlation_alignment_ODNP import correl_align
```
jmfranck, line 5: same as previous

jmfranck:
see comments as to why this was closed w/out merge

PR #52, opened 2021-06-25
correlation PR

guinneak198:
not sure if this will run as it's from before we added the fake_data function.... I had to merge in master for the only_correlation branch but that also pulled in the 90 something other files with doc_src so I had to go back and pull my changes into this new branch....


examples/correlation_alignment_example.py
```diff
@@ -0,0 +1,84 @@
...
+#{{{generate fake data
+fl = figlist_var()
+#this generates fake clean data with a T2 of 0.2s
+# amplitude of 23 (arbitrary), offset of 300 Hz,
+# FWHM 10 Hz
+f_range = (-400,400)
+t2, td, vd, ph1, ph2 = s.symbols('t2 td vd ph1 ph2')
+echo_time = 1e-3
+data = fake_data(
+        23*(1 - 2*s.exp(-vd / 0.2)) * s.exp(+1j*2*s.pi*100*(t2) - abs(t2)*50*s.pi),
```
jmfranck, line 25: I would put everything inside a loop, then loop over two functions (1) this -- the inv. rec. (2) an enhancement curve

guinneak198: I'm confused... loop to make multiple datasets?

examples/correlation_alignment_example.py
```diff
@@ -0,0 +1,84 @@
...
+data /= sqrt(ndshape(data)["t2"]) * data.get_ft_prop('t2','dt')
+# }}}
+fl.next("fake data -- freq domain")
+fl.image(data)
+#}}}
+#{{{Changing sign of all signal to be the same so we don't have to
+# deal with the null of the signal
+print(type(data))
+myslice = data['t2':f_range]
+mysgn = determine_sign(data['ph1',0]['ph2',1])
```
jmfranck, line 46: why is this a function? it's it fairly trivial?
`mysgn = data['ph1',0]['ph2',1].real.sum('t2').run(np.sign)`?

examples/correlation_alignment_example.py
guinneak198: and I am running into trouble running this now because at this point in the master branch fake_data is not a function

examples/correlation_alignment_example.py
```diff
@@ -0,0 +1,84 @@
...
+#this generates fake clean data with a T2 of 0.2s
+# amplitude of 23 (arbitrary), offset of 300 Hz,
+# FWHM 10 Hz
+f_range = (-400,400)
+t2, td, vd, ph1, ph2 = s.symbols('t2 td vd ph1 ph2')
+echo_time = 1e-3
+data = fake_data(
+        23*(1 - 2*s.exp(-vd / 0.2)) * s.exp(+1j*2*s.pi*100*(t2) - abs(t2)*50*s.pi),
+        OrderedDict([
+            ('vd', nddata(r_[0:1:40j],'vd')),
```
jmfranck, line 27: in the loop that I talk about above, this would also change (you would have a power, rather than vd axis)

examples/correlation_alignment_example.py
guinneak198: you had made this into a function.... it doesn't need to be but its there from you

examples/correlation_alignment_example.py
```diff
@@ -0,0 +1,84 @@
...
+#}}}
+#{{{Applying Correlation Routine to Align Data
+data.ft('t2')
+data,opt_shift,sigma = correl_align(data,indirect_dim='vd',
+        signal_pathway = {'ph1':0,'ph2':1},
+        sigma = 50)
+fl.next(r'After Correlation, $\varphi$ domain')
+fl.image(data)
+data.ift('t2')
+data.ft(['ph1','ph2'])
```
jmfranck, line 74: I would not show this

pyspecProcScripts/correlation_alignment.py
```diff
@@ -33,8 +35,15 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',fig_titl
                     sets the tolerance limit for the alignment procedure
     ph1_selection:  int
                     index position of the coherence pathway in phase program 1
```
jmfranck, line 37: this is no longer an argument above?  it's replaced by `signal_pathway`? (which is good)

pyspecProcScripts/correlation_alignment.py
```diff
@@ -52,8 +61,20 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',fig_titl
...
+        indirect = [j for j in s.dimlabels if j in indirect]
+        nScans_len = len(s.getaxis('nScans'))
+        s.smoosh(indirect)
+    s.ift(list(signal_pathway.keys()))
+    ph_lens = []
+    signal_keys = list(signal_pathway)
+    signal_values = list(signal_pathway.values())
+    for x in range(len(signal_keys)):
+        ph_len = len(s.getaxis(signal_keys[x]))
+        ph_lens.append(ph_len)
```
jmfranck, line 77: these could be rewritten as `ph_lens = [len(j) for j in s.getaxis(signal_keys[x])]`

I would eliminate the previous two lines and just do this
`ph_len = {j:ndshape(s)[j] for j in signal_pathway.keys()}`

You will want a dictionary below, otherwise, there is an undetermined order issue here.

pyspecProcScripts/correlation_alignment.py
```diff
@@ -87,18 +109,15 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',fig_titl
...
         s_copy2 = s.C
-        s_copy *= nddata(r_[1.,1.],'DeltaPh2')
-        s_copy *= nddata(r_[1.,1.],'DeltaPh1')
-        s_copy.setaxis('DeltaPh2','#')
-        s_copy.setaxis('DeltaPh1','#')
+        for x in range(len(signal_keys)):
+            ph = [1.]*(len(s.getaxis(signal_keys[x])))
+            ph = np.array(ph)
+            s_copy *= nddata(ph,'DeltaPh%d'%(x+1))
+            s_copy.setaxis('DeltaPh%d'%(x+1),'#')
```
jmfranck, line 116: so something like this:
```
for k,v in ph_len.items():
   ph = ones(v)
   s_copy *= nddata(ph, 'Delta'+k.capitalize())
   s_copy.setaxis('Delta'+k.capitalize(),'#')
```

pyspecProcScripts/correlation_alignment.py
```diff
@@ -87,18 +109,15 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',fig_titl
...
-            print("PH1 INDEX IS:",ph1_index)
-            s_copy['DeltaPh1',ph1_index] = s_copy['DeltaPh1',ph1_index].run(lambda x, 
-                axis=None: roll(x, ph1_index,axis=axis),'ph1')
-        for ph2_index in range(ph2_len):
-            s_copy['DeltaPh2',ph2_index] = s_copy['DeltaPh2',ph2_index].run(lambda x,
-                    axis=None: roll(x,ph2_index,axis=axis),'ph2')
+        for j in range(len(signal_keys)):
+            for ph_index in range(ph_len):
+                s_copy['DeltaPh%d'%(j+1),ph_index] = s_copy['DeltaPh%d'%(j+1),ph_index].run(lambda x, axis=None: roll(x,ph_index,axis=axis),'ph%d'%(j+1))
         for j in range(1,N):
```
jmfranck, line 121: follow idea above

pyspecProcScripts/correlation_alignment.py
```diff
@@ -114,14 +133,18 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',fig_titl
                 fl.image(correl.C.setaxis('vd','#').set_units('vd','scan #'),human_units=False)
         correl.ft_clear_startpoints('t2')
         correl.ft('t2', shift=True, pad=2**14)
-        correl.ft(['DeltaPh1','DeltaPh2'])
-        correl = correl['DeltaPh1',ph1_selection]['DeltaPh2',ph2_selection] + correl['DeltaPh1',0]['DeltaPh2',0]
+        for j in range(len(signal_keys)):
+            correl.ft(['DeltaPh%d'%(j+1)])
+            correl = correl['DeltaPh%d'%(j+1),signal_values[j]]+correl['DeltaPh%d'%(j+1),0]
```
jmfranck, line 138: follow idea above

pyspecProcScripts/correlation_alignment.py
```diff
@@ -114,14 +133,18 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',fig_titl
...
         if my_iter ==0:
             logging.info(strm("holder"))
             if fl is not None:
                 fl.next('correlation function \nfreq domain, after apod')
                 fl.image(correl.C.setaxis('vd','#').set_units('vd','scan #'),human_units=False)
-        f_shift = correl.run(real).argmax('t2')
+        if shift_bounds:
+            f_shift = correl['t2':(-max_shift,max_shift)].run(real).argmax('t2')
+        if not shift_bounds:
```
jmfranck, line 146: else

pyspecProcScripts/correlation_alignment.py
```diff
@@ -52,8 +61,20 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',fig_titl
...
                 the data in the calculation of the correlation function.
     """
-    ph1_len = len(s.getaxis('ph1'))
-    ph2_len = len(s.getaxis('ph2'))
+    logging.info(strm("Applying the correlation routine"))
+    if nScans:
+        phcycdims = [j for j in s.dimlabels if j.startswith('ph')]
+        indirect = set(s.dimlabels)-set(phcycdims)-set([direct])
+        indirect = [j for j in s.dimlabels if j in indirect]
+        nScans_len = len(s.getaxis('nScans'))
```
jmfranck, line 69: this will not hold up if the repeat/averaging dimension is called something else.  All you have to change, though, (I think) is to default it to `None`, and then replace any string instances of `'nScans'` with `nScans` --> making sure that they are always inside a conditional to check `if nScans:` (or `if nScans is not None:`, which should evaluate the same, I think)

jmfranck:
This is looking good.  Work on this.

examples/correlation_alignment_example.py
jmfranck: yes -- this way you will demo the correlation on both an inversion recovery and an enhancement curve.

pyspecProcScripts/correlation_alignment.py
guinneak198: I can make the dictionary but the first guy is not working - I get this error 'TypeError: object of type 'numpy.float64' has no len()' which I can work on trouble shooting but while I adress the other issues I'm going to keep what works but keep this guy in there just commented out.

examples/correlation_alignment_example.py
jmfranck: actually, if you just pull this branch now, it will work -- I think your master is out of date -- when I merged in an up to date master, what you had worked fine for me.

jmfranck: if you want to update master without actually being on master (e.g. if you're working on the correlation_alignment branch), you can usually do `git fetch origin master:master` (you can tab complete the last part).

examples/correlation_alignment_example.py
jmfranck: if what I say here works fine for you, that's better than making extra functions.

pyspecProcScripts/correlation_alignment.py
guinneak198: So I looked through the code and ph_len and ph_lens is not used anywhere in the code and so I deleted this little loop. It still runs on my end without it.

examples/correlation_alignment_example.py
guinneak198: I'm having a hard time visualizing how to do the loop within a loop... I first tried using lists and having a list for the datasets, f_ranges, and signal pathways but ran into trouble with unpacking the list of dictionaries for the signal pathways....

examples/correlation_alignment_example.py
```diff
@@ -0,0 +1,124 @@
...
+    C_SL = 150e-6
+    E_f_range = (-200,400)
+    E_signal_pathway = {'ph1':1}
+    enhancement_data = fake_data(
+            23*(1-(32*power/(0.25 +power))*C_SL*659.33)*s.exp(+1j*2*s.pi*100*(t2) - abs(t2)*50*s.pi) ,
+            OrderedDict([
+                ('power', nddata(r_[0:4:25j],'power')),
+                ('ph1', nddata(r_[0:4]/4.0,'ph1')),
+                ('t2', nddata(r_[0:0.2:256j]-echo_time,'t2'))]),
+            {'ph1':1})
```
jmfranck, line 65: so, we're saying that you would have a loop over a list of tuples, where the elements of the tuple give the a label for the type of data, the OrderedDict of dimension, the signal pathway, anything else that is different.

examples/correlation_alignment_example.py
```diff
@@ -0,0 +1,124 @@
...
+    # deal with the null of the signal
+    myslice = data['t2':f_range]
+    mysgn = select_pathway(myslice,signal_path).real.sum('t2').run(np.sign)
+    data *= mysgn
+    #}}}
+    data = data['t2':f_range]
+    data.ift('t2')
+    #{{{Applying the phase corrections
+    best_shift,max_shift = hermitian_function_test(select_pathway(data,signal_path).C.mean(indirect))
+    data.setaxis('t2', lambda x: x-best_shift).register_axis({'t2':0})
```
jmfranck, line 88: put all the phasing in the same place (zeroth order here, too!)

examples/correlation_alignment_example.py
```diff
@@ -0,0 +1,124 @@
...
+    fl.next('%s After hermitian function test -- Time domain'%label)
+    fl.image(data)
+    data.ft('t2')
+    fl.next('%s After hermitian function test -- Frequency domain'%label)
+    fl.image(data)
+    data.ift('t2')
+    ph0 = select_pathway(data,signal_path)['t2':0]
+    ph0 /= abs(ph0)
+    data /= ph0
```
jmfranck, line 97: (move this up)

examples/correlation_alignment_example.py
```diff
@@ -75,6 +75,8 @@
         fl.image(data)
         data.ft("t2")
         fl.next("After hermitian function test -- Frequency domain")
+        # both of these plots labeled "After hermitian function test" now show
+        # the sign of the signal bouncing back and forth -- why is that!!!????
```
jmfranck, line 79: address/fix this, then remove comment

examples/correlation_alignment_example.py
```diff
@@ -15,7 +15,7 @@
 rcParams["image.aspect"] = "auto"  # needed for sphinx gallery
 
 t2, td, vd, power, ph1, ph2 = s.symbols("t2 td vd power ph1 ph2")
-echo_time = 1e-3
+echo_time = 10e-3
```
jmfranck: it wasn't clear to me whether or not the phasing was working b/c we the echo had been so assymetric, so I move the echo time out a bit just because it should give a pure red echo, which should be a better diagnostic

pyspecProcScripts/first_level/fake_data.py
```diff
@@ -83,7 +83,7 @@ def fake_data(
     ## }}}
     data = clean_data.C
     data.add_noise(fake_data_noise_std)
-    data *= np.exp(1j * 2 * np.pi * frq_noise * (data.fromaxis('t2'))) # the frequency shift
+    data *= np.exp(1j * 2 * np.pi * frq_noise * (data.fromaxis(direct))) # the frequency shift
```
jmfranck, line 86: just a hard-coding I had missed before -- probably from me

pyspecProcScripts/first_level/fake_data.py
```diff
@@ -92,4 +92,5 @@ def fake_data(
     data.ft(direct, shift=True)
     data.ift(direct)
     data.register_axis({direct:0})
+    data.set_units(direct,'s')
```
jmfranck, line 95: also probably from me --- noticed that the plot's didn't have units

jmfranck:
I think that there are too many plots here.  You could definitely eliminate some and still hit on the key points.

Also, there are two errors with the way that the example is running right now:

- The first order (timing) correction for the IR data is wrong.  You can tell this from either the frequency domain or the time domain plots.

- Although both the IR and the enhancement data show a signal inversion before the correction, they do not show it after the correction.

examples/correlation_alignment_example.py
guinneak198: Okay so after some debugging I find that this is a result of extending the echo time. When the echo time is 1e-3 it works as it should... but when we increase this it falls apart... I believe this has to do with the hermitian function as it stands right now but I am unsure. I changed the echo time back to 1ms as this seems to work. I don't know how I would go about fixing it when the echo time is 10 ms as I think that would be editing the hermitian function test which should be another PR separate from this.

examples/correlation_alignment_example.py
guinneak198: so this actually made the phasing fall apart. It works better with the echo time being 1e-3 and as we increase the phasing from the hermitian falls apart.

guinneak198: do we have to adjust the expressions to correlate with the newer echo time? (10 ms) to get this to work?

guinneak198:
so I fixed the issue with the signal inversion not being shown after as well as reducing the number of plots output. As for the hermitian, it seems not to like the higher than 1 ms echo time... Do we need to adjust the expression when we adjust the echo time? Other than that I can't say why it falls apart with higher echo times other than we need a PR to fix the hermitian funciton test maybe?

examples/correlation_alignment_example.py
jmfranck: I'm not sure why you would have a loop within a loop.  What you describe here makes sense .. probably just a syntax issue?

jmfranck: hmm, I realized this is likely an old comment...

examples/correlation_alignment_example.py
jmfranck: OK, then you need a separate PR to demo the hermitian function test.  My response here would be to look and see what your cost function looks like.

examples/correlation_alignment_example.py
jmfranck: if it falls apart for a longer echo time, that's not acceptable.  It could be you need to modify one of the kwargs for the hermitian phasing function, but then that should be obvious from the documentation.

examples/correlation_alignment_example.py
guinneak198: image
My cost function looks as it should and the hermitian does center the echo at 0 so I'm not entirely sure what the issue is?

jmfranck: yes, so make a new branch where you make a copy of this code as a new example, and just delete everything after the part where you apply the hermitian phase correction.

jmfranck:
I would just make one big 2x3 set of subplots and dump all the plots into that -- it would be more legible.  Also, decrease the figure size so that the text is legible.

proc_IR.py
```diff
@@ -7,6 +7,7 @@
 save_npz = False
 #}}}
 coherence_pathway = {'ph1':0,'ph2':1}
+# this should not be part of the PR
```
jmfranck, line 10: note this

proc_echo_mw.py
```diff
@@ -12,6 +12,7 @@
 logger = init_logging("info")
 t2 = symbols('t2')
 signal_pathway = {'ph1':1}
+# shouldn't be part of the pull request
```
jmfranck, line 15: note this

pyspecProcScripts/correlation_alignment.py
```diff
@@ -64,6 +64,7 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',
                 the data in the calculation of the correlation function.
     """
 
+    # review this
```
jmfranck, line 67: note this

pyspecProcScripts/integral_w_error.py
```diff
@@ -39,6 +39,7 @@ def integral_w_errors(s,sig_path,error_path, indirect='vd', direct='t2',fl=None,
     all_labels = set(s.dimlabels)
     all_labels -= set([indirect,direct])
     extra_dims = [j for j in all_labels if not j.startswith('ph')]
+    # review this
```
jmfranck, line 42: note this

pyspecProcScripts/phasing.py
```diff
@@ -6,6 +6,7 @@
 import numpy as np
 import pyspecdata as pysp
 from scipy import linalg
+# where is "logger" coming from? Does "logging" not work as used here?
```
jmfranck, line 9: note this

pyspecProcScripts/third_level/process_IR.py
```diff
@@ -7,6 +7,7 @@
 import numpy as np
 import matplotlib.pyplot as plt
 from sympy import symbols, latex, Symbol
+# this is a circular import!
```
jmfranck, line 10: note this

pyspecProcScripts/third_level/process_IR.py
```diff
@@ -18,6 +19,7 @@ def select_pathway(s,pathway):
 def as_scan_nbr(d):
     '''since we need to relabel vd frequently- we make a method'''
     return d.C.setaxis('vd','#').set_units('vd','scan #')
+# this is written like you're trying to run it as a script!
```
jmfranck, line 22: note this

pyspecProcScripts/third_level/process_enhancement.py
```diff
@@ -27,13 +27,14 @@ def select_pathway(s,pathway):
...
         retval = retval[k,v]
-    return retval    
+    return retval
 # slice out the FID from the echoes,
 # also frequency filtering, in order to generate the
 # list of integrals for ODNP
 # to use: as a rule of thumb, make the white boxes
 # about 2x as far as it looks like they should be
 # leave this as a loop, so you can load multiple files
+# why is this part of this pull request??
```
jmfranck, line 37: note this

examples/correlation_alignment_example.py
```diff
@@ -92,9 +92,13 @@
...
         #{{{ Applying Correlation Routine to Align Data
-        data, opt_shift, sigma = correl_align(
-            data, indirect_dim=indirect, signal_pathway=signal_pathway, sigma=50
+        mysgn = select_pathway(data, signal_pathway).C.real.sum("t2").run(np.sign) # this is the sign of the signal -- note how on the next line, I pass sign-flipped data, so that we don't need to worry about messing with the original signal
+        opt_shift, sigma = correl_align(
+            data*mysgn, indirect_dim=indirect, signal_pathway=signal_pathway, sigma=50
         )
+        s.ift('t2')
+        s *= np.exp(-1j*2*pi*opt_shift*s.fromaxis('t2'))
+        s.ft('t2')
```
jmfranck: the way that it's documented, and the way that makes a bit more sense, is that the correlation alignment calculates and returns your shifts

pyspecProcScripts/correlation_alignment.py
```diff
@@ -167,11 +167,8 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',
...
         fl.next('after correlation\nph0 restored sig. energy=%g'%sig_energy)
         fl.image(s_copy.C.setaxis(indirect_dim,'#').set_units(indirect_dim,'scan #'))
-    s.ift('t2')
-    s *= np.exp(-1j*2*pi*f_shift*s.fromaxis('t2'))
-    s.ft('t2')
     if avg_dim:
         s.chunk(avg_dim,[avg_dim,'power'],[avg_dim_len,-1])
         s.reorder(['ph1',avg_dim,'power','t2'])
-    return s,f_shift,sigma    
+    return f_shift,sigma
```
jmfranck: see [here](https://github.com/jmfrancklab/proc_scripts/pull/52/commits/f95eaf7010e5c71de7328dc04385b8837562c1c3#r670569862}

proc_IR.py
```diff
@@ -7,10 +7,11 @@
...
 for thisfile,exp_type,nodename,postproc,f_range,t_range,clock_correction,IR,ILT in [
-       ('210610_water_TempCont_probe_DNP_1','ODNP_NMR_comp/ODNP',
-           'FIR_34dBm','spincore_IR_v1',
-           (0.1e3,0.345e3),(None,50e-3),True,False,False),
+       ('210614_TEMPOL_500uM_cap_probe_DNP','ODNP_NMR_comp/test_equipment',
+           'FIR_30dBm','spincore_IR_v1',
+           (-0.08e3,0.15e3),(None,60e-3),False,False,False),
```
jmfranck, line 14: changes to this file shouldn't be part of the pull request!

proc_echo_mw.py
```diff
@@ -19,19 +12,20 @@
...
     s = find_file(filename,exp_type=file_location,expno=nodename,
             postproc=postproc,lookup=lookup_table,fl=fl)
     myslice = s['t2':freq_range]
-    mysign = determine_sign(select_pathway(myslice,signal_pathway,mult_ph_dims=False))
-    enhancement,idx_maxpower = process_enhancement(s,searchstr = filename,
-            freq_range=freq_range, t_range=t_range,flip=True,sign=mysign,fl=fl)
+    mysign = select_pathway(myslice, signal_pathway).real.sum('t2').run(np.sign)
+    enhancement = process_enhancement(s,freq_range=freq_range, 
+            t_range=t_range,sgn=mysign,fl=fl)
     fl.show()
```
jmfranck, line 30: these changes should not be part of this pull request!

pyspecProcScripts/third_level/process_IR.py
```diff
@@ -16,21 +19,55 @@ def select_pathway(s,pathway):
...
+            the nddata with the collected signs of the data along the signal that
+            upon multiplication will convert all signal to the same sign. Useful
+            for datasets where a null is present.
+    ILT:    boolean
+            If true, will produce an ILT image plot for visualizing the T1 as a function
+            of offset.
+    fl:     boolean
+            option to show figures produced by processing script.
+        """        
+    s *= sgn
```
jmfranck, line 70: why is this even part of this pull request?

pyspecProcScripts/third_level/process_IR.py
```diff
@@ -105,23 +138,19 @@ def process_IR(s, label='', fl=None,
...
-            ph1_selection=signal_pathway['ph1'],ph2_selection=signal_pathway['ph2'],
+    s,opt_shift,sigma = correl_align(s,indirect_dim='vd',
+            signal_pathway=signal_pathway,
             sigma=10)
-    s.ift('t2')
-    s *= np.exp(-1j*2*pi*opt_shift*s.fromaxis('t2'))
-    s.ft('t2')
     fl.basename = None
     if fl is not None:
         fl.next(r'after correlation, $\varphi$ domain')
```
jmfranck, line 156: again, I'm kind of skipping the rest of this, b/c this file doesn't seem to be something we even want --> it would be replaced by the table of integrals

pyspecProcScripts/third_level/process_enhancement.py
jmfranck: I'm skipping most things with this file b/c it shouldn't be part of this pull request almost definitely, and I also think will be superseded by the table of integrals.

jmfranck (commented):
OK, so I made a bunch of individual comments above --- I'm sorry for my misleading last message -- I had forgotten there were a couple of issues with this pull request, still, but hopefully you should be able to fix them in one round.

pyspecProcScripts/phasing.py
guinneak198: sorry mixed up logger and logging!

pyspecProcScripts/third_level/process_IR.py
jmfranck: you want to save this in your own branch, but this function as a whole will likely be superseded by the table of integrals function.

jmfranck:
also, please fix the conflicts, and make sure everything runs OK, since I didn't test it after my most recent round of changes!

pyspecProcScripts/third_level/process_IR.py
guinneak198: where?

pyspecProcScripts/third_level/process_IR.py
guinneak198: I'm not sure what you mean? it is a script?

pyspecProcScripts/third_level/process_enhancement.py
guinneak198: these are part of the pull request because they use correlation alignment and so they need to be edited since we are only using one cohesive alignment script instead of two now.

guinneak198: The purpose of this PR is to combine the two correlation alignment scripts into one. And so along with that I need to modify the scripts that call correlation alignment to edit accordingly

proc_IR.py
guinneak198: using data I know is good to test the correlation alignment in an actual experiment. The temp control stuff at this point was still sketchy so I'm using data I know is good

proc_echo_mw.py
guinneak198: need to test alignment on actual experimental enhancement data to ensure it works as it should

pyspecProcScripts/third_level/process_IR.py
guinneak198: need to test alignment on actual IR data to ensure it works as it should

pyspecProcScripts/third_level/process_IR.py
guinneak198: the table of integrals will be a separate PR this is again testing the correlation alignment on actual experimental data

proc_IR.py
```diff
@@ -7,10 +7,11 @@
 save_npz = False
 #}}}
 coherence_pathway = {'ph1':0,'ph2':1}
+# this should not be part of the PR
```
guinneak198, line 10: we are editing the correlation alignment function which is used in our exp data processing and so to test that it works on ACTUAL data I add the processing for IR and enhancement

guinneak198 (commented):
biggest confusion seems to be the edits to proc_IR,proc_echo_mw,process_IR, and process_enhancement. These are included to test the new correlation alignment script on actual experimental data as well as the fake data in the example to ensure it is doing what it's supposed to

examples/correlation_alignment_example.py
```diff
@@ -76,6 +76,8 @@
         logger.info(strm("Rough center is:", rough_center))
         data.setaxis("t2", lambda x: x - rough_center).register_axis({"t2": 0})
         data.ft("t2")
+        mysgn = (select_pathway(data,signal_pathway).C.real.sum("t2").run(np.sign))
+        data *= mysgn
```
jmfranck, line 80: why did you add this back in after I removed it?

examples/correlation_alignment_example.py
```diff
@@ -85,27 +87,20 @@
             select_pathway(data.C.mean(indirect), signal_pathway)
         )
         data.setaxis("t2", lambda x: x - best_shift).register_axis({"t2": 0})
-        fl.image(data, ax=ax_list[3])
-        ax_list[3].set_title("Hermitian Test (t)")
-        data.ft("t2")
```
jmfranck: why remove this plot?

examples/correlation_alignment_example.py
```diff
@@ -85,27 +87,20 @@
...
         data.ft("t2")
         data *= mysgn
         fl.image(data, ax=ax_list[1])
-        ax_list[1].set_title("Phase Corrected and Centered")
-        data *= mysgn
+        ax_list[1].set_title("Phased and Centered")
```
jmfranck: you ant to remove the lines having to do with mysgn, I agree, but I don't understand why the plot title is changed.

pyspecProcScripts/third_level/process_enhancement.py
jmfranck: see slack

examples/correlation_alignment_example.py
```diff
@@ -87,20 +85,29 @@
...
+        # {{{ Applying Correlation Routine to Align Data
+        mysgn = (
+            select_pathway(data, signal_pathway).C.real.sum("t2").run(np.sign)
+        )  # this is the sign of the signal -- note how on the next line,
+        #    I pass sign-flipped data, so that we don't need to worry about
+        #    messing with the original signal
         opt_shift, sigma = correl_align(
-            data*mysgn, indirect_dim=indirect, signal_pathway=signal_pathway, 
-            sigma=50, fl=fl
+            data * mysgn, indirect_dim=indirect, signal_pathway=signal_pathway, sigma=50
```
jmfranck: it seems like you had trouble with a merge and didn't merge this in.  This was an important change.

examples/correlation_alignment_example.py
```diff
@@ -76,8 +76,6 @@
         logger.info(strm("Rough center is:", rough_center))
         data.setaxis("t2", lambda x: x - rough_center).register_axis({"t2": 0})
         data.ft("t2")
-        mysgn = (select_pathway(data,signal_pathway).C.real.sum("t2").run(np.sign))
-        data *= mysgn
         data.ift("t2")
```
jmfranck: I think it's best not to do the sign flip anywhere except in the immediate vicinity of the correlation function -- see [the change I had made here](https://github.com/jmfrancklab/proc_scripts/pull/52/commits/a1c8c3a813429aa2bd89d8e4f6462e028dd1872f#r670782511)

examples/correlation_alignment_example.py
```diff
@@ -87,20 +85,29 @@
...
+        ax_list[3].set_title("Hermitian Test (t)")
         data.ft("t2")
-        data *= mysgn
-        fl.image(data, ax=ax_list[1])
-        ax_list[1].set_title("Phased and Centered")
-        #}}}
-        #{{{ Applying Correlation Routine to Align Data
+        fl.image(data, ax=ax_list[2])
+        ax_list[2].set_title("Hermitian Test (v)")
+        fig.tight_layout(rect=[0, 0.03, 1, 0.95])
```
jmfranck: here I ended up rolling back some changes you had made to the images -- this seemed to be tied up with the whole sign-flip business.

examples/correlation_alignment_example.py
```diff
@@ -87,20 +85,29 @@
...
-            sigma=50, fl=fl
+            data * mysgn, indirect_dim=indirect, signal_pathway=signal_pathway, sigma=50
         )
+        s.ift("t2")
+        s *= np.exp(-1j * 2 * pi * opt_shift * s.fromaxis("t2"))
+        s.ft("t2")
         data.ift("t2")
-        data *= np.exp(-1j*2*pi*opt_shift*data.fromaxis('t2'))
+        for k, v in signal_pathway.items():
+            data.ft([k])
```
jmfranck, line 109: if you intended to remove this for a reason, then OK, but things were confusing b/c some of my earlier changes got rolled back during a bad merge, as noted above.

examples/Hermitian_Phasing.py
```diff
@@ -80,20 +80,21 @@
         logger.info(strm("Rough center is:", rough_center))
         data.setaxis("t2", lambda x: x - rough_center).register_axis({"t2": 0})
         data.ft("t2")
-        mysgn = select_pathway(data, signal_pathway).C.real.sum("t2").run(np.sign)
-        data *= mysgn
```
jmfranck, line 84: we should not be sign-flipping our data, because we are not doing a correlation!  (I think I had pointed this out at some point, and it got reverted by accident)

examples/Hermitian_Phasing.py
```diff
@@ -80,20 +80,21 @@
...
         data.ift("t2")
+        # {{{ this is not correct -- you want to use the zeroth order phasing
+        # routine to determine the zeroth order phase, since it allows for
+        # negative data this will flip the negative data to positive
         ph0 = select_pathway(data, signal_pathway)["t2":0]
         ph0 /= abs(ph0)
         data /= ph0
+        # }}}
```
jmfranck, line 90: resolve this here, first, since it will be easier --- feed `select_pathway(data, signal_pathway)["t2":0]` to the zeroth order routine, and show the diagnostic plot.

the correlation example has the same error, and will have the same solution

examples/Hermitian_Phasing.py
guinneak198: this was pulled from the master after you merged in the hermitian phasing PR

examples/correlation_alignment_example.py
```diff
@@ -77,9 +77,13 @@
         data.setaxis("t2", lambda x: x - rough_center).register_axis({"t2": 0})
         data.ft("t2")
         data.ift("t2")
+        # {{{ this is not correct -- you want to use the zeroth order phasing
+        # routine to determine the zeroth order phase, since it allows for
+        # negative data this will flip the negative data to positive
         ph0 = select_pathway(data, signal_pathway)["t2":0]
         ph0 /= abs(ph0)
         data /= ph0
+        # }}}
```
jmfranck: use the same solution as the hermitian function test (and do this after that)

examples/Hermitian_Phasing.py
guinneak198: okay so now this PR is fixing both the correlation alignment example and the phasing example

jmfranck (commented):
I made several comments.  Also note that one of the subplots for the correlation example is currently blank -- the code to generate that plot probably got accidentally deleted at some point.

examples/Hermitian_Phasing.py
jmfranck: well, the should have been fixed in the other PR before it got merged, but something funky happened there

examples/correlation_alignment_example.py
guinneak198: fixed

examples/correlation_alignment_example.py
guinneak198: fixed

examples/correlation_alignment_example.py
guinneak198: yes this was intended to be removed as it's done in the correlation alignment already

examples/Hermitian_Phasing.py
guinneak198: fixed

examples/Hermitian_Phasing.py
guinneak198: fixed both accordingly

examples/correlation_alignment_example.py
guinneak198: done

guinneak198:
okay fixed the zeroth order phasing for hermitian example and correlation and took out sign flipping for both

examples/Hermitian_Phasing.py
```diff
@@ -81,12 +81,9 @@
...
-        # {{{ this is not correct -- you want to use the zeroth order phasing
-        # routine to determine the zeroth order phase, since it allows for
-        # negative data this will flip the negative data to positive
-        ph0 = select_pathway(data, signal_pathway)["t2":0]
-        ph0 /= abs(ph0)
-        data /= ph0
+        # {{{ Zeroth Order Phase Correction
+        retval = zeroth_order_ph(select_pathway(data,signal_pathway))
+        data /= retval
         # }}}
```
jmfranck: yes!  But we do want to show the diagnostic plots for this guy too.  (I would show the diagnostic plots in the Hermitian Phasing example but not in the correlation example)

examples/Hermitian_Phasing.py
```diff
@@ -81,7 +81,7 @@
         data.setaxis("t2", lambda x: x - rough_center).register_axis({"t2": 0})
         data.ft("t2")
         data.ift("t2")
-        data /= zeroth_order_ph(select_pathway(data,signal_pathway))
+        data /= zeroth_order_ph(select_pathway(data,signal_pathway), fl=fl)
```
jmfranck, line 84: here, I just added the diagnostic plot, actually, so you just need to make sure the plot number is set correctly for sphinx.

guinneak198:
okay I edited the plot number (at least I think that's what I did/ is that what the `# sphinx_gallery_thumbnail_number ` is?) and I built it with the shell script and it looks good on my end!

examples/correlation_alignment_example.py
```diff
@@ -94,7 +94,15 @@
...
+            data * mysgn,
+            indirect_dim=indirect,
+            signal_pathway=signal_pathway,
+            sigma=3000 / 2.355,
+            max_shift=300,  # this makes the Gaussian mask 3
+            #                 kHz (so much wider than the signal), and
+            #                 max_shift needs to be set just wide enough to
+            #                 accommodate the drift in signal
+            fl=fl
         )
```
jmfranck: I edited this to give a reasonable set of parameters -- please don't change this.

jmfranck:
So the final result is not actually aligned.
To figure out what's going on, I turned on the diagnostic plot.  Interestingly, it actually shows the data being fully aligned in the diagnostic plots.  So it is able to align things.  But somehow, it returns an array of `f_shift` that is NOT the array of shifts we need to perform this alignment.

So, that needs to be debugged.

jmfranck:
actually, please touch base with @aabeaton -- I see what the problem is, which is that you are not using the most recent version of the alignment code.

In this version, when you iterate, the data is being adjusted with each iteration.  That's not what we want.  We want to add/subtract to the array of f_shift, which is then applied to the *original* data.  From what I remember @aabeaton also said that this worked out to be faster, so you should really make sure to get the most recent version of the code.

guinneak198:
Upon asking for the most up to date version of correlation alignment I was pointed to coherence_pathways/proc_alignment.py using git checkout bf9b560. I notice the biggest difference is when we ift/ft the coherence domains as well as the mask func. I add the original code and edit it to simplify into less lines but keep the original lines commented out (take a look at examples/correlation_alignment_example.py) and it seems to run well for both the enhancement and IR I tried to get rid of my quit statement at the end that I was using just as a quick test (I DID run it with both enhancement and IR and it worked) but I wasnt able to push?
image

jmfranck:
> Upon asking for the most up to date version of correlation alignment I was pointed to coherence_pathways/proc_alignment.py using git checkout bf9b560. I notice the biggest difference is when we ift/ft the coherence domains as well as the mask func. I add the original code and edit it to simplify into less lines but keep the original lines commented out (take a look at examples/correlation_alignment_example.py) and it seems to run well for both the enhancement and IR I tried to get rid of my quit statement at the end that I was using just as a quick test (I DID run it with both enhancement and IR and it worked) but I wasnt able to push?
> image

Try to pull and then push again.

But also, **please don't** leave original code that's commented out ... I just want to diff against what you had before.

jmfranck (commented):
see note above

jmfranck:
The IR in the example is not working.

jmfranck:
actually, no, it's working -- I had the wrong version of pyspecdata checked out!

jmfranck:
so, just look through my more recent changes and see that they're OK by you!

pyspecProcScripts/correlation_alignment.py
```diff
@@ -20,117 +22,141 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',fig_titl
...
+    for x in range(len(signal_keys)):
+        for_nu_center = for_nu_center[signal_keys[x],signal_values[x]]
+    nu_center = for_nu_center.mean(indirect_dim).C.argmax(direct)
     logging.info(strm("Center frequency", nu_center))
     for my_iter in range(100):
         i += 1
         logging.info(strm("*** *** ***"))
         logging.info(strm("CORRELATION ALIGNMENT ITERATION NO. ",i))
         logging.info(strm("*** *** ***"))
+        s_orig.ift(direct)
```
guinneak198, line 105: why?/what is this for?

pyspecProcScripts/correlation_alignment.py
```diff
@@ -20,117 +22,141 @@ def correl_align(s, align_phases=False,tol=1e-4,indirect_dim='indirect',fig_titl
...
-        s_copy.ift('t2')
-        s_copy2 = s.C
-        s_copy *= nddata(r_[1.,1.],'DeltaPh2')
-        s_copy *= nddata(r_[1.,1.],'DeltaPh1')
-        s_copy.setaxis('DeltaPh2','#')
-        s_copy.setaxis('DeltaPh1','#')
+            s_copy /= ph0
+        s_copy.ft(direct)
+        this_mask = exp(-(s_copy.fromaxis(direct)-nu_center)**2/(2*sigma**2))
+        s_copy *= this_mask
```
guinneak198, line 113: good call!

guinneak198 (commented):
okay looks good on my end! Only question I had was why we are doing an ift then copy in the correlation alignment script but other than that it looks good!

pyspecProcScripts/correlation_alignment.py
jmfranck: you had added the ift --> I just changed the name from `s` to `s_orig` to clarify a bit more.

jmfranck: yes, this is just line 82 below in read (from b74fba) --> in addition to changing the name, I also moved it up and changed the fact that you were applying the ph0 to the original data instead of the copy

jmfranck: if you're **rather** saying it would be smarter to just demand signal in the time domain, so that you don't ift the `s_orig` at the beginning and ft it at the end, then by all means, please do that!  You would, though, want to make sure that you pass an assert statement at the beginning, and that the signal energy calculation looks reasonable.

guinneak198: okay no that makes sense! I was just wondering what it was doing is all but it makes sense now!

guinneak198:
I think this is good to go!

PR #55, opened 2021-06-28
Test Integration PR

guinneak198:
compares the propagated error of the inactive CT pathways with the traditional method of finding error as well as std


jmfranck:
lots of extra files.

PR #56, opened 2021-06-29
Fixes regarding erroneous use of abs val in integrate limits

aabeaton:


pyspecProcScripts/integrate_limits.py
```diff
@@ -1,32 +1,43 @@
...
-from pylab import subplots
+from pylab import subplots, axvline
 import numpy as np
 from .fwhm_calculate import fwhm_calculator
 import logging
 
 
 def integrate_limits(s, axis="t2", fwhm=100, fl=None):
-    signal_sign = s.C.sum(axis).run(np.real).run(np.sign)
-    temp = abs(s).real * signal_sign
```
jmfranck, line 10: add a docstring

examples/demo_integrate_limits.py
```diff
@@ -0,0 +1,48 @@
+"""
+Demonstrate Integrate Limits
+============================
+
+Add a nice docstring here!
```
jmfranck, line 5: fill this out

examples/demo_integrate_limits.py
```diff
@@ -0,0 +1,48 @@
...
+    data.setaxis('t2', lambda x: x-echo_time).register_axis({"t2":0}) # this
+    data.reorder(["ph1", "ph2", "vd"])
+    fl.next("fake data -- time domain")
+    fl.image(data)
+    data.ft("t2")
+    # {{{ make it unitary again
+    data /= sqrt(ndshape(data)["t2"]) * data.get_ft_prop('t2','dt')
+    # }}}
+    fl.next("fake data -- freq domain")
+    fl.image(data)
```
jmfranck, line 47: this is based on table_of_integrals -- it just generates fake data

examples/demo_integrate_limits.py
```diff
@@ -0,0 +1,48 @@
...
+    # }}}
```
jmfranck, line 48: here, just add code that applies `integrate_limits` to `data` and which displays the appropriate diagnostic plots (I think this is just 1 line of code?)

aabeaton: ok I have done this

pyspecProcScripts/integrate_limits.py
aabeaton: ok I have added a doc string

aabeaton:
I have this up to date with our most recent discussions regarding the integration limits -- it can choose between a Lorentzian or a Gaussian matched filter, and it also has room to adapt Lorentzian to Gaussian transformation. There are also documentation strings and the example demonstrates Gaussian matched filter and Lorentzian matched filter.

pyspecProcScripts/integrate_limits.py
```diff
@@ -1,43 +1,148 @@
...
+        filter_width = abs(signal_E-1/sqrt(2)).argmin('sigma').item()
+    if Lorentzian_Conv:
+        filter_width = abs(signal_E-signal_E.max()/2).argmin('sigma').item()
+    if fl is not None:
+        fl.next('integration diagnostic -- signal Energy')
         fl.plot(signal_E, human_units=False)
         fl.plot(signal_E['sigma':(filter_width,filter_width+1e-6)],'o', human_units=False)
+    print("*** *** ***")
+    print("FILTER WIDTH IS",filter_width)
+    print("*** *** ***")
```
jmfranck, line 80: please use log statements rather than print statements.

pyspecProcScripts/integrate_limits.py
```diff
@@ -1,43 +1,148 @@
...
+    # Generate the appropriate convolution functions
+    Gaussian_func = lambda x,y: exp(-(x**2)/(2.0*(y**2)))
+    Lorentzian_func = lambda x,y: exp(-(abs(x))/y)
+    if Gaussian_Conv:
+        convfunc = Gaussian_func
+    if Lorentzian_Conv:
+        convfunc = Lorentzian_func
```
jmfranck, line 87: defs should just go inside the conditionals -- second one should be elif

pyspecProcScripts/integrate_limits.py
```diff
@@ -1,43 +1,148 @@
...
+    if Lorentz_to_Gauss:
+        # not immediately sure why, but I need to do this in order to get freq limits around the peak (otherwise they are around 0)
+        rough_center = abs(temp).C.mean_all_but('t2').argmax('t2').item()
+        temp.setaxis('t2', lambda t: t- rough_center).register_axis({'t2':0})
+    if Lorentz_to_Gauss:
+        temp = temp['t2':(7e-3,None)]
```
jmfranck, line 93: think about whether we want this to be part of this function, or a different function -- I think I can see why this would make sense, but worry it's over-complicating this function

pyspecProcScripts/integrate_limits.py
```diff
@@ -1,43 +1,148 @@
...
+        fl.plot(abs(for_manual)/abs(for_manual).max(), alpha=0.5, label='abs data')
+        fl.plot(myfilter, alpha=0.5, label='filter')
+        newdata = for_manual*myfilter
+        fl.next('Overlay abs data')
+        fl.plot(abs(for_manual), alpha=0.5, label='before applying filter')
+        fl.plot(abs(newdata), alpha=0.5, label='after applying filter')
+        temp = newdata.C
+        temp.ft('t2')
+    #}}}
+    if not manual_convolve:
```
jmfranck, line 120: I think just delete the manual convolution option

pyspecProcScripts/integrate_limits.py
```diff
@@ -1,43 +1,148 @@
...
+
+    fl: None or figlist_var()
+    `   to show diagnostic plots, set `fl` to the figure list; set `fl` to None in
+        order not to see any diagnostic plots
 
-    Returns
-    =======
-    retval: ndarray
-        A len 2 ndarray tuple with the start and stop indeces for the integration
-        bounds.
```
jmfranck, line 21: why delete this?  Also, if you're going to use a Lorentzian to Gaussian transform, then you would need to also return the data that has be Lorentzian to Gaussian transformed -- this is part of what makes me wonder if we shouldn't just set that up as a separate function

examples/demo_integrate_limits_realData.py
```diff
@@ -0,0 +1,43 @@
...
+            print("there is only one dimension left -- standard 1D zeroth order phasing")
+            ph0 = ph0/abs(ph0)
+        d /= ph0
+        fl.basename = expno
+        # Put what goes into here in the time domain
+        freq_lim = integrate_limits(select_pathway(d,signal_pathway),
+                Gaussian=False,
+                Lorentzian=True,
+                fl=fl)
+fl.show();quit()
```
jmfranck, line 43: examples cannot contain quit statements (this is not just me being picky -- they really can't)

examples/demo_integrate_limits.py
```diff
@@ -0,0 +1,80 @@
...
+            ("t2" , nddata(r_[0:0.2:256j]-echo_time, "t2"))]),
+            {"ph1": 0, "ph2": 1},
+            scale=10.)
+    data.setaxis('t2', lambda x: x-echo_time).register_axis({"t2":0}) # this
+    data.reorder(["ph1", "ph2", "vd"])
+    fl.next("fake data -- time domain")
+    fl.image(data)
+    data.ft("t2")
+    # {{{ make it unitary again
+    data /= sqrt(ndshape(data)["t2"]) * data.get_ft_prop('t2','dt')
```
jmfranck, line 55: JF note to self: This is dumb -- let's just make a "unitary" flag for the FT.

pyspecProcScripts/integrate_limits.py
aabeaton: they are defined outside the conditionals for the Lorentz to Gauss part

pyspecProcScripts/integrate_limits.py
aabeaton: My opinion is to keep it, because my understanding is that the Lorentzian to Gaussian is meant to ensure that we get the best integrals, which would be relevant here. In fact, I would think, if Lorentzian to Gaussian works out very well, then we wouldn't have any of the other means (they are nice as examples of convolution, but maybe are to be placed elsewhere, not in this function, which is meant to find the best limits for calculating the integral of our peak)

pyspecProcScripts/integrate_limits.py
aabeaton: I will also just note that this example only works correctly if fix_convolve.py is checked out on pyspecdata, the last that I checked, unless those changes are now on master (and master is up to date on your computer)

pyspecProcScripts/integrate_limits.py
aabeaton: I did not mean to delete this

pyspecProcScripts/integrate_limits.py
jmfranck: yes I'm just saying that they should go inside -- unless I'm missing something and they are used elsewhere

(Why define the Gaussian (Lorentzian) function if you're not going to use it.)

pyspecProcScripts/integrate_limits.py
jmfranck: OK, then in the event that we do this, we do need to return the transformed signal.

pyspecProcScripts/integrate_limits.py
jmfranck: so, I actually went to merge that one in, and noticed that there are problems with it.

I think you want to go and make the fixes there that I'm suggesting -- then to maintain the behavior of this code, just invert the width in any call to `convolve` that you make here.

examples/demo_integrate_limits_realData.py
jmfranck: (it actually causes it to quit out of sphinx -- a good thing to know, because it's hard to figure out what happened the first time this happens to you)

pyspecProcScripts/integrate_limits.py
aabeaton: Yes I think that is good now, and here I will invert the line widths so that it is compatible. I did not merge the convolve branch with master yet, which is what I had done with this one before you reviewed it -- if you want me to do that, just let me know

pyspecProcScripts/integrate_limits.py
aabeaton: The only reason it is like that is the Lorentzian-to-Gaussian. This boolean controls the method used to calculate the filter width, based on the energies. For Loretzian-to-Gaussian, I want to use the filter width for the Lorentzian (so I set Lorentzian to True), but I also convolve with a Gaussian (before dividing by the Lorentzian), and so rather than define the Gaussian twice in the same script, I set it outside the booleans so it can get called when it is needed, either for 1. Gaussian convolution or 2. Lorentzian-to-Gaussian

pyspecProcScripts/integrate_limits.py
aabeaton: ok I think that can be easily done through two different return statements separated by a conditional

examples/demo_integrate_limits_realData.py
aabeaton: I did not know, but I figured something to do with the gallery images..

aabeaton:
The latest commit here should have everything in this PR. I do see what you mean about the code being complicated by Lorentzian to Gaussian part. I don't know if we want it in this function yet, so as a temporary thing maybe we could take it out, but I do believe ultimately it will be the way to do this (via Lorentzian to Gaussian transform to determine int limits)

pyspecProcScripts/integrate_limits.py
```diff
@@ -1,43 +1,132 @@
...
+    elif Lorentzian_Conv:
+        filter_width = abs(signal_E-signal_E.max()/2).argmin('sigma').item()
+    if fl is not None:
+        fl.next('integration diagnostic -- signal Energy')
         fl.plot(signal_E, human_units=False)
         fl.plot(signal_E['sigma':(filter_width,filter_width+1e-6)],'o', human_units=False)
+    logger.info(strm("FILTER WIDTH IS",filter_width))
+    # Generate the appropriate convolution functions
+    Gaussian_func = lambda x,y: exp(-(x**2)/(2.0*(y**2)))
+    Lorentzian_func = lambda x,y: exp(-(abs(x))/y)
```
jmfranck, line 89: move these function definitions into the conditionals below.

jmfranck: also, see comments below

pyspecProcScripts/integrate_limits.py
```diff
@@ -1,43 +1,132 @@
...
+        set `Lorentzian` to True in order to calculate the matched-filter via
+        Lorentzian convolution of the frequency domain signal (i.e., apply a
+        matched Lorentzian filter to peak in frequency domain).
+        If `Lorentzian` is set to True, set `Gaussian` to False.
+
+
+    Lorentzian_to_Gaussian: boolean
+        set `Lorentzian_to_Gaussian` to True in order to apply this
+        transformation to your data, in order to determine integral limits.
+        Must set `Lorentzian` to True and `Gaussian` to False.
```
jmfranck, line 48: make a single keyword with three string options

pyspecProcScripts/integrate_limits.py
```diff
@@ -1,43 +1,132 @@
...
+    if Lorentz_to_Gauss:
+        print("Must set Lorentzian convolve to True for Lorentzian to Gaussian transformation... doing so now")
+        Lorentzian_Conv = True
+        Gaussian_Conv = False
+    #}}}
+    # Calculating matched filter
+    sigma = nddata(np.linspace(1e-10,1e-1,1000),'sigma').set_units('sigma','s')
+    if Gaussian_Conv:
+        gaussians = np.exp(-temp.C.fromaxis(axis)**2/2/sigma**2)
+        signal_E = (abs(temp * gaussians)**2).sum(axis)
```
jmfranck, line 73: this and line 76 can be condensed if you just call "gaussians" and "lorentzians" the same thing (defined differently based on the filter that you're using)

pyspecProcScripts/integrate_limits.py
```diff
@@ -1,43 +1,132 @@
...
+        # thus we need to multiply our filter by π, in order to have filters of
+        # the appropriate range
+        this_filter = (filter_width*pi)/2
+        logger.info(strm("Filter width for Lorentz-to-Gauss",this_filter))
+        temp.convolve('t2', 1/this_filter, convfunc=Gaussian_func)
+        temp.ift('t2')
+        temp /= Lorentzian_func(temp.C.fromaxis('t2'),filter_width)
+        temp.ft('t2')
+    else:
+        temp.convolve('t2', (1/filter_width), convfunc=convfunc)
```
jmfranck, line 116: I think that everything would be less complicated if we didn't use convolve at all.

Already, what you have above is not a Lorentzian function, but a complex exponential -- it's the IFT of the Lorentzian.

There is no need to use convolve to handle the Gaussian -- convolving with a Gaussian of 1/filter_width is the same as multiplying by a Gaussian of filter_width in the inverse domain.

pyspecProcScripts/integrate_limits.py
```diff
@@ -1,43 +1,132 @@
...
+        convfunc = Gaussian_func
+    elif Lorentzian_Conv:
+        convfunc = Lorentzian_func
+    if Lorentz_to_Gauss:
+        # not immediately sure why, but I need to
+        # do this in order to get freq limits
+        # around the peak (otherwise they are
+        # around 0)
+        rough_center = abs(temp).C.mean_all_but('t2').argmax('t2').item()
+        temp.setaxis('t2', lambda t: t- rough_center).register_axis({'t2':0})
```
jmfranck, line 100: I would not do this.  We should be assuming the data has already been properly phased at this point.

pyspecProcScripts/integrate_limits.py
aabeaton: ok, I have moved these

pyspecProcScripts/integrate_limits.py
aabeaton: ok, I have done this

pyspecProcScripts/integrate_limits.py
aabeaton: I have done this too

aabeaton:
I have made the suggested changes here. Everything works on my end and seems reasonable.

examples/demo_integrate_limits.py
```diff
@@ -0,0 +1,78 @@
+"""
+Demonstrate Integrate Limits
+============================
+
+For this demonstration, we generate inversion
```
jmfranck, line 5: this doesn't run for me -- see below:
```
$ py demo_integrate_limits.py
logging output to C:\Users\johnf\pyspecdata.0.log
C:\Users\johnf\pyspecdata.0.log appears to be locked or otherwise inaccessible:
I'm going to explore other options for fileno
logging output to C:\Users\johnf\pyspecdata.1.log
--> integrate_limits.py(70):root     integrate_limits 2021-08-24 09:23:08,218
INFO: FILTER WIDTH IS 0.0043043044
--> integrate_limits.py(70):root     integrate_limits 2021-08-24 09:23:08,218
INFO: FILTER WIDTH IS 0.0043043044
Determined frequency limits via Lorentzian filter of [  0.         194.23828125]

Traceback (most recent call last):
  File "demo_integrate_limits.py", line 68, in <module>
    freq_lim = integrate_limits(data['ph1',0]['ph2',1],
TypeError: integrate_limits() got an unexpected keyword argument 'Gaussian'
(base)
johnf@DESKTOP-O82GMEV MINGW64 ~/git_repos/proc_scripts/examples (fix_integrate_limits)
$ git rev-parse @
b9d5e4530ac19c32b8b7113ca6a7f3b52949a151
(base)
johnf@DESKTOP-O82GMEV MINGW64 ~/git_repos/proc_scripts/examples (fix_integrate_limits)
$ cd ~/notebook/pyspecdata/
(base)
johnf@DESKTOP-O82GMEV MINGW64 ~/notebook/pyspecdata (master)
$ git rev-parse @
cc24092916c5b97a3b1656f5b04b8ad14c5cc6dc
```

jmfranck: (the `git rev-parse` lines are to show you the commits that I'm using)

aabeaton: I see - I only made the necessary change to the real data script, not the fake data script. I will add it now.

aabeaton: New version has been pushed and should work now for you.

examples/demo_integrate_limits.py
```diff
@@ -46,33 +46,34 @@
             ("ph2" , nddata(r_[0:4] / 4.0, "ph2")),
             ("t2" , nddata(r_[0:0.2:256j]-echo_time, "t2"))]),
             {"ph1": 0, "ph2": 1},
-            scale=10.)
-    data.setaxis('t2', lambda x: x-echo_time).register_axis({"t2":0}) # this
+            scale=20.)
```
jmfranck, line 49: I just thought more noise was better

examples/demo_integrate_limits.py
```diff
@@ -46,33 +46,34 @@
...
+    # {{{ just have the data phase (not testing phasing here)
+    data.setaxis('t2', lambda x: x-echo_time).register_axis({"t2":0})
+    data = data['t2',0:-3] # dropping the last couple points avoids aliasing
+    #                        effects from the axis registration
+    #                        (otherwise, we get "droop" of the baseline)
```
jmfranck, line 54: something I noticed while being picky -- might be best to roll this into the register_axis function -- I'm not sure.

examples/demo_integrate_limits.py
```diff
@@ -46,33 +46,34 @@
...
     data.reorder(["ph1", "ph2", "vd"])
     fl.next("fake data -- time domain")
     fl.image(data)
-    data.ft("t2", unitary=True)
+    fl.next("FID sliced -- time domain")
+    data = data['t2':(0,None)]
+    data['t2',0] *= 0.5
+    ph0 = data['t2',0].data.mean()
+    ph0 /= abs(ph0)
+    data /= ph0
```
jmfranck, line 64: we will  always be working with FID data for the integration

examples/demo_integrate_limits.py
```diff
@@ -46,33 +46,34 @@
...
+    for method in ['Lorentzian','Gaussian']:
+        fl.basename = method
+        freq_lim = integrate_limits(data['ph1',0]['ph2',1],
+                convolve_method='Lorentzian',
+                fl=fl)
+        fl.next("fake data -- show freq limit selection, Lorentzian filter")
+        fl.plot(data['ph1',0]['ph2',1])
+        axvline(x=freq_lim[0])
+        axvline(x=freq_lim[-1])
+        print("Determined frequency limits via",method,"filter of",freq_lim)
```
jmfranck: here:

- I've just condensed into one loop.
- I pushed the ft/ift into the function, since we will actually want to write the function to accept data in the frequency domain, which is more natural for an integration

pyspecProcScripts/integrate_limits.py
```diff
@@ -47,23 +47,27 @@ def integrate_limits(s, axis="t2",
         order not to see any diagnostic plots
 
     """
+    s.ift('t2') # move ift into function to eventually discard
+    #assert s.get_ft_prop(axis), "data must be in the frequency domain along %s!!"%axis
```
jmfranck, line 51: the goal here is to get rid of line 50 (and the corresponding ft at the end) and then uncomment line 51

pyspecProcScripts/integrate_limits.py
```diff
@@ -47,23 +47,27 @@ def integrate_limits(s, axis="t2",
...
     temp = s.C.mean_all_but(axis)
+    convolve_method = convolve_method.lower()
```
jmfranck, line 53: just for flexibility

pyspecProcScripts/integrate_limits.py
```diff
@@ -47,23 +47,27 @@ def integrate_limits(s, axis="t2",
...
-        convolution_set = np.exp(-temp.C.fromaxis(axis)**2/2/sigma**2)
-    elif convolve_method == 'Lorentzian':
-        convolution_set = np.exp(-abs(temp.C.fromaxis(axis))/sigma)
+    if convolve_method == 'gaussian':
+        convolution_set = np.exp(-temp.fromaxis(axis)**2/2/sigma**2)
+    elif convolve_method == 'lorentzian':
+        convolution_set = np.exp(-abs(temp.fromaxis(axis))/sigma)
     elif convolve_method == 'Lorentzian_to_Gaussian':
-        convolution_set = np.exp(-abs(temp.C.fromaxis(axis))/sigma)
+        convolution_set = np.exp(-abs(temp.fromaxis(axis))/sigma)
```
jmfranck: lots of extra copies -- definitely not needed when calling fromaxis

pyspecProcScripts/integrate_limits.py
aabeaton: The data is currently given in the frequency domain. We are doing the calculations in the time domain. So why would I get rid of line 50 here? The data needs to be put into the time domain at this point so that we can carry out the calculations.

jmfranck: yes, you would need to ift at some later point in the code.  The first plot is in the frequency domain, so the ift should be moved at least after that.  If you apply the ift to only the temp copy of the data, which I think is possible, then you would not need the ft that happens at the end of the function, currently

examples/demo_integrate_limits.py
```diff
@@ -64,17 +67,25 @@
...
                 fl=fl)
-        fl.next("fake data -- show freq limit selection, %s filter"%method)
+        logger.debug("about to run potentially problematic next command")
+        fl.next("fake data -- show freq limit selection")
         fl.plot(data['ph1',0]['ph2',1])
         axvline(x=freq_lim[0])
         axvline(x=freq_lim[-1])
         print("Determined frequency limits via",method,"filter of",freq_lim)
+    time_list.append(time.time())
+    print("figure list-----\n",fl) # there were extra/blank figures, and I added this to help me figure out which they were (update to pyspecdata needed)
```
jmfranck: note this feature (I end up pulling it out b/c it was for debugging, but good to know about) --> I added this to pyspecdata today

pyspecProcScripts/apod_matched_filter.py
```diff
@@ -3,7 +3,6 @@
 def apod_matched_filter(s, axis='t2',
         convolve_method='gaussian',
         ret_width=False,
-        basename=None,
```
jmfranck: basename is an attribute of fl, so as long as you pass fl...

pyspecProcScripts/integrate_limits.py
```diff
@@ -61,9 +61,10 @@ def integrate_limits(s, axis="t2",
     # }}}
     convolve_method = convolve_method.lower()
     if fl is not None:
+        logger.debug(strm("from integrate_limits, basename is",fl.basename,"and current",fl.current))
+        fl.push_marker()
         forplot = temp
         fl.next('integration diagnostic')
-        fl.push_marker()
```
jmfranck, line 34: push marker and pull marker come at the very beginning and end of your interaction with fl.

examples/demo_integrate_limits.py
```diff
@@ -64,17 +67,25 @@
     data /= ph0
     fl.image(data)
     data.ft("t2")
+    time_list.append(time.time())
     fl.next("fake data -- freq domain")
     fl.image(data)
     for method in ['Lorentzian','Gaussian']:
-        fl.basename = method
+        time_list.append(time.time())
+        fl.basename = method + " filter:"
```
jmfranck, line 70: this way, you get two sets of plots, called "Lorentzian" and "Gaussian"

pyspecProcScripts/apod_matched_filter.py
```diff
@@ -43,23 +43,26 @@ def apod_matched_filter(s, axis='t2',
     elif convolve_method == 'lorentzian':
         convolution_set = np.exp(-abs(temp.fromaxis(axis))/sigma)
     signal_E = (abs(temp * convolution_set)**2).sum(axis)
-    signal_E /= signal_E.data.max()
+    signal_E /= abs(signal_E.data).max()
```
jmfranck: on looking at this again, I realize that the data is positive to start with, so you can remove abs if you want.

pyspecProcScripts/apod_matched_filter.py
```diff
@@ -43,23 +43,26 @@ def apod_matched_filter(s, axis='t2',
...
     if convolve_method == 'gaussian':
         filter_width = abs(signal_E-1/sqrt(2)).argmin('sigma').item()
     elif convolve_method == 'lorentzian':
-        filter_width = abs(signal_E-signal_E.max()/2).argmin('sigma').item()
+        filter_width = abs(signal_E-1/2).argmin('sigma').item()
     logger.info(strm("FILTER WIDTH IS",filter_width))
```
jmfranck: should do the same thing, but max is redundant w/ above and this way it's easier to compare to Gaussian

pyspecProcScripts/apod_matched_filter.py
```diff
@@ -43,23 +43,26 @@ def apod_matched_filter(s, axis='t2',
...
     if fl is not None:
         fl.next('matched filter diagnostic -- time domain')
         fl.plot(abs(temp), alpha=0.6, label='abs val, before mult')
-    if convolve_method == 'gaussian':
-        temp *= np.exp(-temp.fromaxis(axis)**2/2/filter_width**2)
-    elif convolve_method in 'lorentzian':
-        temp *= np.exp(-abs(temp.fromaxis(axis))/filter_width)
+        norm_cnst = abs(temp)[axis,1] / abs(filter_data)[axis,1]
+        fl.plot(abs(filter_data) * norm_cnst, alpha=0.6, label='matched apod. filter')
+    temp *= filter_data
```
jmfranck: this is just so I can see the filter itself, which I feel is more useful than just before + after mult.

pyspecProcScripts/lorentzian_to_gaussian.py
```diff
@@ -48,4 +48,5 @@ def L2G(s, axis='t2',
         fl.plot(temp.C.ft('t2'),
                 alpha=0.6,
                 label='after')
+        axhline(y=0,color='k')
```
jmfranck, line 51: I was concerned that the baseline was depressed near the sides of the peak -- this might be a slight issue, but adding the axhline actually shows it's not quite as bad as I thought.

pyspecProcScripts/apod_matched_filter.py
jmfranck: I had only tested in L2G.  I'm committing a fix where you can see all 3 lines for integrate as well.

jmfranck: (i.e. how I normalize the functions)

pyspecProcScripts/integral_w_error.py
```diff
@@ -30,7 +30,11 @@ def integral_w_errors(s,sig_path,error_path, indirect='vd', direct='t2',fl=None,
              not included in the signal pathway.
     """
     assert s.get_ft_prop(direct), "need to be in frequency domain!"
-    frq_slice = integrate_limits(select_pathway(s,sig_path),fl=fl)
+    s.ift('t2')
+    frq_slice = integrate_limits(select_pathway(s,sig_path),
+            convolve_method='Gaussian',
+            fl=None)
+    s.ft('t2')
```
jmfranck, line 37: I'm confused by this -- as we defined it now, isn't the signal supposed to be in teh frequency domain when you call "integrate_limits"?

jmfranck: actually this breaks the check_integration_error example, so I'm doing to roll it back.

jmfranck: rather, I leave the central function call alone, but remove the bracketing expressions that put it in the time domain.

aabeaton: yes it should be in the frequency domain when given to integrate_limits - I had not touched integral_w_error for some time and was mostly focused on integrate_limits, so I think that just got missed (I hadn't been running check_integration_error as one of the example scripts)

jmfranck (commented):
Ok, I made a subsequent commit! Going to comment/explain that now.

examples/demo_integrate_limits_realData.py
```diff
@@ -14,15 +13,8 @@
...
-        ph1_val = 1
-        ph2_val = 0
-        signal_pathway = {"ph1": ph1_val, "ph2": ph2_val}
+        signal_pathway = {"ph1": 1, "ph2": 0}
         excluded_pathways = [(0, 0),(0,-1)]
-        def select_pathway(s,pathway):
-            retval = s
-            for k,v in pathway.items():
-                retval = retval[k,v]
-            return retval
```
jmfranck: we have this function defined so that we don't have to keep defining it FYI

examples/demo_integrate_limits_realData.py
```diff
@@ -36,7 +28,9 @@ def select_pathway(s,pathway):
...
         fl.basename = expno
         d.ft('t2')
-        freq_lim = integrate_limits(select_pathway(d,signal_pathway),
-                convolve_method='Lorentzian',
-                fl=fl)
+        for method in ['Lorentzian','Gaussian']:
+            fl.basename = "%s method: "%method
+            freq_lim = integrate_limits(select_pathway(d,signal_pathway),
+                    convolve_method=method,
+                    fl=fl)
```
jmfranck: just for completeness

pyspecProcScripts/apod_matched_filter.py
```diff
@@ -59,9 +59,13 @@ def apod_matched_filter(s, axis='t2',
...
     if fl is not None:
         fl.next('matched filter diagnostic -- time domain')
-        norm_cnst = abs(temp)[axis,1] / abs(filter_data)[axis,1]
+        fl.plot(abs(temp), alpha=0.6, label='abs val, before mult')
+        norm_cnst = abs(temp)[axis:(0+temp.get_ft_prop(axis,"dt"))] / abs(filter_data)[axis:(0+temp.get_ft_prop(axis,"dt"))]
         fl.plot(abs(filter_data) * norm_cnst, alpha=0.6, label='matched apod. filter')
     temp *= filter_data
+    if fl is not None:
+        fl.next('matched filter diagnostic -- time domain')
+        fl.plot(abs(temp), alpha=0.6, label='abs val, after mult')
```
jmfranck: this is the fix of the scaling I talked about earlier.

pyspecProcScripts/integrate_limits.py
```diff
@@ -7,32 +7,26 @@
 from matplotlib.pyplot import annotate
 from .apod_matched_filter import apod_matched_filter
 
-logger = init_logging("debug")
```
jmfranck: only initialize the logger in your main script, never in the library functions.

pyspecProcScripts/integrate_limits.py
```diff
@@ -7,32 +7,26 @@
...
     Limits are found by applying a matched filter
-    (choice of Lorentzian or Gaussian) to the
+    (choice of Lorentzian or Gaussian ``convolve_method``) to the
     frequency domain data, which it then uses to
     determine the integration limits based on a
-    cut off from the maximum signal intensity.
+    ``cutoff`` amplitude based as a faction of the maximum signal intensity.
+
+    Parameters
+    ==========
```
jmfranck: hopefully this is self-explanatory

pyspecProcScripts/integrate_limits.py
```diff
@@ -44,6 +38,13 @@ def integrate_limits(s, axis="t2",
         Lorentzian convolution of the frequency
         domain signal.
 
+    cutoff: float
+        set the integration limit to the points where the convolved signal
+        crosses this fraction of its maximum.
+        (e.g. cutoff=0.5 would give integration limits that correspond to the
+        frequencies marking the full width half maximum of the convolution of
+        the signal and its matched filter)
+
```
jmfranck, line 47: also hopefully self-explanatory -- please rephrase if not

pyspecProcScripts/lorentzian_to_gaussian.py
```diff
@@ -33,20 +39,24 @@ def L2G(s, axis='t2',
...
+        fl.plot(temp,
                 alpha=0.6,
                 label='after')
         axhline(y=0,color='k')
+    # {{{ actually apply the transform to the data and return
+    temp = s.C.ift(axis)
+    temp *= np.exp(-temp.fromaxis(axis)**2/2/(filter_width*pi/2)**2)
+    temp /= np.exp(-abs(temp.fromaxis(axis))/filter_width)
+    temp.ft(axis)
+    # }}}
```
jmfranck, line 61: previously, the summed 1D data was returned, rather than the 2D data w/ transform applied/

pyspecProcScripts/simple_functions.py
```diff
@@ -1,23 +1,31 @@
...
+        pathway = kwargs
     else:
-        for k in pathway.items():
-            retval=retval[k]
-        return retval    
+        raise ValueError("your arguments don't make any sense!!")
+    retval = s
+    for k, v in pathway.items():
+        retval = retval[k,v]
+    return retval
```
jmfranck, line 28: I didn't understand this multi_ph_dims arg, and it's only used in one spot.  I got rid of that and modified to allow the user to pass either a dictionary (as we have been doing), or just key/value kwargs, like ``select_pathway(data,ph1=1,ph2=-2)`` --> if that's not clear, please feel free to mod docstring.

pyspecProcScripts/third_level/process_enhancement.py
```diff
@@ -24,11 +25,6 @@
 logger = init_logging("info")
 t2 = symbols('t2')
 thesecolors = cycle(list('bgrcmykw'))
-def select_pathway(s,pathway):
-    retval = s
-    for k,v in pathway.items():
-        retval = retval[k,v]
-    return retval
```
jmfranck, line 31: while I was at it, I searched for all instances of `def select_pathway` and imported instead.

pyspecProcScripts/simple_functions.py
aabeaton: this looks fine

pyspecProcScripts/integrate_limits.py
aabeaton: I am assuming fraction is what you meant above, so I will change that in a commit

aabeaton:
ok I agree with the above, just changed faction to fraction in the doc string in this last commit

examples/demo_integrate_limits_realData.py
```diff
@@ -1,3 +1,4 @@
+"""docstring required"""
```
jmfranck, line 1: please include a docstring here

aabeaton: ok it has been added

jmfranck:
the example "Validate Inactive CT Error" breaks -- checking on master

jmfranck:
I think that @guinneak198 should be able to help with fixing this -- it seems like the `integral_w_errors` function isn't returning limits that match with the new function (which probably means that it's not using the new function the way it should be).   I hope that by quickly touching base, the two of you can resolve this very fast.

jmfranck:
(this came up b/c before closing a PR, I always run the `build_docs.sh` in `doc_src` to make sure all the examples build)

examples/test_error.py
```diff
@@ -101,13 +101,17 @@
...
+            convolve_method='Gaussian')
     s_integral = s["t2":frq_slice].C # the "official" copy of the integral
     s_integral = select_pathway(s_integral, signal_pathway)
     s_integral.integrate("t2")
     avg_d = s_integral.C.mean().item()
     s_integral /= avg_d
+    # the line below changes the frequency slice for some reason
     s /= avg_d
+    frq_slice = integrate_limits(select_pathway(s, signal_pathway),
+            convolve_method='Gaussian')
```
aabeaton, line 114: after dividing by `avg_d` for some reason the `frq_slice` changes... in lines 104-105, it is returned as `-43.73959504  55.45762978` but after line 112 it is returned as `-43.73959504  75.29707474` - not sure why a normalization would have that effect, but by setting `frq_slice` after the division, this matches the `frq_slice_check` returned by `integral_w_errors` and satisfies the assert condition

examples/test_error.py
```diff
@@ -101,17 +101,14 @@
...
     s_integral = s["t2":frq_slice].C # the "official" copy of the integral
     s_integral = select_pathway(s_integral, signal_pathway)
     s_integral.integrate("t2")
-    avg_d = s_integral.C.mean().item()
+    avg_d = s_integral.C.mean().real.item()
     s_integral /= avg_d
     # the line below changes the frequency slice for some reason
     s /= avg_d
-    frq_slice = integrate_limits(select_pathway(s, signal_pathway),
-            convolve_method='Gaussian')
```
jmfranck: b/c you were dividing by a complex number (I remove the comment in subsequent commit)

examples/test_error.py
```diff
@@ -101,17 +101,14 @@
     # }}}
 
     # {{{Normalization
-    frq_slice = integrate_limits(select_pathway(s, signal_pathway),
-            convolve_method='Gaussian')
+    frq_slice = integrate_limits(select_pathway(s, signal_pathway))
```
jmfranck: it's confusing to have the `integrate_w_error` to select the non-default option for `integrate_limits` by default, so I modify the former instead of what you did here.

examples/test_error.py
```diff
@@ -160,7 +155,7 @@
     # }}}
 
     # {{{ Calculating propagated error along active CT on noise slice
-    active_error,_ = active_propagation(s, signal_pathway, indirect="nScans")
+    active_error = active_propagation(s, signal_pathway, indirect="nScans")
```
jmfranck: don't know what this was -- it broke the script for me.

pyspecProcScripts/integral_w_error.py
```diff
@@ -27,7 +27,6 @@ def integral_w_errors(s,sig_path,error_path, indirect='vd', direct='t2',fl=None,
     """
     assert s.get_ft_prop(direct), "need to be in frequency domain!"
     frq_slice = integrate_limits(select_pathway(s,sig_path),
-            convolve_method='Gaussian',
```
jmfranck: I have this function use the default for `integrate_limits` --> might make sense to add this as a kwarg for this function, which I didn't do.

examples/test_error.py
aabeaton: weird, mine did not run without that... active_propagation on my computer wanted to return 2 things, but we only cared about the first one

examples/test_error.py
aabeaton: `gaussian` is the default - that was more of a trouble shooting thing that we left in, so it is fine to remove

examples/test_error.py
jmfranck: Does it run for you now?

aabeaton: It does. I do not know why` active_propagation` it was returning `frq_slice` but it no longer is

pyspecProcScripts/integral_w_error.py
guinneak198: made into kwarg

guinneak198:
builds in sphinx and good on my end

PR #59, opened 2021-07-07
example for hermitian phasing

guinneak198:
Demonstrates the Hermitian Function test on IR and enhancement fake data with a echo of 10 ms


examples/Hermitian_Phasing.py
```diff
@@ -0,0 +1,76 @@
...
+        data /= ph0
+        fl.next("Zeroth order phasing correction applied")
+        fl.image(data)
+        #{{{ Applying the phase corrections
+        best_shift, max_shift = hermitian_function_test(
+            select_pathway(data.C.mean(indirect), signal_pathway),fl=fl)
+        data.setaxis("t2", lambda x: x - best_shift).register_axis({"t2": 0})
+        fl.next('After Hermitian Test')
+        fl.image(data)
+        fl.show();quit()
```
jmfranck, line 75: In general, you should really not use quit in code that you commit.  Also, you actually can't use quit within examples *at all* --> the sphinx will not build.

examples/Hermitian_Phasing.py
```diff
@@ -0,0 +1,76 @@
...
+        mysgn = select_pathway(myslice, signal_pathway).real.sum("t2").run(np.sign)
+        data *= mysgn
+        data = data["t2":f_range]
+        data.ift('t2')
+        rough_center = abs(data)['ph1',0]['ph2',1].C.convolve('t2',0.01).mean_all_but('t2').argmax('t2').item()
+        logger.info(strm('Rough center is:',rough_center))
+        data.setaxis('t2', lambda x: x - rough_center).register_axis({"t2": 0})
+        ph0 = select_pathway(data, signal_pathway)["t2":0]
+        ph0 /= abs(ph0)
+        data /= ph0
```
jmfranck, line 66: why are you rough centering and zeroth-order phasing? Do you always do this?

examples/Hermitian_Phasing.py
```diff
@@ -0,0 +1,76 @@
...
+                ("ph1", nddata(r_[0:4] / 4.0, "ph1")),
+                ("t2", nddata(r_[0:0.2:256j] - echo_time, "t2")),
+            ],
+            {"ph1": 1},
+            "power",
+            "enhancement",
+        ),
+    ]:
+        fl.basename = label
+        data = fake_data(expression, OrderedDict(orderedDict), signal_pathway)
```
jmfranck, line 50: again, you were not on the most recent version of master -- this seems to have been forked off of da6b0dcc4b0b4727761fb48063ba19cda2a1b9d7, which is way back from 6/11 --> maybe you hadn't pulled master before making this branch?  Anyways, I merged in master to get it to work

examples/Hermitian_Phasing.py
```diff
@@ -30,21 +30,6 @@
...
-            ),
-            [
-                ("power", nddata(r_[0:4:25j], "power")),
-                ("ph1", nddata(r_[0:4] / 4.0, "ph1")),
-                ("t2", nddata(r_[0:0.2:256j] - echo_time, "t2")),
-            ],
-            {"ph1": 1},
-            "power",
-            "enhancement",
-        ),
```
jmfranck: this actually failed -- you want to ad in an example that does an enhancement curve, but this was actually not run b/c of the quit statement that you had at the end!

examples/Hermitian_Phasing.py
```diff
@@ -67,10 +52,13 @@
         fl.next("Zeroth order phasing correction applied")
         fl.image(data)
         #{{{ Applying the phase corrections
+        fl.basename = "(Hermitian diagnostic)"
         best_shift, max_shift = hermitian_function_test(
             select_pathway(data.C.mean(indirect), signal_pathway),fl=fl)
+        fl.basename = None
```
jmfranck: I use the basename to clearly label the diagnostic plots.

note that the xlims of the diagnostic plot are chosen very weirdly.

examples/Hermitian_Phasing.py
```diff
@@ -67,10 +52,13 @@
...
         data.setaxis("t2", lambda x: x - best_shift).register_axis({"t2": 0})
-        fl.next('After Hermitian Test')
+        fl.next('After Hermitian Test, time domain')
+        fl.image(data)
+        data.ft("t2")
+        fl.next('After Hermitian Test, frequency domain')
         fl.image(data)
```
jmfranck: added in frequency domain, as well -- so what shows up actually does work, so the issue might be your implementation elsewhere

examples/Hermitian_Phasing.py
guinneak198: oh shoot good to know sorry about that old habits die hard!

examples/Hermitian_Phasing.py
guinneak198: yes I typically zeroth order phase correct AFTER the hermitian but after discussing with alec he does it before, and he said the rough centering is pretty important for this version of the hermitian function test (sometimes falls apart or doesn't work if it isn't already roughly centered prior)

examples/Hermitian_Phasing.py
guinneak198: thank you I really don't know what's going on with my master because as shown on slack I am doing every step you say (pull from master make new branch off master yadda yadda)

examples/Hermitian_Phasing.py
guinneak198: so without the quit statement it runs on my end!

examples/Hermitian_Phasing.py
guinneak198: I'm sorry for needing an explanation I'm still pretty hand wavy with what the hermitian function test is doing other than centering the echo at 0. I'm assuming you are talking about the line with: `xlim(np.array(rel_shift)*max_shift/1e-3)` and I think that is using the relative shift that we input and goes to the max shift so what is weird about this?

examples/Hermitian_Phasing.py
guinneak198: I think you are right. So is this example necessary if it is shown in the correlation alignment example?

guinneak198:
when I pulled this after switching it added a bunch of stuff.... I'm gonna attempt to push but this may end up adding more than we want...
image

examples/Hermitian_Phasing.py
jmfranck: so, I think we fixed the issue with master now?

examples/Hermitian_Phasing.py
jmfranck: I was just looking at the plot.  I think you just don't want to set the xlims, so that it does the thing where the data automatically fills the window.

examples/Hermitian_Phasing.py
jmfranck: yes -- for two reasons:

- in case it seems like it's doing something funky, and we want to check (the reason we're using it now)
- in case someone *just* wants to phase their data, without correlation, etc -- which is a lot of the time, actually

pyspecProcScripts/phasing.py
```diff
@@ -336,7 +337,7 @@ def real_imag_mirror(forplot, ax):
         real_imag_mirror(forplot, ax2)
         ax2.set_title('shift=0')
         #}}}
-        fl.next('cost functions')
+        fl.next('Hermitian Diagnostic - Cost Functions')
```
jmfranck, line 340: so, this plot is the one with the xlim problem that we discuss in the other comments.  This is the only real problem with the code that I see offhand.

jmfranck (commented):
I think just see the comments about xlim and review my responses to your questions, etc, and we're done.

examples/Hermitian_Phasing.py
guinneak198: yes all good!

examples/Hermitian_Phasing.py
guinneak198: okay so I think by getting rid of the xlims it's fixed?

pyspecProcScripts/phasing.py
guinneak198: okay and getting rid of the xlims fixes the issue :)

jmfranck (commented):
I changed a bunch of stuff -- definitely look at those to see what I'm doing.

Before merging, please fix the logger errors that show up in the "out" box -- I think you are passing multiple arguments to logging.debug or logging.info without wrapping them with strm fist.

examples/Hermitian_Phasing.py
```diff
@@ -0,0 +1,102 @@
...
+                ("ph1", nddata(r_[0:4] / 4.0, "ph1")),
+                ("t2", nddata(r_[0:0.2:256j] - echo_time, "t2")),
+            ],
+            {"ph1": 1},
+            "power",
+            "enhancement",
+        ),
+    ]:
+        fl.basename = "(%s)" % label
+        fig, ax_list = subplots(1, 4, figsize=(20, 20))
```
guinneak198: I got rid of the figsize because it was cutting off the x axis name and so was slightly confusing. Without the figsize it looks good on my end

guinneak198 (commented):
okay looks good on my end and I ran the build shell script to ensure everything runs and it looks good on my end!

jmfranck:
just decrease the figure size on these images, so that the text is legible (leave aspect ratio alone, but decrease ~ 3 fold?)
image

guinneak198:
okay reduced from (20,20) to (7,7) and I think it looks okay!

PR #62, opened 2021-08-25
Manual enhancement processing example scripts

bett8500:
Two new example scripts were constructed: one to process enhancement data for individual samples - `manual_enhancement.py,` and another for a series of enhancement datasets from separate samples - `manual_enhancement_multiple.py` (should be used after `manual_enhancement.py`). Both have the capability to correct for concentration, T1(p=0), T1(p) either loaded from a csv file or entered manually, and ppt value (NMR over ESR frequency). They also have the option of saving csv files and useful figures.
Minimal changes were made to `load_data.py` as well to enable the processing of older 8-step phase-cycled data.


examples/manual_enhancement.py
```diff
@@ -0,0 +1,263 @@
...
+    if export_csv:
+        import pandas as pd
+
+        reenhancementdf = pd.DataFrame(data=s.real.data, columns=["Re[E(p)]"])
+        imenhancementdf = pd.DataFrame(data=s.imag.data, columns=["Im[E(p)]"])
+        enhancementdf = pd.DataFrame(data=s.data, columns=["enhancement"])
+        powerdf = pd.DataFrame(data=s.getaxis("power"), columns=["power"])
+        enhancement = powerdf.join(reenhancementdf)
+        enhancement2 = enhancement.join(imenhancementdf)
+        enhancement3 = enhancement2.join(enhancementdf)
```
jmfranck, line 209: see if [this](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.from_dict.html) works here, instead

pyspecProcScripts/load_data.py
```diff
@@ -385,8 +385,7 @@ def proc_spincore_ODNP_v1(s,fl=None):
     nPoints = s.get_prop('acq_params')['nPoints']
     SW_kHz = s.get_prop('acq_params')['SW_kHz']
     nScans = s.get_prop('acq_params')['nScans']
-    nPhaseSteps = s.get_prop('acq_params')['nPhaseSteps']
-    s.chunk('t',['ph1','t2'],[4,-1])
+    #s.chunk('t',['ph1','t2'],[4,-1])
     s.set_units('t2','s')
```
jmfranck, line 389: you can't make this change -- it will affect how other files are loaded.

Possibly, you need to use a different version to load your files?  Ask @guinneak198

bett8500: I can't load my files with that line, so yes I would need a different version.

jmfranck:
I am closing this as a pull request, which will  not affect the existence of the branch.

That's b/c I believe this is the "complicated" version of the code, and we eventually want a pull request for a simpler version.

PR #63, opened 2021-08-27
New hermitian

guinneak198:
updated hermitian function integrated into phasing.py and in example.

### DIFF
--- a/examples/Hermitian_Phasing.py
+++ b/examples/Hermitian_Phasing.py
@@ -72,6 +72,7 @@ with figlist_var() as fl:
         data /= zeroth_order_ph(select_pathway(data,signal_pathway), fl=fl)
         fl.image(data, ax=ax_list[1], human_units=False)
         ax_list[1].set_title("Zeroth Order \n Phase Corrected")
+        fl.basename = "(%s)"%label
         best_shift = hermitian_function_test(
             select_pathway(data.C.mean(indirect), signal_pathway), 
             fl=fl


guinneak198:
okay reviewed and edited accordingly. I also ran and it runs on my end. For formatting of the fake data portion I copied from this hash: dd68594ac896d4f028e2cd27b6a023989d64e793 hopefully that's correct format for indents and such at the loading data stage

pyspecProcScripts/phasing.py
```diff
@@ -172,190 +172,126 @@ def costfun(ph1):
...
+        Slice that extends to about 2 times the time domain signal.
+    ini_delay:       float
+        initial delay
+    band_mask_no:    int
+        Helps determing number of points for rectangular mask
+    band_mask:       boolean
+        Masks out aliasing components around the middle of the 
+        selection range.
+        When true, uses rectangular shape for applied mask.
+        When false, uses triangular shape for applied mask.
```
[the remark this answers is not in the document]
jmfranck: yes

pyspecProcScripts/phasing.py
```diff
@@ -274,7 +284,7 @@ def hermitian_function_test(s, selection_range=(None,0.04), ini_delay = 0e6,
         ax_list[5].set_title('Masked Residual -- Relabeled')
         xlim(0,15)
         fig.tight_layout()
-    residual.mean('t2')
+    residual.mean(direct)
     residual.set_units('center','s')
     best_shift = residual['center':(500e-5,None)].C.argmin('center').item()
```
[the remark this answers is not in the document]
jmfranck: that's fine, but here, you are hard-coding that the echo is centered at 5 ms or later -- that's pretty far!  If I understand correctly, you argument says that this should be set to 1 ms.

examples/Hermitian_Phasing.py
```diff
@@ -23,27 +23,27 @@
...
+            ],    
+            {"ph1": 0, "ph2": 1},    
             "vd",
             "IR",
-        ),
-        (
-            (
+        ),    
+        (    
+            (
```
jmfranck, line 46: it's showing that you added spaces at the end of the line.  Just get rid of them so it matches.

examples/Hermitian_Phasing.py
```diff
@@ -56,8 +56,8 @@
             {"ph1": 1},
             "power",
             "enhancement",
-        ),
-    ]:
+        ),    
+    ]:
```
jmfranck, line 60: same comment as above.

pyspecProcScripts/phasing.py
```diff
@@ -172,190 +172,150 @@ def costfun(ph1):
...
-    r"""determine the center of the echo
+def hermitian_function_test(s, 
+        direct = 't2',
+        frq_range=(-5e3/2,5e3/2),#assumes signal is somewhat on resonance
+        selection_range=(None,0.04),
+        ini_delay = 0e6, 
+        band_mask = 0,
+        final_range = (500e-5,None),
+        fl=None):
+    r"""determine the center of the echo via
```
jmfranck, line 183: description missing?

pyspecProcScripts/phasing.py
```diff
@@ -172,190 +172,126 @@ def costfun(ph1):
...
-    #I want to do this center_point = s_hermitian['t2':0], but why not be consistent
-    s_hermitian /= center_point/abs(center_point)
-    if fl:
-        fl.image(s_hermitian.C.mean_all_but(['shift','t2']),
-                ax=ax_list[1])
-        ax_list[1].set_title('Hermitian Data')
+    orig_dt = s.getaxis('t2')[1] - s.getaxis('t2')[0]
+    s.ft('t2')
+    s = s['t2':(-5e3/2,5e3/2)] #assumes signal is somewhat on resonance
+    s.ift('t2',pad=2048*8)
```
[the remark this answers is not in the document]
jmfranck: you didn't do this!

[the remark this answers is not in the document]
guinneak198: I don't think you pulled... this says outdated and all there is no 't2' in the hermitian_function_test except in the kwargs where we set the direct to t2 as default...

pyspecProcScripts/phasing.py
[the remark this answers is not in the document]
guinneak198: Again this is changed to not be hard coded so I'm not sure if you pulled before commenting? I changed the kwarg thought to 1 ms instead of 5 ms

pyspecProcScripts/phasing.py
```diff
@@ -172,190 +172,150 @@ def costfun(ph1):
...
+    orig_dt = s.get_ft_prop(direct, "dt")
+    s.ft(direct)
+    s = s[direct:frq_range] 
+    s.ift(direct,pad=2048*8)
+    if fl is not None:
+        fig, ax_list = subplots(6, 1, figsize=(10,10))
+        fl.next('Hermitian Function Test Diagnostics',fig=fig)
+        s = s[direct:(ini_delay,None)]
+        fl.plot(abs(s),'.',ax=ax_list[0])
+        ax_list[0].set_title('Data with Padding')
```
jmfranck, line 226: this set of figures is very hard to see as-is when I run it from the command line.  Can you break it up into a couple different figures, so that we can actually see the subplots?  Or maybe make it into a 2 column grid that's wider in size?

pyspecProcScripts/phasing.py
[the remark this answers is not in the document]
jmfranck: you're right!

pyspecProcScripts/phasing.py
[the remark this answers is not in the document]
jmfranck: yes!

pyspecProcScripts/phasing.py
jmfranck: this is not fixed (I checked the "outdated" here, and it's just b/c I ran black on the code)

jmfranck: (I did make sure I pulled and was on the right branch this time!!!)

guinneak198: oops sorry! okay hopefully now it's good I changed it to a 2 x 3 plot

ipynb/20190501_homogeneous_Liouville.py
```diff
@@ -1,224 +0,0 @@
-
-# coding: utf-8
-
-# In[ ]:
-
-
-from pyspecdata import *
-from scipy.linalg import expm
-
```
jmfranck, line 9: I deleted the ipynb guys b/c they have similar functions defined to the main code, which is confusing.

pyspecProcScripts/phasing.py
```diff
@@ -167,8 +167,8 @@ def ph1_real_Abs(s, dw, ph1_sel=0, ph2_sel=1, fl=None):
         fl.plot(s_cost, ".")
     s_cost.smoosh(["vd", "phcorr"], "transients")
     ph1_opt = np.asarray(s_cost.argmin("transients").item())
-    logging.info(strm("THIS IS PH1_OPT", ph1_opt))
-    logging.info(strm("optimal phase correction", repr(ph1_opt)))
+    logging.debug(strm("THIS IS PH1_OPT", ph1_opt))
+    logging.debug(strm("optimal phase correction", repr(ph1_opt)))
```
jmfranck: at some point, all these info messages become white noise, so I just made them all debug-level statements (so that they show up only in the log).  If there are 1 or 2 that you can justify appearing at the command line, feel free to change back, but there were just too many.

Once the message becomes so long, anyways, it's easiest to go look at the log file, which (unlike the command line) is searchable.

pyspecProcScripts/phasing.py
```diff
@@ -199,7 +199,7 @@ def hermitian_function_test(
     s,
     direct="t2",
     frq_range=(-2.5e3, 2.5e3),  # assumes signal is somewhat on resonance
-    aliasing_slop=3,
+    aliasing_slop=2, # if this is set to 3 rather than 2, the IR version is failing.  That shouldn't be happening!  (Equivalently, if we set ini_delay to 3ms in 7093, this also fails!)
```
jmfranck, line 202: I'm sorry but there are two things going on here at the same time:

- I establish a new convention, where the selection range, which is a bit hard-coded is removed in favor of an "aliasing slop" -- defined in my new docstring
- If I use either this new convention or the old convention, setting the initial part of the slice even slighly far out causes everything to fail -- I don't yet understand why this is, and hope you might have some ideas.

pyspecProcScripts/phasing.py
```diff
@@ -242,16 +238,20 @@ def hermitian_function_test(
...
+    new_dt = s.get_ft_prop(direct, "dt")
+    non_aliased_range = r_[3,-3]*int(orig_dt/new_dt)
+    ini_delay = non_aliased_range[0]*new_dt
+    s = s[direct,non_aliased_range[0]:non_aliased_range[1]]
     if fl is not None:
-        plt.rcParams["figure.figsize"] = [7.5, 3.5]
-        plt.rcParams["figure.autolayout"] = True
         fig, ax_list = subplots(2, 3, figsize=(15, 15))
         fl.next("Hermitian Function Test Diagnostics", fig=fig)
-        s = s[direct:(ini_delay, None)]
```
jmfranck: this was done inside the `if fl is not None` -- which is a big no-no b/c data would be processed different based on whether you called the diagnostic plots or not.

pyspecProcScripts/phasing.py
```diff
@@ -242,16 +238,20 @@ def hermitian_function_test(
     orig_dt = s.get_ft_prop(direct, "dt")
     s.ft(direct)
     s = s[direct:frq_range]
-    s.ift(direct, pad=2048 * 8)
+    s.ift(direct, pad=1024 * 16)
+    new_dt = s.get_ft_prop(direct, "dt")
+    non_aliased_range = r_[3,-3]*int(orig_dt/new_dt)
+    ini_delay = non_aliased_range[0]*new_dt
+    s = s[direct,non_aliased_range[0]:non_aliased_range[1]]
```
jmfranck, line 246: as noted earlier, this convention replaced both the `ini_delay` kwarg and the `selection_range` kwarg.

pyspecProcScripts/phasing.py
```diff
@@ -242,16 +238,20 @@ def hermitian_function_test(
...
-        plt.rcParams["figure.autolayout"] = True
         fig, ax_list = subplots(2, 3, figsize=(15, 15))
         fl.next("Hermitian Function Test Diagnostics", fig=fig)
-        s = s[direct:(ini_delay, None)]
-        fl.plot(abs(s), ".", ax=ax_list[0, 0])
+        fl.plot(abs(s), ax=ax_list[0, 0])
         ax_list[0, 0].set_title("Data with Padding")
-    selection = s[direct:selection_range]
+    probable_center = abs(s).convolve(direct,orig_dt*3).argmax(direct).item() # convolve just for some signal averaging
+    selection = s[direct:(0,probable_center*2)]
```
jmfranck, line 253: to take care of the large side of the `selection_range` kwarg, I do this.

pyspecProcScripts/phasing.py
```diff
@@ -284,6 +284,9 @@ def hermitian_function_test(
         else:
             fl.image(residual, ax=ax_list[0, 1])
         ax_list[0, 1].set_title("shifted and phased")
+    # I tried removing the following line, and the 2D residual actually
+    # looks sharper, but then the 1D plot of the cost function disappears
+    # -- what's up with that?
     residual /= abs(center_point) # weight 'residual' by 1/(signal amplitude)
```
jmfranck: read comment -- I don't understand this --> any insight here?

pyspecProcScripts/phasing.py
guinneak198: agreed

guinneak198: why are you making edits to functions that are not included in this pull request? doesn't that muddy the waters of the primary purpose of fixing/updating hermitian function test?

pyspecProcScripts/phasing.py
guinneak198: okay, my thought is that it was working even if it was taking extra kwargs. We should merge the updated version before you added this aliasing_slop since this is adding more issues and then we can make a separate PR for incorporating it. This PR I feel is getting really convoluted since it was working prior to these aliasing_slop changes...

pyspecProcScripts/phasing.py
```diff
@@ -129,233 +149,219 @@ def ph1_real_Abs(s,dw,ph1_sel=0,ph2_sel=1,fl = None):
...
+    ini_delay = non_aliased_range[0]*new_dt
+    logger.debug(strm("ini delay is",ini_delay))
+    s = s[direct:(non_aliased_range[0],non_aliased_range[1])]
+    if fl is not None:
+        fig, ax_list = subplots(2, 3, figsize=(15, 15))
+        fl.next("Hermitian Function Test Diagnostics", fig=fig)
+        fl.plot(abs(s), ax=ax_list[0, 0])
+        ax_list[0, 0].set_title("Data with Padding")
+    probable_center = abs(s).convolve(direct,orig_dt*3).argmax(direct).item() # convolve just for some signal averaging
+    selection = s[direct:(0,probable_center*3)]
```
guinneak198, line 253: I had to multiply probable_center by 3 instead of 2 as it was cutting it off too short.

pyspecProcScripts/phasing.py
```diff
@@ -129,233 +149,219 @@ def ph1_real_Abs(s,dw,ph1_sel=0,ph2_sel=1,fl = None):
...
-def hermitian_function_test(s, down_from_max=0.5, rel_shift=1.5,shift_points=1200, fl=None):
-    r"""determine the center of the echo
+    # }}}
+
+
+def hermitian_function_test(
+    s,
+    direct="t2",
+    frq_range=(-2.5e3, 2.5e3),  # assumes signal is somewhat on resonance
+    aliasing_slop=3, # if this is set to 3 rather than 2, the IR version is failing.  That shouldn't be happening!  (Equivalently, if we set ini_delay to 3ms in 7093, this also fails!)
```
guinneak198, line 202: to show this works for 3 now

pyspecProcScripts/phasing.py
```diff
@@ -129,233 +149,219 @@ def ph1_real_Abs(s,dw,ph1_sel=0,ph2_sel=1,fl = None):
...
-        x *= dt
+    orig_dt = s.get_ft_prop(direct, "dt")
+    s.ft(direct)
+    s = s[direct:frq_range]
+    s.ift(direct, pad=1024 * 16)
+    new_dt = s.get_ft_prop(direct, "dt")
+    non_aliased_range = r_[-aliasing_slop,aliasing_slop]*int(orig_dt/new_dt)
+    ini_delay = non_aliased_range[0]*new_dt
+    logger.debug(strm("ini delay is",ini_delay))
+    s = s[direct:(non_aliased_range[0],non_aliased_range[1])]
```
guinneak198, line 246: indexing error but works now that I changed the format

guinneak198 (commented):
okay went through and debugged/got it to work for slop=3 also cost function and everything plots as it should

pyspecProcScripts/phasing.py
jmfranck: Hmm.  This isn’t right.  You’re treating it like these are time values, but they’re not, they are integer indices

pyspecProcScripts/phasing.py
jmfranck: But it shouldn’t be.  Can you give an example of where this happens?

guinneak198: the selection range as originally defined was based on the time domain. the probably center it takes is about 0.01 but looking at the t domain the signal lasts to about 0.03. By multiplying by 2 we are cutting out parts of the signal in the time domain. If we multiply by 3 it is included. you can also see the phasing looks better when multiplied by 3 instead of 2 as you originally had it

pyspecProcScripts/phasing.py
guinneak198: right but how you had it before you were indexing two different points instead of a range. This is now slicing based on the index values

pyspecProcScripts/phasing.py
```diff
@@ -199,7 +199,7 @@ def hermitian_function_test(
     s,
     direct="t2",
     frq_range=(-2.5e3, 2.5e3),  # assumes signal is somewhat on resonance
-    aliasing_slop=2, # if this is set to 3 rather than 2, the IR version is failing.  That shouldn't be happening!  (Equivalently, if we set ini_delay to 3ms in 7093, this also fails!)
+    aliasing_slop=3,
```
jmfranck: really 2 is a fine choice, I'm just showing that 3 works as well, now.

pyspecProcScripts/phasing.py
```diff
@@ -241,9 +241,9 @@ def hermitian_function_test(
     s.ift(direct, pad=1024 * 16)
     new_dt = s.get_ft_prop(direct, "dt")
     non_aliased_range = r_[aliasing_slop,-aliasing_slop]*int(orig_dt/new_dt)
-    ini_delay = non_aliased_range[0]*new_dt
-    logger.debug(strm("ini delay is",ini_delay))
     s = s[direct,non_aliased_range[0]:non_aliased_range[1]]
+    ini_delay = s.getaxis(direct)[0]
```
jmfranck, line 245: I was worried there was a mismatch here, so I calculate it differently.

pyspecProcScripts/phasing.py
```diff
@@ -282,6 +273,15 @@ def hermitian_function_test(
...
         )
+    dt = residual.get_ft_prop(direct, "dt")
+    # the shifts themselves run from 0 to mid_idx -- the echo-centers
+    # these correspond to are different. Also, we're not really
+    # worried about the sub-integer shift here, because we can just
+    # use sinc interpolation before we start -- see:
+    # https://jmfrancklab.slack.com/archives/CLMMYDD98/p1623354066039100
+    shifts = nddata(dt * (r_[0:mid_idx]), "shift")
+    shifts.set_units("shift", "s")
+    logger.debug(strm("Length of shifts dimension:", ndshape(shifts)["shift"]))
```
jmfranck, line 284: here I'm just moving the position of some of the plots

pyspecProcScripts/phasing.py
```diff
@@ -302,7 +302,6 @@ def hermitian_function_test(
     # I tried removing the following line, and the 2D residual actually
     # looks sharper, but then the 1D plot of the cost function disappears
     # -- what's up with that?
-    residual /= abs(center_point) # weight 'residual' by 1/(signal amplitude)
```
jmfranck: so... I'm not positive if this is a good thing for not (probably best to get @aabeaton to test on some real data), but removing this does give a much sharper cost function

pyspecProcScripts/phasing.py
```diff
@@ -329,7 +328,7 @@ def hermitian_function_test(
         mask = mask.astype(float)
         norm = np.floor(mid_idx - B)
         norm = norm.astype(float)
-        mask /= (norm) * 2
+        mask /= ((norm) * 2)**2
```
jmfranck: this is a another change that I just made pretty empirically, but it does seem to prevent the low residual for the low choice of echo center -- roughly here, the argument is that it's harder to match more points, and we seemed to be insufficiently accounting for that.  This one I'm really not sure about but worth @aabeaton testing on real data?

pyspecProcScripts/phasing.py
```diff
@@ -358,8 +357,10 @@ def hermitian_function_test(
         fl.twinx(orig=False, color="red")
         residual.name("cost function")
         fl.plot(residual, color="r", alpha=0.5, human_units=False)
-        ylim(0, residual["center" : (best_shift - 4e-3, best_shift)].data.max())
-        axvline(x=best_shift, c="k", linestyle=":")
+        fl.plot(residual["center" : (best_shift - 4e-3, best_shift)], ':', alpha=0.5, human_units=False)
```
jmfranck, line 360: it was hard to distinguish the best shift from the other lines, so I changed the plotting style.

jmfranck:
`test_error.py` doesn't run -- it seems like you are passing a multi-dimensional array to the hermitian function, and that it doesn't like this? We should verify that that's what's up and what we want to do about it.

examples/correlation_alignment_example.py
```diff
@@ -66,21 +66,10 @@
...
-            abs(select_pathway(data, signal_pathway))
-            .C.convolve("t2", 0.3e-3)
-            .mean_all_but("t2")
-            .argmax("t2")
-            .item()
-        )
-        logger.info(strm("Rough center is:", rough_center))
-        data.setaxis("t2", lambda x: x - rough_center).register_axis({"t2": 0})
-        data.ft("t2")
-        data.ift("t2")
```
jmfranck, line 79: this makes it not work.  While it shouldn't *help*, it also shouldn't *break* the code.  I think that the way the code is written, it doesn't deal well if it starts at a value that's not near 0?

pyspecProcScripts/correlation_alignment.py
```diff
@@ -96,12 +96,12 @@ def correl_align(s_orig, align_phases=False,tol=1e-4,indirect_dim='indirect',
...
-    logging.info(strm("Center frequency", nu_center))
+    logging.debug(strm("Center frequency", nu_center))
     for my_iter in range(100):
         i += 1
-        logging.info(strm("*** *** ***"))
-        logging.info(strm("CORRELATION ALIGNMENT ITERATION NO. ",i))
-        logging.info(strm("*** *** ***"))
+        logging.debug(strm("*** *** ***"))
+        logging.debug(strm("CORRELATION ALIGNMENT ITERATION NO. ",i))
+        logging.debug(strm("*** *** ***"))
```
jmfranck, line 104: just eliminating noise

pyspecProcScripts/phasing.py
```diff
@@ -201,7 +201,6 @@ def hermitian_function_test(
     frq_range=(-2.5e3, 2.5e3),  # assumes signal is somewhat on resonance
     aliasing_slop=3,
     band_mask=False,
-    final_range=(100e-5, None),
```
jmfranck: we can't do this b/c it completely freaks out of the axis starts w/ a negative number (which we should allow)

examples/correlation_alignment_example.py
jmfranck: fixed below

guinneak198: I don't see how this pertains to the updated hermitian function

guinneak198:
https://github.com/jmfrancklab/proc_scripts/pull/63#issuecomment-914640113 test_error does not pertain to this PR

examples/correlation_alignment_example.py
jmfranck: you are always supposed to make sure you can rebuild the sphinx gallery,  and it had broken this.

pyspecProcScripts/phasing.py
```diff
@@ -129,233 +149,220 @@ def ph1_real_Abs(s,dw,ph1_sel=0,ph2_sel=1,fl = None):
...
-    r = minimize(costfun,
-            ph1_opt,
-            bounds = [(ph1_opt-dx,ph1_opt+dx)])
+
+    r = minimize(costfun, ph1_opt, bounds=[(ph1_opt - dx, ph1_opt + dx)])
     assert r.success
-    s = applyphase(s,r.x.item())
-    fl.plot(r.x,r.fun,'x')
+    s = applyphase(s, r.x.item())
+    fl.plot(r.x, r.fun, "x")
```
jmfranck, line 192: here and above, I had just run black and gotten rid of excessive info  statements.

pyspecProcScripts/correlation_alignment.py
```diff
@@ -152,19 +152,19 @@ def correl_align(s_orig, align_phases=False,tol=1e-4,indirect_dim='indirect',
...
+        logging.debug(strm('signal energy per transient (recalc to check that it stays the same):',(abs(s_copy**2).data.sum().item() / N)))
 
         this_E = (abs(s_copy.C.sum(indirect_dim))**2).data.sum().item() / N**2
         energy_vals.append(this_E / sig_energy)
-        logging.info(strm('averaged signal energy (per transient):', this_E))
+        logging.debug(strm('averaged signal energy (per transient):', this_E))
         if last_E is not None:
             energy_diff = (this_E - last_E)/sig_energy
-            logging.info(strm(energy_diff))
+            logging.debug(strm(energy_diff))
```
jmfranck, line 167: again, this actually came from building sphinx -- I see all this noise printed on the gallery page, and so I get rid of it.

guinneak198:
Okay so it takes FOREVER to run so I took out the zeroth order phasing and focused on producing the table of programmed tau, estimated tau, and the difference thereof. I installed tabulate for this purpose and the resulting table looks nice. My only concern is the time it takes to run so I'm not sure if my loop/copies are most efficient. I include a copy in this line simply because without the copy it says 'tau' does not exist in the data for the second tau time. 
`this_data = tau_data.C`

guinneak198:
I further extended the time domain to 2.5 times the longest tau and see SOME improvement. Some of the longer echoes still have absurd estimates but I'm not sure what else we can do as it is only these echoes that give estimates with a difference more than 5 ms: 35.67 ms, 65.67 ms, 70.67 ms, 85.67 ms, and 90.67-105.67 ms echoes.  However, the cost functions now are plotted individually rather than all on one plot.

guinneak198:
looks good and builds in sphinx on my end

examples/Hermitian_Phasing_Actual_var_tau.py
```diff
@@ -57,5 +56,5 @@
         diff = abs(best_shift - programmed_tau)
         table[j+1].append(str(diff))
     logger.info(strm(tabulate(table)))
-    fl.show();quit()
+    fl.show()
```
jmfranck: (sad face)

pyspecProcScripts/phasing.py
```diff
@@ -237,32 +237,50 @@ def hermitian_function_test(
     orig_dt = s.get_ft_prop(direct, "dt")
     if not s.get_ft_prop(direct):
         s.ft(direct)
-    s.ift(direct, pad=1024 * 4)
+    s.ift(direct, pad=ndshape(s)[direct]*50)
```
jmfranck: rather than absolute, set this up as an upsampling

pyspecProcScripts/phasing.py
```diff
@@ -237,32 +237,50 @@ def hermitian_function_test(
...
     new_dt = s.get_ft_prop(direct, "dt")
     non_aliased_range = r_[aliasing_slop,-aliasing_slop]*int(orig_dt/new_dt)
     if aliasing_slop > 0:
         s = s[direct,non_aliased_range[0]:non_aliased_range[1]]
+    s_envelope = s.C.mean_all_but(direct).run(abs)
```
jmfranck, line 245: I'm trying to clarify the code here.

pyspecProcScripts/phasing.py
```diff
@@ -237,32 +237,50 @@ def hermitian_function_test(
...
     if fl is not None:
         fl.push_marker()
         fig, ax_list = subplots(2, 2, figsize=(15, 15))
         fl.next("Hermitian Function Test Diagnostics", fig=fig)
-        fl.plot(abs(s), ax=ax_list[0, 0], human_units=False)
+        fl.plot(s_envelope, ax=ax_list[0, 0], human_units=False)
         ax_list[0, 0].set_title("Data with Padding")
-    signal_range = abs(s).mean_all_but(direct).contiguous(lambda x: x > 0.5 * abs(x).data.max())
```
jmfranck: we had agreed you were going to go down to 1/4

pyspecProcScripts/phasing.py
```diff
@@ -237,32 +237,50 @@ def hermitian_function_test(
...
+    # {{{ the peak triple gives the left and right thresholds, with the
+    #     peak max in the middle
+    #     we move the left point as needed to make sure that peak is in
+    #     the first half of the resulting slice
+    peak_triple = r_[ peak_triple[0],
+            s[direct:peak_triple].mean_all_but(direct).run(abs).argmax().item(),
+            peak_triple[1] ]
     if fl is not None:
+        for j in range(3):
+            ax_list[0, 0].axvline(x=peak_triple[j], ls=':', color='k', alpha=0.5, linewidth=2)
```
jmfranck, line 266: your code was choosing bad ranges and I needed to do this to diagnose what was going on.

jmfranck: (they were frequently only a very small slice of the signal)

jmfranck: (and the cost function relies on averaging the rms from many points)

pyspecProcScripts/phasing.py
```diff
@@ -237,32 +237,50 @@ def hermitian_function_test(
...
+    # }}}
+    if fl is not None:
+        ax_list[0,0].axvspan(0,orig_dt,color='k',alpha=0.1)
+        for j in range(3):
+            ax_list[0, 0].axvline(x=peak_triple[j], color='k', alpha=0.5, linewidth=2)
         fl.plot(abs(s), ':', ax=ax_list[0, 0], linewidth=4, human_units=False)
-    ini_delay = s.getaxis(direct)[0]
-    logger.debug(strm("ini delay is",ini_delay))
+    direct_startpoint = s.getaxis(direct)[0]
+    logger.debug(strm("ini delay is",direct_startpoint))
```
jmfranck: I rename this for clarity -- there is a "startpoint" ft property that also refers to the same concept (start of the axis that doesn't start at 0)

pyspecProcScripts/phasing.py
```diff
@@ -312,17 +330,17 @@ def hermitian_function_test(
         ["shift", direct]
     )
     if fl is not None:
-        fl.image(s, ax=ax_list[0, 1],human_units=False)
+        fl.image(s, ax=ax_list[0, 1])
```
jmfranck: seemed gratuitous.

pyspecProcScripts/phasing.py
```diff
@@ -364,7 +382,7 @@ def hermitian_function_test(
         s.name("cost function")
         fl.plot(s, color="r", alpha=0.5, human_units=False)
         cost_scale = s['center':(-4*orig_dt+best_shift,4*orig_dt+best_shift)].max().item()
-        #ylim((0,cost_scale))
+        ylim((0,cost_scale))
```
jmfranck: this was deliberate -- please leave in

PR #64, opened 2021-09-24
Table integrals

guinneak198:
properly phase and integrate a set of data with kwargs to align and add error bars


guinneak198:
pretty sure sphinx builds. I added subplots for the function to clean it up a bit. correlating and integrating WITH error are kwargs.

pyspecProcScripts/integral_w_error.py
```diff
@@ -18,6 +13,9 @@ def integral_w_errors(s,sig_path,error_path, indirect='vd', direct='t2',fl=None,
     error_path: dict
                 Dictionary of all coherence pathways that are 
                 not the signal pathway.
```
jmfranck, line 15: don't do this -- rather pass the pathways you DON'T want used for the error -- I had written code for this in one of the examples using some set - based math, so you should be able to reuse that.

jmfranck: we decided that we don't want to change this,  but we should make a note for the user to look for examples that contain the variable `excluded_pathways`, so that they can easily specify excluded (rather than  included) pathways

pyspecProcScripts/table_integrals.py
```diff
@@ -0,0 +1,121 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+from sympy import symbols, latex, Symbol
+from pyspecProcScripts import *
+from .simple_functions import select_pathway
+from .correlation_alignment import correl_align
+from .integral_w_error import integral_w_errors
+
+t2 = symbols('t2')
+logger = init_logging('info')
```
jmfranck, line 14: don't do this.

jmfranck:
please make sure that sphinx builds + nice figures are chosen for the gallery

guinneak198:
So the examples table_integrals.py and table_integrals_RealData.py run on my end. I attempted to rebuild the sphinx but got an error that the scaled transform example was missing documentation added it in and it seems to build on my end!

pyspecProcScripts/table_integrals.py
```diff
@@ -0,0 +1,254 @@
...
+    t_start = t_range[-1] / 4
+    t_start *= 3
+    rx_offset_corr = s['t2':(t_start,None)]
+    rx_offset_corr = rx_offset_corr.data.mean()
+    s -= rx_offset_corr
+    s.ft(direct)
+    s.ft(list(signal_pathway))
+    #}}}
+    zero_crossing = abs(select_pathway(s[direct:f_range],signal_pathway)).C.sum(direct).argmin(indirect,raw_index=True).item()
+    print("ZERO CROSSING IS",zero_crossing)
```
jmfranck, line 95: we have the function that determiens the sign of the signal --  why is it doing this?

examples/table_of_integrals.py
```diff
@@ -17,33 +17,64 @@
 seed(2021)
 rcParams['image.aspect'] = 'auto' # needed for sphinx gallery
 
-# sphinx_gallery_thumbnail_number = 1
+# sphinx_gallery_thumbnail_number = 7
```
jmfranck, line 20: when I run this, I don't get the nice image with subplots that we had gotten previously -- is the issue that you have the code to do that with the old-style image plot, but not with DCCT?

pyspecProcScripts/integral_w_error.py
jmfranck: when I say "make a note" I mean to add to the docstring here

guinneak198 (commented):
cleaning up a bit. Updated DC offset correction

PR #68, opened 2022-01-20
field sweep plots ppt values instead of field

guinneak198:
processing script for field sweep with the x axis as the ppt value using effective gamma as hard coded variable. Maybe we can make that an acq_param in the ppg?


proc_fieldSweep.py
```diff
@@ -0,0 +1,74 @@
...
+            'field_sweep_1','field_sweep_v1',
+            'TEMPO field sweep',(-250,250)),
+        ]:
+    s = find_file(thisfile,exp_type=exp_type,expno=nodename)
+    #{{{Obtain ESR frequency and chunk/reorder dimensions
+    v_esr = (s.get_prop('acq_params')['mw_freqs'][0])/1e9
+    s.reorder('t',first=True)
+    s.chunk('t',['ph1','t2'],[4,-1])
+    s.setaxis('ph1',r_[0.,1.,2.,3.]/4)
+    s.reorder('t2',first=False)
```
jmfranck, line 21: this is preprocessing

proc_fieldSweep.py
```diff
@@ -0,0 +1,74 @@
...
+    fl.next('frequency domain raw data')
+    fl.image(s)
+    s = s['t2':(-1e3,1e3)]
+    s=s['ph1',1]['power',0]['nScans',0].C
+    s.ift('t2')
+    s.ft('t2')
+    s = s['t2':(-filter_bandwidth/2,filter_bandwidth/2)]
+    s.ift('t2')
+    rough_center = abs(s).C.convolve('t2',0.0001).mean_all_but('t2').argmax('t2').item()
+    s.setaxis(t2-rough_center)
```
jmfranck, line 44: use hermitian, not rough center

proc_fieldSweep.py
```diff
@@ -0,0 +1,74 @@
...
+    s.ft('t2')
+    fl.next('line plots')
+    for z in range(len(s.getaxis('Field'))):
+        fl.plot(abs(s['Field',z]),label='%d'%z)
+    s_ = s['t2':freq_slice].sum('t2')
```
jmfranck, line 49: messy notation.

proc_fieldSweep.py
```diff
@@ -0,0 +1,74 @@
...
+    #}}}
+    #{{{Convert field to v_NMR
+    field = s_.getaxis('Field')
+    new_field = field * gamma_eff
```
jmfranck, line 53: the [relevant task](https://www.wrike.com/open.htm?id=826379511) specifically notes not to use the Hall probe, which is doing.  Rather use the carrier frequency and offset of the NMR signal.

proc_fieldSweep.py
```diff
@@ -0,0 +1,74 @@
...
+filter_bandwidth = 20e3
+gamma_eff = (14.903537/3507.48) # MHz / G
+for thisfile,exp_type,nodename,postproc,label_str,freq_slice in [
+        ('220112_150uM_TEMPOL_field_dep_1','ODNP_NMR_comp/field_dependent',
+            'field_sweep_1','field_sweep_v1',
+            'TEMPO field sweep',(-250,250)),
+        ]:
+    s = find_file(thisfile,exp_type=exp_type,expno=nodename)
+    #{{{Obtain ESR frequency and chunk/reorder dimensions
+    v_esr = (s.get_prop('acq_params')['mw_freqs'][0])/1e9
```
jmfranck, line 17: is this actually what we call nu_B12?

jmfranck: (it should be, and I think from context that it is)

proc_fieldSweep.py
```diff
@@ -0,0 +1,74 @@
...
+    for z in range(len(s.getaxis('Field'))):
+        fl.plot(abs(s['Field',z]),label='%d'%z)
+    s_ = s['t2':freq_slice].sum('t2')
+    #}}}
+    #{{{Convert field to v_NMR
+    field = s_.getaxis('Field')
+    new_field = field * gamma_eff
+    #}}}
+    #{{{convert x axis to ppt = v_NMR/v_ESR
+    ppt = new_field / v_esr
```
jmfranck, line 56: so, here,  `new_field` should be replaced by `nu_NMR`, which is the carrier either plus or minus the offset frequency of the signal.

proc_fieldSweep.py
```diff
@@ -0,0 +1,74 @@
...
+    rough_center = abs(s).C.convolve('t2',0.0001).mean_all_but('t2').argmax('t2').item()
+    s.setaxis(t2-rough_center)
+    s.ft('t2')
+    fl.next('line plots')
+    for z in range(len(s.getaxis('Field'))):
+        fl.plot(abs(s['Field',z]),label='%d'%z)
+    s_ = s['t2':freq_slice].sum('t2')
+    #}}}
+    #{{{Convert field to v_NMR
+    field = s_.getaxis('Field')
```
jmfranck, line 52: You don't want to use these numbers at all -- they are the settings for the Hall probe, and are therefore inaccurate.

proc_fieldSweep.py
guinneak198: I'm not sure if you mean take the labels out of the line itself and just plot one of the fields? For now I took out the labels

proc_fieldSweep.py
guinneak198: so you're saying to find gamma eff by doing something like:
`gamma_eff = s.get_prop('acq_params')['carrierFreq_MHz']/3507.48` where 3507.48 is the field I tuned to? the goal is to change the 'field' axis to ppt axis (clarifying and correct me if I am wrong). I'm confused on this conversion

guinneak198: I'm confused in what part I am using the Hall probe? Is that when I call getaxis('Field')?

proc_fieldSweep.py
guinneak198: Yes, I will change the name

proc_fieldSweep.py
guinneak198: then how do I get a variance in the x axis? the ppt v_NMR and v_B12 are both one number which would only result in one number along the x axis?

proc_fieldSweep.py
guinneak198: this has been addressed now

proc_fieldSweep.py
jmfranck: (1) `s_` is not a good thing to call a variable (2) is it possible you are creating unnecessary copies?

examples/proc_fieldSweep.py
```diff
@@ -0,0 +1,97 @@
...
+
+A DC offset correction, initial frequency slice 
+and rough centering are applied to a previously
+acquired TEMPOL dataset. It is then sliced more based
+on the 1D diagnostic plot labeled 'Field Slicing'. The 
+field is converted to ppt using the frequency stored as
+an acquisition parameter in the dataset and the effective
+gamma used that day. This will verify the previously required
+ppt value for a given spin label.
+"""
```
jmfranck, line 12: there are 2 files in the PR that conflict -- `examples/proc_fieldSweep.py` and `pro_fieldSweep.py`

proc_fieldSweep.py
guinneak198: I checked and yes s is not used again after this (which I think means yes I was making unnecessary copies?) so I changed s_ to just s

examples/proc_fieldSweep.py
guinneak198: I'm confused, when I "git merge master --no-ff" it just says Already up to date. When I diff against master it shows I have proc_fieldsweep.py and examples/proc_fieldsweep.py as well as the changes in load_data which would be expected as these were not in the master branch. These are new

jmfranck: you shouldn't be adding scripts to the main repo directory.

also, these two files genuinely seem to have the same purpose -- so which one am I reviewing?

guinneak198: resolved in https://jmfrancklab.slack.com/archives/CV553TPRV/p1643747211231269

examples/proc_fieldSweep.py
```diff
@@ -0,0 +1,95 @@
...
+    # }}}
+    #{{{frequency filtering and phase correct
+    s = s['t2':(-1e3,1e3)]
+    s.ift('t2')
+    best_shift = hermitian_function_test(select_pathway(s,signal_pathway))
+    s.setaxis('t2', lambda x: x-best_shift).register_axis({'t2':0})
+    s.ft('t2')
+    s = select_pathway(s,signal_pathway)
+    s /= zeroth_order_ph(s.C.mean('t2'))
+    ## JF review to here
```
jmfranck, line 59: we should come back and look over any changes (since last JMF)  before this point, and read through all the code after this point.

PR #72, opened 2022-03-27
added versions for 4x4 chunking for IR and ODNP

guinneak198:
add a preprocessing for 3 uM where there are 16 phase steps for both IR and ODNP

### DIFF
--- a/pyspecProcScripts/load_data.py
+++ b/pyspecProcScripts/load_data.py
@@ -313,9 +313,13 @@ def proc_Hahn_echoph(s, fl=None):
 
 def proc_spincore_IR(s, fl=None):
     vd_axis = s.getaxis("vd")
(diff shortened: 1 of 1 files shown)


pyspecProcScripts/load_data.py
```diff
@@ -492,6 +517,46 @@ def proc_spincore_ODNP_v2(s, fl=None):
     # }}}
     return s
 
+def proc_spincore_ODNP_v3(s, fl = None):
+    if 't' in s.dimlabels:
+        t.chunk('t',['ph1','t2'],[4,-1])
+        s.setaxis('ph1',r_[0,1,2,3]/4)
+        s.labels({'ph1':r_[0.,1.,2.,3.]/4})
```
jmfranck, line 524: These are redundant, FYI

PR #73, opened 2022-05-11
add cutoff kwarg for integrate_limits

guinneak198:


pyspecProcScripts/integral_w_error.py
```diff
@@ -53,7 +54,7 @@ def integral_w_errors(
     else:
         kwargs = {}
     frq_slice = integrate_limits(
-        select_pathway(s, sig_path), convolve_method=convolve_method, fl=fl
+        select_pathway(s, sig_path), convolve_method=convolve_method, cutoff=cutoff, fl=fl
```
jmfranck, line 57: why not rather change the default of `integrate_limits`?

jmfranck: AG says -- because some are easier than others

PR #74, opened 2022-05-25
Test error ex

guinneak198:
make example with both fake data and real that compares the propagated error estimate from the active CT pathway, from individual inactive CT pathways, the integral_w_error method that averages over all inactive CT pathways, and the traditional numpy std calculation.


pyspecProcScripts/integral_w_error.py
```diff
@@ -85,7 +85,7 @@ def integral_w_errors(
         # mean divides by N₁ (indirect), integrate multiplies by Δf, and the
         # mean sums all elements (there are N₁N₂ elements)
         s_forerror -= s_forerror.C.mean_all_but([indirect, direct]).mean(direct)
-        s_forerror.run(lambda x: abs(x) ** 2 / 2).mean_all_but([direct, indirect]).mean(
+        s_forerror.run(lambda x: abs(x) ** 2).mean_all_but([direct, indirect]).mean(
```
jmfranck, line 88: example `check_integration_error.py` now gives blue bars that don't match the others, and I think this is what breaks it.  You do want to divide by 2 here!

pyspecProcScripts/integral_w_error.py
```diff
@@ -147,9 +149,12 @@ def active_propagation(
...
     s_forerror.run(np.real).run(lambda x: abs(x) ** 2).mean_all_but(
-        [direct, indirect]
-    ).mean(direct)
+        [direct, indirect]).mean(direct)
     s_forerror *= df ** 2
     s_forerror *= N
-    return s_forerror.run(psp.sqrt)
+    s_full = select_pathway(full_s,signal_path)
+    retval = s_full.integrate(direct).set_error(psp.sqrt(s_forerror.data))
+    return retval
```
jmfranck, line 159: what is the point of this change? when we talk, you'll have  to explain it to me.

examples/validate_CT_errors.py
```diff
@@ -0,0 +1,165 @@
+""" Validate Inactive CT Error
+==============================
+
+Estimates the error of the integral of an actual data set of a standard echo
+experiment. Three methods of acquiring the error associated with the data are 
+compared:
```
jmfranck, line 6: when we meet, I'll also diff this against the real data one to make sure we are using the exact same code.

pyspecProcScripts/integral_w_error.py
guinneak198: added in the "/2"!

jmfranck:
way too big and also old

PR #78, opened 2023-01-02
deal with multiple calibrations

jmfranck:


pyspecProcScripts/first_level/QESR_rescale.py
```diff
@@ -1,24 +1,34 @@
...
-if dint_conversion is not None:
-    dint_conversion = float(dint_conversion)
+class calib_info (object):
+    "class for caching calibration info"
+    def __init__():
+        self.current_calib_name = calibration_name
+    def use_calibration(calibration_name):
+        if calibration_name != self.current_calib_name:
+            assert type(calibration_name) is str, "the calibration name must be specified as a string!"
+            assert len(calibration_name)>0, "you MUST use a calibration name and set the values `[calibration_name] q` and `[calibration_name] conversion` in your _pyspecdata file!"
```
jmfranck, line 11: Note that this *deliberately* no longer uses the `default q` and `QESR conversion` parameters in pyspecdata.  From now on, all calibrations must be explicitly named.

So, for example, you'll need to add a parameter in `_pyspecdata` called something like `oldcalib q` that has the same value as `default q` does now and also `oldcalib conversion` to match `qesr conversion` (I would just leave the old parameters there, too, so that if you run old code, it still runs)

pyspecProcScripts/first_level/QESR_rescale.py
```diff
@@ -1,24 +1,34 @@
...
+            self.default_Q = float(pyspec_config.get_setting(f"{calibration_name} Q"))
+            # note that this will fail if you don't have "QESR Conversion" set in your pyspecdata file
+            self.dint_conversion = float(pyspec_config.get_setting(f"{calibration_name} Conversion"))
+            if dint_conversion is not None:
+                dint_conversion = float(dint_conversion)
+calibcache = calib_info()
```
jmfranck, line 17: the point of this class construct is just so that we don't have to go back and re-read the settings file every time we ask for the calibration file (that's what I mean by "caching")

pyspecProcScripts/first_level/QESR_rescale.py
```diff
@@ -1,31 +1,48 @@
...
+    d = Q_(calibcache.d, "mm")  # hard set Q value
     n_B = Q_(1, "dimensionless")  # calculate this
     S = Q_(0.5, "dimensionless")
     c = Q_(
         1, "dimensionless"
     )  # the first fraction on pg 2-17 -- essentially the conversion factor
-    signal_denom = G_R * C_t * sqrt(power) * B_m * n_B * S * (S + 1) * Q
-    signal_denom = signal_denom.to(Q_("G") * sqrt(Q_("W")) * Q_("s"))
+    signal_denom = G_R * C_t * sqrt(power) * B_m * n_B * S * (S + 1) * Q * d**2
+    signal_denom = signal_denom.to(Q_("G") * sqrt(Q_("W")) * Q_("s") * Q_("m")**2)
```
jmfranck, line 45: as I noted [here](https://www.wrike.com/open.htm?id=1031155470) the actual values of the conversion factor will need to be multiplied by d^2 (with d in mm) to compensate for the fact that we are now including d here.  We should only make this change for the new-style (named) conversion factors, so that we can still use the old code.

PR #83, opened 2023-01-25
example that shows calculating variance using mask and np apply along…

guinneak198:
uses apply_along_axis to find the variance and mean along a 1D axis in order to calculate the variance of masked data


examples/check_unitary_std.py
```diff
@@ -4,9 +4,16 @@
...
+domain.
+We propose that if we use a unitary transform,
+Parseval's theorem
+tells us we can calculate :math:`\sigma_t^2` (time domain variance)
+directly from the frequency-domain variance, *i.e.* :math:`\sigma_\nu^2=\sigma_t^2`.
+To confirm this, we construct a "spectrum" of pure noise
+and generate a frequency-masked noise,
+and show that is the same as the 
+unmasked time-domain.
+"""
```
jmfranck: I just reworded slightly

examples/check_unitary_std.py
```diff
@@ -14,22 +14,41 @@
...
+    assert axis is not None
+    def masked_var(x):
+        "this only works for 1D data"
+        if var_has_imag: # take average of variance along real and image
+            return np.var(x[np.isfinite(x)], ddof=1)/2
+        else:
+            return np.var(x[np.isfinite(x)], ddof=1)
+    return np.apply_along_axis(masked_var,axis,x)
+# }}}
+#{{ {generate data with just noise with a phase cycling dimension and repeats dimension
```
jmfranck: let's not spin up new files until we need to.  If we're not using these functions in more than one modules, then we don't need to.

examples/check_unitary_std.py
```diff
@@ -38,16 +57,18 @@
 print("These values are NOT the same so we need a unitary FT for this to work")
 example_data.ft('t', shift=True, unitary=True)
 example_data.ft('ph', unitary=True)
-print("When we apply a unitary FT the std of the frequency is:",
-        sqrt(example_data.C.run(np.var,'t').mean('ph')/2))
+freq_dom_std = sqrt(example_data.C.run(np.var,'t').mean('ph')/2)
+print("When we apply a unitary FT the std over all the frequency domain is:",
+        freq_dom_std)
+print("Because we have no signal, this again corresponds to our noise.")
```
jmfranck: I'm converting some commentary to print statements -- but also, watch the fold markers.  The comment at the top of the fold marker should describe **everything** inside the fold marker.

pyspecProcScripts/__init__.py
```diff
@@ -44,7 +43,5 @@
         'zeroth_order_ph',
         'apod_matched_filter',
         'L2G',
-        'masked_mean_multi',
-        'masked_var_multi'
         ]
```
jmfranck, line 46: let's not spin up new files until we need to. If we're not using these functions in more than one modules, then we don't need to.

jmfranck (commented):
let me know if this looks good -- if it does, go ahead and do `git merge --no-ff JF_masked_var_fn` in your branch, so that it's also available in your branch

PR #86, opened 2023-05-03
Voigt fit ex

guinneak198:
fit ESR spectra in u domain using a voigt

### DIFF
--- a/examples/fit_ESR_voigt.py
+++ b/examples/fit_ESR_voigt.py
@@ -1,0 +1,119 @@
+from matplotlib.pyplot import axvline, axhline, gca
+from pint import UnitRegistry
+from pyspecdata import *
+from itertools import cycle
+from pyspecProcScripts import *
+from pyspecProcScripts import QESR_scalefactor
+from sympy import symbols, Symbol, latex
+from sympy import exp as s_exp
+import numpy as np
+colors = plt.rcParams["axes.prop_cycle"]() # this is the default matplotlib cycler for line styles
+fieldaxis = '$B_0$'
+exp_type = "francklab_esr/alex"
+all_conc = []
+with figlist_var() as fl:
+    for filenum, (thisfile, calibration, diameter, background) in enumerate(
+            [
+              ('230418_200uM_TEMPOL_3_redo2.DSC', '230202', 'QESR caps',
+                  find_file('230418_water.DSC',exp_type=exp_type)['harmonic',0]),
+             ]):
+        #{{{ load data in and rescale and center about 0
+        d = find_file(thisfile,exp_type = exp_type)
+        d -= d[fieldaxis, -100:].data.mean()
+        if "harmonic" in d.dimlabels:
+            d = d['harmonic',0]
+        color = d.get_plot_color()
+        d /= QESR_scalefactor(d, calibration_name = calibration,
+                diameter_name = diameter)
+        center_field = d.getaxis(fieldaxis)[r_[0,-1]].mean()
+        d -= d[fieldaxis, -100:].data.mean()
+        d.setaxis(fieldaxis, lambda x: x-center_field)
+        background = ndshape(d).alloc()
+        background.copy_axes(d)
+        background = background.interp(fieldaxis, d.getaxis(fieldaxis))
+        d -= background
+        #}}}
+        d.ift(fieldaxis, shift=True)
+        d *= 2
+        d[fieldaxis:0] *= 0.5
+        d.ft(fieldaxis)
+        #{{{ Plot starting spectra
+        d.set_units(fieldaxis,'G')
+        fl.next("Starting spectrum")
+        fl.plot(d, label = 'actual data')
+        centers = [-21,-5,13] #make a list of the centers of the starting data for the fits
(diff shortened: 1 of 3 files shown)


examples/fit_ESR_voigt.py
```diff
@@ -0,0 +1,137 @@
...
+with figlist_var() as fl:
+    for filenum, (thisfile, calibration, diameter, background) in enumerate(
+        [
+            (
+                "230418_200uM_TEMPOL_3_redo2.DSC",
+                "230202",
+                "QESR caps",
+                find_file("230418_water.DSC", exp_type=exp_type)["harmonic", 0],
+            ),
+        ]
```
jmfranck, line 21: why not just do for multiple spectra?

examples/fit_ESR_voigt.py
```diff
@@ -0,0 +1,137 @@
...
+        if "harmonic" in d.dimlabels:
+            d = d["harmonic", 0]
+        color = d.get_plot_color()
+        d /= QESR_scalefactor(d, calibration_name=calibration, diameter_name=diameter)
+        center_field = d.getaxis(fieldaxis)[r_[0, -1]].mean()
+        d -= d[fieldaxis, -100:].data.mean()
+        d.setaxis(fieldaxis, lambda x: x - center_field)
+        background = ndshape(d).alloc()
+        background.copy_axes(d)
+        background = background.interp(fieldaxis, d.getaxis(fieldaxis))
```
jmfranck, line 34: what is this? what you are doing here doesn't seem to be doing anything.

examples/fit_ESR_voigt.py
```diff
@@ -0,0 +1,137 @@
...
+    ):
+        # {{{ load data in and rescale and center about 0
+        d = find_file(thisfile, exp_type=exp_type)
+        if "harmonic" in d.dimlabels:
+            d = d["harmonic", 0]
+        color = d.get_plot_color()
+        d /= QESR_scalefactor(d, calibration_name=calibration, diameter_name=diameter)
+        center_field = d.getaxis(fieldaxis)[r_[0, -1]].mean()
+        d -= d[fieldaxis, -100:].data.mean()
+        d.setaxis(fieldaxis, lambda x: x - center_field)
```
jmfranck, line 31: why do this? why not just leave the field axis alone?

examples/fit_ESR_voigt.py
```diff
@@ -0,0 +1,137 @@
...
+        background = ndshape(d).alloc()
+        background.copy_axes(d)
+        background = background.interp(fieldaxis, d.getaxis(fieldaxis))
+        d -= background
+        # }}}
+        d.ift(fieldaxis, shift=True)
+        d *= 2
+        d[fieldaxis:0] *= 0.5
```
jmfranck, line 39: why do this? this is only relevant when you are converting from symmetric to causal signal.

examples/fit_ESR_voigt.py
```diff
@@ -0,0 +1,137 @@
...
+        # }}}
+        # {{{ make symbols and prep lists for symbols
+        list_of_lambda_ls = [
+            symbols("lambda_l%d" % (j + 1), real=True) for j in range(3)
+        ]
+        list_of_lambda_gs = [
+            symbols("lambda_g%d" % (j + 1), real=True) for j in range(3)
+        ]
+        list_of_amps = [symbols("amp%d" % (j + 1), real=True) for j in range(3)]
+        list_of_Bc = [symbols("B%d" % (j + 1), real=True) for j in range(3)]
```
jmfranck, line 73: each of these should be nested in a dictionary comprehension, something like this:
```
all_symb = {symbname:[symbols("%s%d" % (symbname,(j+1)), real=True) for j in range(3)] for symbname in ["lambda_l", "lambda_g", "amp", "B"]}
```

examples/fit_ESR_voigt.py
```diff
@@ -0,0 +1,137 @@
...
+        # {{{ index and make the functions using the symbols - we will have 3 lorentzians, 3 gaussians and 3 shifts
+        Lorentz = []  # the list needs place holders for my loop to work
+        Gauss = []
+        Shifts = []
+        for j in range(3):
+            Lorentz.append(s_exp(-pi * list_of_lambda_ls[j] * abs(u)))
+            Gauss.append(
+                s_exp(-(pi**2 * list_of_lambda_gs[j] ** 2 * u**2) / (4 * np.log(2)))
+            )
+            Shifts.append(s_exp(+1j * pi * 2 * u * list_of_Bc[j]))
```
jmfranck, line 85: so, this gets modified to use the dict above.

examples/fit_ESR_voigt.py
```diff
@@ -0,0 +1,137 @@
...
+            for j in range(3)
+        }
+        f.set_guess(
+            **all_lambda_l,
+            **all_lambda_g,
+            **all_shift,
+            **all_amp,
+        )
+        f.settoguess()
+        guess = f.eval()
```
jmfranck, line 118: this is part of plotting the guess (just move under comment below)

jmfranck (changes requested):

PR #87, opened 2023-10-06
Fitting E(p)

guinneak198:
https://jmfrancklab.slack.com/archives/CTD29CZPE/p1689104438985819


examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,105 @@
+"""Fitting ODNP Datasets for Ksigma
+===================================
+The T1(p) and E(p) integrals are pulled from an H5 file and are fit to extract the cross relaxivity of the sample. The R1(p) is first interpolated by fitting the self relaxivity to a polynomial. This fit is then used to fit the linear regime of the enhancement data. The cross relaxivity is then calculated using the output of these fits as well as the sample parameters that are fed to it (e.g., ppt value, and concentration).
```
jmfranck, line 3: What does "fit the linear regime" mean? Only the linear regime?

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,105 @@
...
+"""
+from pyspecdata import *
+from sympy import symbols, Symbol, latex
+
+target_directory = os.path.normpath(getDATADIR("AG_processed_data"))
+h5_folder = "ras.h5"
```
jmfranck, line 9: This is the file name. The other is the node, no?

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,105 @@
...
+filename = "220616_E37"
+# {{{Dataset Parameters
+SL_conc_M = 455.2223e-6
+ppt = 1.5167e-3
```
jmfranck, line 13: Why don't these come from a file?

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,105 @@
...
+    list_symbs = "\n".join(list_symbs)
+    # }}}
+    return list_symbs
+
+
+# }}}
+with figlist_var() as fl:
+    # {{{ load data
+    Ep = nddata_hdf5(f"{h5_folder}/{filename}/Ep", directory=target_directory)
+    T1p = nddata_hdf5(f"{h5_folder}/{filename}/T1p", directory=target_directory)
```
jmfranck, line 32: Why not stored as R1?

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,105 @@
...
+    plt.ylabel("E(p)")
+    # }}}
+    # {{{ plot T1p data
+    R1p = 1 / T1p
+    fl.next(r"$R_{1}(p)$")
+    fl.plot(R1p, "o", label="Experimental Data")
+    # }}}
+    # {{{Fit R1p
+    T100 = 3.71
+    dT10 = 0.2
```
jmfranck, line 47: Why not pulled from a file?

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,105 @@
...
+    a, b, c, power = symbols("a b c power", real=True)  # symbols
```
jmfranck, line 48: Why not just called p?

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,105 @@
...
+    f.functional_form = (1 / (T100 + dT10 * power)) + (
+        SL_conc_M / (a + b * power + c * power**2)
+    )  # declare fitting function
+    f.set_guess(
+        a=dict(value=1.12e-4, min=1e-5, max=5e-3),
+        b=dict(value=7.5e-5, min=-0.001, max=0.001),
+        c=dict(value=-1.5e-5, min=-0.001, max=0.001),
+    )
+    f.settoguess()
+    f.fit()
```
jmfranck, line 59: This is overly complicated. The point is to fit to a polynomial

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,105 @@
...
+    R1p_fit = f.eval(Ep_pts)
+    fl.plot(R1p_fit, ls=":", color="k", label="Fit", alpha=0.5)
+    plt.ylabel(r"$R_{1} / s^{-1}$")
+    plt.xlabel("Power / W")
+    # }}}
+    # {{{ Fit E(p)
+    fl.next("E(p)")
+    M0, A, phalf, power = symbols("M0 A phalf power", real=True)
```
jmfranck, line 67: If you call it p_{half} on the right, does it look nicer?

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,105 @@
...
+    sp = power / (power + phalf)
+    R1p = (1 / (T100 + dT10 * power)) + (
+        SL_conc_M / (f.output("a") + f.output("b") * power + f.output("c") * power**2)
+    )
+    Ep_fit = lmfitdata(Ep["power", :-3])
+    Ep_fit.functional_form = M0 - ((M0 * A * sp) / R1p)
+    Ep_fit.set_guess(
+        M0=dict(value=8e4, min=6.4e4, max=11e4),
```
jmfranck, line 75: Guesses should be based off of the data

jmfranck (changes requested):

examples/fit_ODNP_data.py
guinneak198: good catch

examples/fit_ODNP_data.py
guinneak198: These are saved in the acquired dataset.h5 file. Here we are pulling the table of integrals h5 file. I can load the acquired dataset to pull the acquisition parameters if you think that'd be better. I think that would overcomplicate but you know best

examples/fit_ODNP_data.py
guinneak198: This question should be for the script that makes the table of integrals. At this point in processing this is what we have.

examples/fit_ODNP_data.py
guinneak198: the last T100 dataset we have is from 230412, added the mat file to this repo so we can pull from it to get T100 and dT10

examples/fit_ODNP_data.py
guinneak198: "power" is a little more clear to the user that that's the axis we are fitting, p doesn't give much information but changed the code anyway

examples/fit_ODNP_data.py
guinneak198: `"'p_{half}' is not a valid Parameters name"`

examples/fit_ODNP_data.py
guinneak198: my attempt at emphasizing that the initial linear regime should have a decent number of points... suggestions for rewording?

examples/fit_ODNP_data.py
guinneak198: edited

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,99 @@
...
+    # }}}
+    # {{{load in T100 dataset
+    T10_p = loadmat('T10_DI_water_230412')['a'][0,:]
+    R10_p = nddata((T10_p[0]+T10_p[1]*R1p.getaxis('power'))**-1,'power')
+    R10_p.setaxis('power',R1p.getaxis('power'))
+    #}}}
+    #{{{ fit krho and R1p
+    krho_inv = SL_conc_M/(R1p - R10_p)
+    krho_inv_fit = krho_inv.polyfit('power',order=2)
+    krho_inv_fine = nddata(R1p.C.getaxis('power'),'power').eval_poly(krho_inv_fit,'power')
```
jmfranck, line 54: This should be a simple fromaxis command

jmfranck (changes requested):

examples/fit_ODNP_data.py
guinneak198: much better solution thank you

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,99 @@
+"""Fitting ODNP Datasets for Ksigma
+===================================
+The T1(p) and E(p) integrals are pulled from an H5 file and are fit to extract the cross relaxivity of the sample. The R1(p) is first interpolated by fitting the self relaxivity to a polynomial. This fit is then used as a parameter for fitting the linear regime of the enhancement data that is extrapolated to the highest power. The cross relaxivity is then calculated using the output of these fits as well as the sample parameters that are fed to it (e.g., ppt value, and concentration).
+"""
+from pyspecdata import *
+from sympy import symbols, Symbol, latex
+from scipy.io import loadmat
+
+target_directory = os.path.normpath(getDATADIR("AG_processed_data"))
+h5_folder = "ras.h5"
```
jmfranck, line 10: this is not a folder.  it is a file

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,99 @@
...
+nodename = "220616_E37"
+# {{{Dataset Parameters
+SL_conc_M = 455.2223e-6
+ppt = 1.5167e-3
+Ep_pts = 18
```
jmfranck, line 15: this is a processing file -- why does this need to be set.

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,99 @@
...
+# {{{ plotting fn
+def list_symbs(f):
+    # {{{ this is just to show all the parameters
+    list_symbs = []
+    for j, k in f.output().items():
+        s_repr = latex(Symbol(j))
+        list_symbs.append(f"${s_repr} = {k:0.5g}$")
+    list_symbs = "\n".join(list_symbs)
+    # }}}
+    return list_symbs
```
jmfranck, line 26: this could be written much more simply with a list comprehension

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,99 @@
...
+    # }}}
+    # {{{ plot T1p data
+    R1p = 1 / T1p
+    fl.next(r"$R_{1}(p)$")
+    fl.plot(R1p, "o", label="Experimental Data")
+    # }}}
+    # {{{load in T100 dataset
+    T10_p = loadmat('T10_DI_water_230412')['a'][0,:]
+    R10_p = nddata((T10_p[0]+T10_p[1]*R1p.getaxis('power'))**-1,'power')
+    R10_p.setaxis('power',R1p.getaxis('power'))
```
jmfranck, line 49: use fromaxis

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,99 @@
...
+    plt.ylabel(r"$R_{1} / s^{-1}$")
+    plt.xlabel("Power / W")
+    # }}}
+    # {{{ Fit E(p)
+    fl.next("E(p)")
+    M0, A, phalf, p = symbols("M0 A phalf power", real=True)
+    sp = p / (p + phalf)
+    R1p = ((T10_p[0]+T10_p[1]*p)**-1 + (
+        SL_conc_M / (krho_inv_fit[0] + krho_inv_fit[1] * p + krho_inv_fit[2] * p**2)))
+    Ep_fit = lmfitdata(Ep["power", :-3])
```
jmfranck, line 66: see above -- 3 not hard-coded

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,99 @@
...
+    fl.plot(R1p_fit, ls=":", color="k", label="Fit", alpha=0.5)
+    plt.ylabel(r"$R_{1} / s^{-1}$")
+    plt.xlabel("Power / W")
+    # }}}
+    # {{{ Fit E(p)
+    fl.next("E(p)")
+    M0, A, phalf, p = symbols("M0 A phalf power", real=True)
+    sp = p / (p + phalf)
+    R1p = ((T10_p[0]+T10_p[1]*p)**-1 + (
+        SL_conc_M / (krho_inv_fit[0] + krho_inv_fit[1] * p + krho_inv_fit[2] * p**2)))
```
jmfranck, line 65: why can't this just be `R1p_fit` -- alternately, why can't this be shown on the R1(p) plot?

jmfranck (changes requested):

examples/fit_ODNP_data.py
guinneak198: this is processing the table of integrals for a dataset - Not the raw acquired dataset. The only thing saved to the h5 file here is the table of integrals for Ep and T1

examples/fit_ODNP_data.py
guinneak198: `Traceback (most recent call last):
  File "fit_ODNP_data.py", line 48, in <module>
    R10_p = nddata((T10_p[0]+T10_p[1]*R1p.fromaxis('power'))**-1,'power')
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\core.py", line 2170, in __init__
    if len(args[0].shape) == 1 and isinstance(args[1], str):
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\core.py", line 5666, in __getattribute__
    return super().__getattribute__(arg)
AttributeError: 'nddata' object has no attribute 'shape'
`

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,99 @@
...
+# }}}
+with figlist_var() as fl:
+    # {{{ load data
+    Ep = nddata_hdf5(f"{h5_folder}/{nodename}/Ep", directory=target_directory)
+    T1p = nddata_hdf5(f"{h5_folder}/{nodename}/T1p", directory=target_directory)
+    # }}}
+    # {{{Plot Ep
+    fl.next("E(p)")
+    fl.plot(Ep["power", :-3], "o", label="Experimental Data", capsize=6, alpha=0.5)
+    fl.plot(Ep["power", -3:], "rx", label="Returning Power Check", capsize=6, alpha=0.5)
```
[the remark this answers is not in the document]
guinneak198: okay I made it more versatile by using diff and it does the job but might not be what you originally were thinking

examples/fit_ODNP_data.py
guinneak198: the sympy expression (which R1p is used in) must be symbolic 
`AssertionError: for now, the functional form must be a sympy expression`

guinneak198: But edited the R1(p) fit plot to use this expression instead

examples/fit_ODNP_data.py
jmfranck: Isn't the easier solution to use set_prop on the table of integrals and transfer them?

examples/fit_ODNP_data.py
jmfranck: Why can't you use ndshape to figure out how many points there are?

jmfranck (changes requested):
Already found a couple issues that weren't addressed... Please look back at them. Comments, address them, and read through the code line by line in a very critical way.

examples/fit_ODNP_data.py
guinneak198: ndshape returns `[('power',19)]` so instead I set it to `len(Ep.getaxis('power'))`

examples/fit_ODNP_data.py
guinneak198: rephrased

examples/fit_ODNP_data.py
guinneak198: fixed

examples/fit_ODNP_data.py
guinneak198: changed the processing script that generates this particular dataset to save R1p instead but note this will need to be done for all future post processing prior to running this script

examples/fit_ODNP_data.py
guinneak198: The other parameters are wide enough to accompany the majority of cases

examples/fit_ODNP_data.py
jmfranck: You if the inverse of the self relaxivity

examples/fit_ODNP_data.py
jmfranck: ???

examples/fit_ODNP_data.py
jmfranck: List comprehension

examples/fit_ODNP_data.py
jmfranck: You should write a short script that can adapt old format for this and the other issue you noted (I noted you could write a short script using h5py

examples/fit_ODNP_data.py
jmfranck: Didn’t

examples/fit_ODNP_data.py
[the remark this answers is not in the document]
jmfranck: No use diff

examples/fit_ODNP_data.py
jmfranck: Do not hard code use diff

jmfranck (changes requested):

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,119 @@
...
+    # }}}
+    # {{{ fit krho inverse with two degrees of freedom and then apply to fit R1p
+    krho_inv = SL_conc_M / (R1p - R10_p)
+    krho_inv_fit = krho_inv.polyfit("power", order=2)
+    krho_inv_fine = R1p.fromaxis("power").eval_poly(krho_inv_fit, "power")
+    R1p_fit = (T10_p[0] + T10_p[1] * R1p.fromaxis("power")) ** -1 + (
+        SL_conc_M
+        / (
+            krho_inv_fit[0]
+            + krho_inv_fit[1] * R1p.fromaxis("power")
```
jmfranck, line 72: Why not just evaluate-polynomial here?

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,119 @@
...
+
+# }}}
+# {{{ load data
+Ep = nddata_hdf5(f"{h5_file}/{nodename}/Ep", directory=target_directory)
+R1p = nddata_hdf5(f"{h5_file}/{nodename}/R1p", directory=target_directory)
+# }}}
+# {{{Dataset Parameters
+SL_conc_M = Ep.get_prop("acq_params")["concentration"]
+ppt = Ep.get_prop("acq_params")["guessed_MHz_to_GHz"] * 1e-3
+Ep_pts = Ep.getaxis("power")
```
jmfranck, line 27: Why do you do this? You should not need this.

jmfranck: (Just from axis in the place it is needed)

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,119 @@
...
+        alpha=0.5,
+    )
+    # }}}
+    # {{{ plot R1p data
+    fl.next(r"$R_{1}(p)$")
+    fl.plot(R1p, "o", label="Experimental Data")
+    # }}}
+    # {{{load in T100 dataset
+    T10_p = loadmat("T10_DI_water_230412")["a"][0, :]
+    R10_p = nddata((T10_p[0] + T10_p[1] * R1p.getaxis("power")) ** -1, "power")
```
jmfranck, line 61: Use 'fromaxis'

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,119 @@
...
+    R10_p.setaxis("power", R1p.getaxis("power"))
```
jmfranck, line 62: See prev!

examples/fit_ODNP_data.py
jmfranck: You are calculate the same thing twice

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,119 @@
...
+    fl.next("Integrated Enhancement")
+    M0, A, phalf, p = symbols("M0 A phalf power", real=True)
+    sp = p / (p + phalf)
+    R1p = (T10_p[0] + T10_p[1] * p) ** -1 + (
+        SL_conc_M / (krho_inv_fit[0] + krho_inv_fit[1] * p + krho_inv_fit[2] * p**2)
+    )  # Symbolic expression for R1p that is used in the symbolic function for the fitting of E(p)
+    Ep_fit = lmfitdata(Ep["power", :flip_idx])
+    Ep_fit.functional_form = M0 - ((M0 * A * sp) / R1p)
+    Ep_fit.set_guess(
+        M0=dict(value=Ep["power", 0].real.data, min=1e4, max=11e4),
```
jmfranck, line 90: Use `.item()`

examples/fit_ODNP_data.py
guinneak198: `Traceback (most recent call last):
  File "fit_ODNP_data.py", line 61, in <module>
    R10_p = nddata((T10_p[0] + T10_p[1] * R1p.fromaxis("power")) ** -1, "power")
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\core.py", line 2170, in __init__
    if len(args[0].shape) == 1 and isinstance(args[1], str):
  File "c:\users\alex guinness\gitstuff\pyspecdata\pyspecdata\core.py", line 5666, in __getattribute__
    return super().__getattribute__(arg)
`

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+# }}}
+# {{{ load data
+Ep = nddata_hdf5(f"{h5_file}/{nodename}/Ep", directory=target_directory)
+R1p = nddata_hdf5(f"{h5_file}/{nodename}/R1p", directory=target_directory)
+# }}}
+# {{{Find the index where the return powers begin
+flip_idx = np.where(Ep.fromaxis("power").diff("power").data < 0)[0][0]
+flip_idx += 1
```
jmfranck, line 25: good!  I would just move the +1 up and make one line

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,108 @@
...
+    plt.xlabel("Power / W")
+    # }}}
+    # {{{ Fit E(p)
+    fl.next("Integrated Enhancement")
+    M0, A, phalf, p = symbols("M0 A phalf power", real=True)
+    sp = p / (p + phalf)
+    R1p = (T10_p[0] + T10_p[1] * p) ** -1 + (
+        Ep.get_prop("acq_params")["concentration"]
+        / (krho_inv_fit[0] + krho_inv_fit[1] * p + krho_inv_fit[2] * p**2)
+    )  # Symbolic expression for R1p that is used in the symbolic function for the fitting of E(p)
```
jmfranck, line 73: This is repeated from above, but I think the difference is that this is calculated on a different power axis.

The best way to resolve this is to make one function that takes arguments (1) an nddata power axis, (2) polynomial coeffs for kρ^-1 and (3) T10 and that outputs R1 as a function of power.

examples/fit_ODNP_data.py
jmfranck: you still need to write this script

jmfranck: the other issue that I note is discussed [here](https://jmfrancklab.slack.com/archives/CLMMYDD98/p1697637883567909?thread_ts=1697637620.788759&cid=CLMMYDD98)

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,108 @@
...
+    Ep_fit = lmfitdata(Ep["power", :flip_idx])
+    Ep_fit.functional_form = M0 - ((M0 * A * sp) / R1p)
+    Ep_fit.set_guess(
+        M0=dict(value=Ep["power", 0].real.item(), min=1e4, max=11e4),
+        A=dict(value=13, min=0.5, max=17),
```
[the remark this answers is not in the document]
guinneak198: flip_idx is 17 R1p only has 5 indices at most... but I think I see what you are getting at

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,102 @@
...
+    # {{{ Fit E(p)
+    fl.next("Integrated Enhancement")
+    M0, A, phalf, p = symbols("M0 A phalf power", real=True)
+    sp = p / (p + phalf)
+    Ep_fit = lmfitdata(Ep["power", :flip_idx])
+    # Symbolic expression for Ep that is used in the symbolic function for the fitting of E(p)
+    Ep_fit.functional_form = M0 - ((M0 * A * sp) / (T10_p[0] + T10_p[1] * p) ** -1 + (
+        Ep.get_prop("acq_params")["concentration"]
+        / (krho_inv_fit[0] + krho_inv_fit[1] * p)))
+    A_guess = 1 - (Ep["power", flip_idx].data / Ep["power", 0].data).real
```
jmfranck, line 66: I think you also want to include relaxation as well

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,102 @@
...
+            "T10_DI_water_230412", exp_type="AG_processed_data", unique=True
+        )
+    )["a"][0, :]
+    R10_p = R1p.fromaxis("power").eval_poly(T10_p, "power") ** -1
+    # }}}
+    # {{{ fit krho inverse with two degrees of freedom and then apply to fit R1p
+    krho_inv = Ep.get_prop("acq_params")["concentration"] / (R1p - R10_p)
+    krho_inv_fit = krho_inv.polyfit("power", order=1)
+    krho_inv_fine = R1p.fromaxis("power").eval_poly(krho_inv_fit, "power")
+    R1p_fit = R10_p + Ep.get_prop("acq_params")["concentration"] / krho_inv_fine
```
jmfranck, line 52: Remember you need to add a script that can add concentration to old data (see comments in main conversation of PR)

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,102 @@
...
+    T10_p = loadmat(
+        search_filename(
+            "T10_DI_water_230412", exp_type="AG_processed_data", unique=True
+        )
+    )["a"][0, :]
+    R10_p = R1p.fromaxis("power").eval_poly(T10_p, "power") ** -1
+    # }}}
+    # {{{ fit krho inverse with two degrees of freedom and then apply to fit R1p
+    krho_inv = Ep.get_prop("acq_params")["concentration"] / (R1p - R10_p)
+    krho_inv_fit = krho_inv.polyfit("power", order=1)
```
jmfranck, line 50: Calling this coeff rather than fit would help with the code legibility

jmfranck (changes requested):

examples/fit_ODNP_data.py
guinneak198: included in the `(T10_p[0] +T10_p[1]*p)**-1`

jmfranck: I'm talking about the guess for A

jmfranck (commented):
Replied to your comment

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,102 @@
...
+    plt.ylabel(r"$R_{1} / s^{-1}$")
+    plt.xlabel("Power / W")
+    # }}}
+    # {{{ Fit E(p)
+    fl.next("Integrated Enhancement")
+    M0, A, phalf, p = symbols("M0 A phalf power", real=True)
+    sp = p / (p + phalf)
+    Ep_fit = lmfitdata(Ep["power", :flip_idx])
+    # Symbolic expression for Ep that is used in the symbolic function for the fitting of E(p)
+    Ep_fit.functional_form = M0 - ((M0 * A * sp) / (T10_p[0] + T10_p[1] * p) ** -1 + (
```
jmfranck, line 63: There's a problem with your parentheses here, so that you don't end up doing the correct equation. You want to add together their relaxation rates, and divide your second term here by the relaxation rates (because it's supposed to be the term that involves the relaxivity times the T1)

jmfranck (changes requested):

examples/fit_ODNP_data.py
jmfranck: I would actually break this into two lines.... Construct the symbolic expression for the relaxation on the previous line, and then include it here.

jmfranck (changes requested):
We discussed both how to implement the h5py script using r+, and how to avoid repeating calculations of the power dependence of R1 by defining a single symbolic expression that is used both to generate the plot of R1 versus power as well as to form the denominator in the second term of your model for enhancements.

jmfranck:
Remember to set up one symbolic expression for R1(p) as we discussed in person

examples/append_conc_to_h5file.py
```diff
@@ -0,0 +1,11 @@
+from pyspecdata import *
+import h5py
+filename = '220126_Ras_M67R1a_capProbe.h5'
+actual_conc = 72e-6 #M
+h5 = search_filename(f"{filename}" , exp_type = 'ODNP_NMR_comp/ODNP')
```
jmfranck, line 5: unique=True

examples/append_conc_to_h5file.py
```diff
@@ -0,0 +1,11 @@
...
+h5 = ['/'.join(path.split('\\')) for path in h5]
```
jmfranck, line 6: what is this doing?

examples/append_conc_to_h5file.py
```diff
@@ -0,0 +1,11 @@
+from pyspecdata import *
```
jmfranck, line 1: docstring required

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,125 @@
+"""Fitting ODNP Datasets for Ksigma
+===================================
+The T1(p) and E(p) integrals are stored in previous post processing pulled from an H5 file. The inverse of the T1(p) data is taken to produce the R1(p) data which is fit using the polyfit of the inverse of krho with two degrees of freedom. The R1(p) fit is then fed into the fitting routine for the enhancement data prior to normalization. The cross relaxivity is then calculated using the output of these fits as well as the sample parameters that are fed to it (e.g., ppt value, and concentration).
```
jmfranck, line 3: What example generates these?

jmfranck: also, this line is way too long -- you need to wrap it so that it can be diffed

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,125 @@
...
+    p,
+):  # symbolic expression for R1p used in both fitting the R1p data as well as the symbolic expression for E(p)
+    expression = (T10_p[0] + T10_p[1] * p) ** -1 + (
+        Ep.get_prop("acq_params")["concentration"]
+        / (krho_inv_coeff[0] + krho_inv_coeff[1] * p)
+    )
+    return expression
+
+
+# }}}
```
jmfranck, line 24: you should not have a function to define the expression.   You should just define the expression inline

examples/fit_ODNP_data.py
guinneak198: currently generate_integrals.py generates the integrals however, this function returns the normalized Ep and so I can either adjust that here in this PR and update the function to also return the R1p rather than the T1p integrals or make that a separate PR-your call

examples/append_conc_to_h5file.py
guinneak198: h5.File requires / NOT \ so I need to convert

guinneak198: fixed using normpath

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,126 @@
...
+    plt.ylabel(r"$R_{1} / s^{-1}$")
+    plt.xlabel("Power / W")
+    # }}}
+    # {{{ Fit E(p)
+    fl.next("Integrated Enhancement")
+    sp = p / (p + phalf)
+    Ep_fit = lmfitdata(Ep["power", :flip_idx])
+    # Symbolic expression for Ep that is used in the symbolic function for the fitting of E(p)
+    Ep_fit.functional_form = M0 - ((M0 * A * sp) / R1p_expression)
+    A_guess = (
```
jmfranck, line 85: explain what you're doing in a comment

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,126 @@
...
+    # }}}
+    # {{{ fit krho inverse with two degrees of freedom and then apply to fit R1p
+    krho_inv = Ep.get_prop("acq_params")["concentration"] / (R1p - R10_p)
+    krho_inv_coeff = krho_inv.polyfit("power", order=1)
+    krho_inv_fine = R1p.fromaxis("power").eval_poly(krho_inv_coeff, "power")
+    M0, A, phalf, p = symbols("M0 A phalf power", real=True)
+    R1p_expression = (T10_p[0] + T10_p[1] * p) ** -1 + (
+        Ep.get_prop("acq_params")["concentration"]
+        / (krho_inv_coeff[0] + krho_inv_coeff[1] * p)
+    )
```
jmfranck, line 73: yes!

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,126 @@
...
+    # }}}
+    # {{{load in T100 dataset
+    T10_p = loadmat(
+        search_filename(
+            "T10_DI_water_230412", exp_type="AG_processed_data", unique=True
+        )
+    )["a"][0, :]
+    R10_p = R1p.fromaxis("power").eval_poly(T10_p, "power") ** -1
+    # }}}
+    # {{{ fit krho inverse with two degrees of freedom and then apply to fit R1p
```
jmfranck, line 65: move this up above line 57 (load in T100...) since loading the T10 is part of this

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    # }}}
+    # {{{ plot R1p data
+    fl.next(r"$R_{1}(p)$")
+    fl.plot(R1p, "o", label="Experimental Data")
+    # }}}
+    # {{{load in T100 dataset
+    T10_p = loadmat(
+        search_filename(
+            "T10_DI_water_230412", exp_type="AG_processed_data", unique=True
```
jmfranck, line 60: please remove all references to filenames and to explicit `exp_type` within the code itself -- move up above, where you give the main filename.

I would suggest loading into dictionaries:
```
main_data = dict(
    data_dir='AG_processed_data',
    filename='whatever.h5',
    nodename='whatever',
    )
```
etc. but if you have an equally legible way of doing it, that's good, too.

examples/append_conc_to_h5file.py
```diff
@@ -0,0 +1,20 @@
...
+'concentration' does not exist in the acq_params, this will 
+add the concentration defined by 'actual_conc' into acq_params
+under 'concentration'. If the attribute already exists it will
+simply replace the current concentration with the value of 
+'actual_conc'.
+"""
+from pyspecdata import *
+import h5py,os
+filename = '220126_Ras_M67R1a_capProbe.h5'
+actual_conc = 72e-6 #M
```
jmfranck, line 14: add a comment block pointing out that this is what you change ... see comment below about loading this as dicts

examples/append_conc_to_h5file.py
```diff
@@ -0,0 +1,20 @@
...
+h5 = search_filename(f"{filename}" , exp_type = 'ODNP_NMR_comp/ODNP',unique=True)
```
jmfranck, line 15: `"f{filename}"` is the same as `filename` --> but this is obsolete (see other comments)

jmfranck (changes requested):

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,126 @@
+"""Fitting ODNP Datasets for Ksigma
+===================================
+The T1(p) and E(p) integrals are stored in
+previous post processing pulled from an H5 file.
```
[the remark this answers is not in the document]
guinneak198: the current example is generate_integrals.py, however a. this saves the normalized E(p) (not what we want here), and b. doesn't save the R1(p) (saves the T1p). I can either change that in this PR or make a followup to address those issues. This current dataset of integrals is generated independently in my notebook

examples/append_conc_to_h5file.py
```diff
@@ -0,0 +1,21 @@
...
+Loads in a previously saved HDF5 dataset. If the attribute
+'concentration' does not exist in the acq_params, this will 
+add the concentration defined by 'actual_conc' into acq_params
+under 'concentration'. If the attribute already exists it will
+simply replace the current concentration with the value of 
+'actual_conc'.
+"""
+from pyspecdata import *
+import h5py,os
+data_dict = {"filename":'220126_Ras_M67R1a_capProbe.h5', #file that is being edited
```
jmfranck, line 13: Call this data file info or something... It's not a dictionary of the data.

While you're at it, you can use the dict( method of building the dictionary which is a little bit more clear

examples/append_conc_to_h5file.py
```diff
@@ -0,0 +1,21 @@
...
+        "file_location":"ODNP_NMR_comp/ODNP", #location of file
+        "actual_conc": 72e-6} #concentration of the dataset in M
```
jmfranck, line 15: Logically, the concentration is a separate thing... This is not information about where the file is located as is true of the dictionary... It is the concentration information you are incorporating into that file.

examples/fit_ODNP_data.py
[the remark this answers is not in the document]
jmfranck: That would be a follow-up, but so that you don't forget about it, you need to talk about that example in this docstring

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,128 @@
...
+as the sample parameters that are fed to in (e.g.,
+ppt value, and concentration).
+"""
+from pyspecdata import *
+from sympy import symbols, Symbol, latex,lambdify
+from scipy.io import loadmat
+
+data_sets = dict({"data_dir":'AG_processed_data', #directory of the dataset of table of integrals
+    "integrals_filename":"ras.h5", #h5 file that contains a series of integrals for different samples
+    "T100_filename": "T10_DI_water_230412", #mat file with polyfit water relaxation
```
jmfranck, line 21: I was specifically suggesting that you have a dictionary that contains the information about the location of each data set... The dictionary would be named differently, but the keys would be the same.  Even though this is a stylistic, it's a clearly more logical way of organizing the code.

jmfranck (changes requested):

examples/append_conc_to_h5file.py
guinneak198: so I changed the name to data_info which I think encapsulates file name, location and the concentration but if you think this is still weird I can adjust easily

examples/fit_ODNP_data.py
guinneak198: that does make more sense okay!

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,135 @@
...
+from scipy.io import loadmat
+
+data_info = dict({
+    "filename":"ras.h5", #h5 file containing table of integrals for different datasets
+    "data_dir":"AG_processed_data", #directory of the dataset of table of integrals
+    "nodename": "230706_M67_a"}) #specific nodename of the dataset of interest
+T100_info = dict({
+    "filename":"T10_DI_water_230412",
+    "data_dir":"AG_processed_data", 
+    "nodename":None})
```
jmfranck, line 29: This key is not needed

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,135 @@
...
+calculated using the output of these fits as well
+as the sample parameters that are fed to in (e.g.,
+ppt value, and concentration).
+"""
+from pyspecdata import *
+from sympy import symbols, Symbol, latex,lambdify
+from scipy.io import loadmat
+
+data_info = dict({
+    "filename":"ras.h5", #h5 file containing table of integrals for different datasets
```
jmfranck, line 23: No curly brackets, then `filename="thename"` etc

jmfranck: You can specify keys and values as keyword arguments to the dict function

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,135 @@
...
+    "data_dir":"AG_processed_data", 
+    "nodename":None})
+
+# {{{ load data
+Ep = find_file(data_info['filename'], exp_type=data_info['data_dir'], expno=f"{data_info['nodename']}/Ep")
+# Some older h5 files save the T1p rather than the R1p. If there isn't an R1p expno then it will load the T1p integrals and convert to R1p by taking the inverse
+try:
+    R1p = find_file(data_info['filename'], exp_type=data_info["data_dir"], expno=f"{data_info['nodename']}/R1p")
+except:
+    T1p = find_file(data_info['filename'], exp_type=data_info['data_dir'], expno=f"{data_info['nodename']}/T1p")
```
jmfranck, line 37: The try-except here is nice!

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,135 @@
...
+    A_guess = (
+        1
+        - (
+            R1p["power", 0].data * (Ep["power", flip_idx].data / Ep["power", 0].data)
+        ).real
+    )
+    Ep_fit.set_guess(
+        M0=dict(value=Ep["power", 0].real.item(), min=1e4, max=11e4),
+        A=dict(value=A_guess, min=0.5 * A_guess, max=3 * A_guess),
+        phalf=dict(value=0.2, min=0.1, max=0.4),
```
jmfranck, line 103: Wider bounds?

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,135 @@
...
+    )
+    Ep_fit.settoguess()
+    Ep_fit.fit()
+    thisfit = Ep_fit.eval(100)
+    fl.plot(thisfit, ls=":", color="k", label="Fit", alpha=0.5)
+    ksig = (
+        Ep_fit.output("A") * Ep.get_prop("acq_params")["guessed_MHz_to_GHz"] * 1e-3
```
jmfranck, line 110: Why are you using the guess of the PPT value here?

guinneak198: there is no "actual ppt" value in the config file. the guessed is used to set the bounds of the field sweep and then we plug in the actual value we find before we run ODNP - but I can edit the append_conc... to be edit_h5_acq_params and make that an option?

examples/append_conc_to_h5file.py
```diff
@@ -9,14 +9,17 @@
 'actual_conc'.
 """
 from pyspecdata import *
-import h5py,os
-data_info = dict({"filename":'220126_Ras_M67R1a_capProbe.h5', #file that is being edited
-        "file_location":"ODNP_NMR_comp/ODNP", #location of file
-        "actual_conc": 72e-6}
-        )#concentration of the dataset in M
```
jmfranck: you should review my changes carefully, but be mindful that some of these come from running black

examples/fit_ODNP_data.py
```diff
@@ -1,42 +1,63 @@
...
+(as of 11/8/23 the `generate_integrals` is not
 updated), and stored in
 previous post processing pulled from an H5 file.
-The inverse of the T1(p) data processing is taken
-to produce the R1(p) data which is fit using the
+The inverse of the :math:`T_1(p)` data processing is taken
+to produce the :math:`R_1(p)` data which is fit using the
 polyfit of the inverse of krho with two degrees of
-freedom. The R1(p) fit is then fed into the
+freedom. The :math:`R_1(p)` fit is then fed into the
```
jmfranck, line 11: sphinx will format this as latex.  If you're using my vimrc, you can get this in python just by typing `$$`

examples/fit_ODNP_data.py
```diff
@@ -61,25 +82,35 @@
...
     R1p_expression = (T10_p[0] + T10_p[1] * p) ** -1 + (
         Ep.get_prop("acq_params")["concentration"]
         / (krho_inv_coeff[0] + krho_inv_coeff[1] * p)
     )
-    R1p_fit = lambdify(p,R1p_expression)
-    fl.plot(R1p_fit(R1p.fromaxis('power')), ls=":", color="k", label="Fit", alpha=0.5)
+    R1p_fit = lambdify(p, R1p_expression)
+    fl.plot(
+        R1p_fit(R1p.fromaxis("power")),
+        ls=":",  # PR COMMENT: again, why are you plotting this as :, since it's only evaluated at the powers of interest -- maybe you want to construct a new axis?  Why not just plot as a solid line, anyways?
```
jmfranck, line 109: I was wasting time jumping between browser and vim, so I did this instead.

All of these need to be dealt with and removed before finalizing, please (you can also just search for PR COMMENT) in the code.

They should all be pretty straightforward.

jmfranck (changes requested):

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,177 @@
...
+    # 1-E(pₘₐₓ)R₁(pₘₐₓ) ≈ A
+    # because R₁(0) is typically larger and also easier to access than
+    # R₁(pₘₐₓ), we just use it instead.
+    A_guess = (
+        1
+        - (
+            R1p["power", 0].real.item()
+            * (
+                Ep["power", flip_idx].real.item()
+            )  # PR COMMENT: you were dividing by 1?
```
guinneak198, line 139: no, the Ep is not normalized! without this dividion A is wayyyyyyyy too high of a guess!

guinneak198: Remember here we are saying 1 - Ep NOT M0-M0*Ep

examples/fit_ODNP_data.py
```diff
@@ -106,7 +103,6 @@
     R1p_fit = lambdify(p, R1p_expression)
     fl.plot(
         R1p_fit(R1p.fromaxis("power")),
```
jmfranck, line 105: this didn't fix the problem that this is not actually a smooth curve

jmfranck: that's the point of the "fine" nddata that I create above.

jmfranck (changes requested):

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,174 @@
...
+                /integral_vs_p["power",0].real.item()
+            )
+        ).real
+    )
+    integral_vs_p_fit.set_guess(
+        M0=dict(value=integral_vs_p["power", 0].real.item(), min=1e4, max=11e4),
+        A=dict(value=A_guess, min=0.2 * A_guess, max=3 * A_guess),
+        phalf=dict(value=0.2, min=0.05, max=1.0),
+    )
+    integral_vs_p_fit.settoguess()
```
jmfranck, line 144: Set to guess is not required if you are about to call fit. You only need it if you are about to call eval so that you can plug your guess.

jmfranck: In principle, I think that we could actually edit the class so that we don't need `set_to_guess` at all… as long as it knew whether we called `set_to_guess` or `fit` last, it should be able to figure out what we want when we call eval... put in your own words and create an issue on pySpecData.

jmfranck (changes requested):

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,174 @@
+"""Fitting ODNP Datasets for Ksigma
+=================================== 
+The T1(p) and E(p) integrals are generated
+using the example, `generate_integrals.py`
+(as of 11/8/23 the `generate_integrals` is not
+updated), and stored in
+previous post processing pulled from an H5 file.
+The inverse of the :math:`T_1(p)` data processing is taken
+to produce the :math:`R_1(p)` data which is fit using the
+polyfit of the inverse of krho with two degrees of
```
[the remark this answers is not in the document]
guinneak198: I added the equation but am unfamiliar with this :math: formatting, I think I did this right? but double check

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,171 @@
+"""Fitting ODNP Datasets for Ksigma
+=================================== 
+The T1(p) and E(p) integrals are generated
+using the example, `generate_integrals.py`
+(as of 11/8/23 the `generate_integrals` is not
+updated), and stored in an H5 file.
+The inverse of the :math:`T_1(p)` data processing is taken
+to produce the :math:`R_1(p)` data which is fit using the
+polyfit of the inverse of krho with two degrees of
+freedom (*e.g.*, :math:`R_{100}(p) + \frac{C}{k_{\rho}^{-1}}`.
```
jmfranck, line 10: I think that you are saying:
```
so that :math:`R_1(p) = R_{1,0}(p) + \frac{C}{k_{\rho}^{-1}(p)}`
```

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,171 @@
...
+using the example, `generate_integrals.py`
+(as of 11/8/23 the `generate_integrals` is not
+updated), and stored in an H5 file.
+The inverse of the :math:`T_1(p)` data processing is taken
+to produce the :math:`R_1(p)` data which is fit using the
+polyfit of the inverse of krho with two degrees of
+freedom (*e.g.*, :math:`R_{100}(p) + \frac{C}{k_{\rho}^{-1}}`. 
+The :math:`R_1(p)` fit is then fed into the
+fitting routine for the integrals as a function of power
+prior to normalization. The cross relaxivity is then
```
jmfranck, line 13: I don't follow.

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,171 @@
...
+from scipy.io import loadmat
+
+# {{{ This block changes when the data changes -- everything else should be
+# left alone for most circumstances
+data_info = dict(
+    filename="ras.h5",  # h5 file containing table of integrals for different datasets
+    data_dir="AG_processed_data",  # directory of the dataset of table of integrals
+    nodename="230706_M67_a",# specific nodename of the dataset of interest
+)  
+T100_info = dict(
```
jmfranck, line 29: why T100? Where is this notation from?

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,171 @@
...
+        data_info["filename"],
+        exp_type=data_info["data_dir"],
+        expno=f"{data_info['nodename']}/T1p",
+    )
+    R1p = 1 / T1p
+# }}}
+# }}}
+# {{{ The powers go up, and then go back down in order to check for
+# reproducibility. Figure out where this flip occurs
+flip_idx = np.where(np.diff(integral_vs_p.fromaxis("power").data) < 0)[0][0] + 1
```
jmfranck, line 61: `fromaxis("power").data` is equivalent to `getaxis("power")` here, so use the latter

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,171 @@
...
+    fl.plot(
+        integral_vs_p["power", :flip_idx],
+        "o",
+        label="Progressive Saturation Data",
+        capsize=6,
+        alpha=0.5,
+    )
+    fl.plot(
+        integral_vs_p["power", flip_idx:],
+        "rx",
```
jmfranck, line 75: "ro"

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,171 @@
...
+# {{{ The powers go up, and then go back down in order to check for
+# reproducibility. Figure out where this flip occurs
+flip_idx = np.where(np.diff(integral_vs_p.fromaxis("power").data) < 0)[0][0] + 1
+# }}}
+with figlist_var() as fl:
+    # {{{Plot integrals as a function of power
+    fl.next("Integrals vs power")
+    fl.plot(
+        integral_vs_p["power", :flip_idx],
+        "o",
```
jmfranck, line 68: "ko"

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,171 @@
...
+    krho_inv = integral_vs_p.get_prop("acq_params")["concentration"] / (R1p - R10_p)
+    krho_inv_coeff = krho_inv.polyfit("power", order=1)
+    M0, A, phalf, p = symbols("M0 A phalf power", real=True)
+    R1p_expression = (T10_p[0] + T10_p[1] * p) ** -1 + (
+        integral_vs_p.get_prop("acq_params")["concentration"]
+        / (krho_inv_coeff[0] + krho_inv_coeff[1] * p)
+    )
+    R1p_fit = lambdify(p, R1p_expression)
+    fl.plot(
+        R1p_fit(powers_fine),
```
jmfranck, line 103: I would just combine: `lambdify(p, R1p_expression)(powers_fine)`

examples/fit_ODNP_data.py
```diff
@@ -0,0 +1,171 @@
...
+    plt.ylabel(r"$R_{1} / s^{-1}$")
+    plt.xlabel("Power / W")
+    # }}}
+    # {{{ Fit NMR integrals as function of power
+    fl.next("Integrals vs power")
+    sp = p / (p + phalf)
+    integral_vs_p_fit = lmfitdata(integral_vs_p["power", :flip_idx])
+    # Symbolic expression for integrals as a function of power that is used
+    # in the symbolic function for the fitting of the integrals as a function of power
+    integral_vs_p_fit.functional_form = M0 - ((M0 * A * sp) / R1p_expression)
```
jmfranck, line 117: call `sp` `sp_expression`  (or use `_expression` instead of `_expr` across the board)

this is for legibility --  so you know what type of object it is

jmfranck (changes requested):
this is looking pretty good.  It should be possible to quickly deal with these comments one by one, and then we're done.  If you do that, I don't have to read through the whole diff again.

examples/fit_ODNP_data.py
guinneak198: T10 = relaxation of water with no spin label. T100 = relaxation of water with no spin label at 0 power - good catch as this is not at 0 power. Edited to T10. Notation from methods in enzymology

examples/fit_ODNP_data.py
jmfranck: in modified version, you need double backslash, or to put r in front of triple quotes.

examples/fit_ODNP_data.py
jmfranck: This still isn't quite specific enough.  You are using the polynomial coefficients as part of the fit expression, though they fixed, and not fit parameters.

Also, the last sentence of this docstring is still sketchy -- ksigma is a parameter of the final fit.  Explain how.

So, please edit the docstring with these thoughts in mind.

examples/fit_ODNP_data.py
jmfranck: that's not correct.

examples/fit_ODNP_data.py
guinneak198: Tried to be as succinct and to the point as I could.

examples/fit_ODNP_data.py
guinneak198: renamed to water_relax_info

examples/fit_ODNP_data.py
jmfranck: this was removed? why?

examples/fit_ODNP_data.py
jmfranck: I just edit this.

examples/fit_ODNP_data.py
```diff
@@ -1,14 +1,24 @@
...
+(where
+:math:`R^{-1}_{1,HH}/C_{SL}`
+is assumed to be small or constant),
+and fit it to a line.
+The linear fit coefficients are used
+for two purposes:
+
+*   directly, to directly calculate the fit value of :math:`R_1(p)`
+*   they are plugged in, as raw numbers, as part of the nonlinear expression
+    that is used to fit the :math:`M_0 E(p)` (*i.e.* the raw integral) values.
```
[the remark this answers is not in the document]
guinneak198: okay wrote out math and this looks okay

PR #90, opened 2023-12-11
all examples and other calls to hermitian function test updated

guinneak198:
use hermitian function test as described in coherence paper and update examples/functions that use it


jmfranck:
As discussed, let's instead pull from jf_review_actual. Also, it's likely too big. PR should be as small as possible to make a useful change.  There were 10 fingers changed here

PR #92, opened 2024-02-13
AG edits to phasing

guinneak198:


pyspecProcScripts/phasing.py
```diff
@@ -186,15 +186,48 @@ def costfun(ph1):
     # }}}
 
 
-def fid_from_echo(d, signal_pathway, fl=None, add_rising=False, fraction_nonnoise=0.1, direct="t2",
```
guinneak198, line 189: fraction_nonnoise wasn't used anywhere

pyspecProcScripts/phasing.py
```diff
@@ -186,15 +186,48 @@ def costfun(ph1):
...
+    fl:             figlist or None (default)
+                    If you want the diagnostic plots (showing the
+                    distribution of the data in the complex plane),
+                    set this to your figlist object.
+    add_rising:     boolean
+                    when true the script will add a rising edge to the echo
+    direct:         string
+                    Name of the direct dimension
+    exclude_rising: int
+                    if add_rising is True this is the number of points left out of the
```
guinneak198, line 209: maybe go over these together?

pyspecProcScripts/phasing.py
```diff
@@ -339,6 +371,7 @@ def filter_ranges(B, A):
             d[direct, 1 : N_rising + 1]
             + d_rising[direct, ::-1]
         ) / 2
+    d *= 2
```
guinneak198, line 374: I might be imagining this but I thought for the FID slice we multiply by 2 to conserve something (energy?) since we cut the first half out in the line above?

pyspecProcScripts/phasing.py
```diff
@@ -186,12 +186,44 @@ def costfun(ph1):
...
+                                    diagnostic in checking the sign of the signal prior
+                                    to the hermitian phase correction
+    show_shifted_residuals: boolean
+                            diagnostic in analyzing the residuals after the hermitian 
+                            phase correction.
+    Returns                                 
+    =======
+    d: FID of properly sliced and phased signal
+    """
     # {{{ autodetermine slice range
```
guinneak198, line 227: Take envelope of echo and convolve. Then pick the narrow and wider ranges using contiguous function

pyspecProcScripts/phasing.py
```diff
@@ -221,7 +253,7 @@ def filter_ranges(B, A):
         return [np.array(b) for b in B if
```
guinneak198, line 253: I really don't fully understand what this is doing/why it is better than just using contiguous once to decide the range? notes in notebook didn't help

pyspecProcScripts/phasing.py
```diff
@@ -186,12 +186,44 @@ def costfun(ph1):
...
+    Parameters
+    ==========
+    signal_pathway: dict
+                    coherence transfer pathways that correspond to the signal
+    fl:             figlist or None (default)
+                    If you want the diagnostic plots (showing the
+                    distribution of the data in the complex plane),
+                    set this to your figlist object.
+    add_rising:     boolean
+                    when true the script will add a rising edge to the echo
```
jmfranck, line 205: It's taking the first part of the echo... The part that rises up to the maximum, and it's adding it to the decaying (FID like) part. We want to do this if it works well in order to increase the SNR of the early points of the signal... But of course, there are some potential pitfalls

pyspecProcScripts/phasing.py
jmfranck: In general, you assume that the first few points of the signal might be messed up in some fashion from dead time and other issues. So, when doing what I describe above, you might want to leave out the first few points that's what this is for

pyspecProcScripts/phasing.py
```diff
@@ -186,12 +186,44 @@ def costfun(ph1):
...
+    direct:         string
+                    Name of the direct dimension
+    exclude_rising: int
+                    if add_rising is True this is the number of points left out of the 
+                    start echo
+    slice_multiplier:   int
+                        in determining the autoslice this is a multiplier where the higher
+                        the value the wider the slice
```
jmfranck, line 213: Is it possible to describe in words roughly what we are multiplying? I want to say it's the width of the signal when you take the absolute value or something like that

pyspecProcScripts/phasing.py
```diff
@@ -186,12 +186,44 @@ def costfun(ph1):
...
+    peak_lower_thresh:  float
+                        multiplier in deciding the wider range of the signal, the lower
```
jmfranck, line 215: I don't think this is a multiplier... I think it's a fraction of the peak signal intensity

pyspecProcScripts/phasing.py
```diff
@@ -186,12 +186,44 @@ def costfun(ph1):
...
+                        the value the wider the range
+    show_hermitian_sign_flipped:    boolean
```
jmfranck, line 217: What you say here is good, and can help to guide the choice of this number, but but you want to add to this to explain how this number is used... Used... I think that it has something to do with choosing a width by looking at when the peak rises above a certain threshold fraction of the max

This does not pertain to the hermitian parameter... It pertains to the line above it... Sorry

pyspecProcScripts/phasing.py
guinneak198: edited the description of add_rising

pyspecProcScripts/phasing.py
```diff
@@ -186,12 +186,54 @@ def costfun(ph1):
...
+    slice_multiplier:   int
+                        The calculated frequency slice is calculated by taking the
+                        center frequency and extending out to values that are included
+                        in the peak times this multiplier. Therefore the larger this 
+                        value the larger the frequency slice.
+    peak_lower_thresh:  float
+                        fraction of the signal intensity used in calculating the 
+                        frequency slice. The smaller the value, the larger 
+                        wider the slice. This is especially useful for noisy data
+                        where this value may need to be decreases in order to include
```
jmfranck, line 225: This is not a good suggestion, I think. If it's noisy, you would want to raise this and raise the slice multiplier, right?

jmfranck (changes requested):

pyspecProcScripts/phasing.py
guinneak198: no lowering this value increases the wide_range (also in using this with the notebook examples I found this to be true. Increasing this or using the default for noisy data results in the assert statement at line 266. But will add the increase to the slice multiplier

jmfranck: Yes, but that's problematic because you're increasing the range by locking on to noise

pyspecProcScripts/phasing.py
```diff
@@ -221,7 +264,7 @@ def filter_ranges(B, A):
         return [np.array(b) for b in B if
                 any(b[0] <= a[0] and b[1] >= a[1] for a in A)]
     peakrange = filter_ranges(wide_ranges, narrow_ranges)
-    assert len(peakrange) == 1
+    assert (len(peakrange) == 1),"""Your wide range is too small try decreasing the peak_lower_thresh"""
```
jmfranck, line 267: That's not what this means.  Rather, have it say it found two peaks, and give the location of both peaks.

jmfranck: If this is not working, I think I'll need to help by seeing a specific example of a dataset where it doesn't work, with MWE showing that it doesn't work

jmfranck (approved):
On line 241, we assume 0 is set to the ppg tau. Is that clear in the docstring?

I disagree that you want to increase wide range with lower noise.  Why would you do that?

See my comments.  I am thinking that if you are having low SNR that's throwing the assertion error, we're going to want to increase the convolution width rather than doing what you're doing.

pyspecProcScripts/phasing.py
guinneak198: the lower threshold is multiplied by the max of the signal (e.g. the peak) assuming a low SNR whether that's lower signal or larger noise simply pushes the wide range limits out. Additionally, if the signal is convolved and simply has low signal and a very broad linewidth, decreasing this value ensures that the lower limits of the signal is included. Further this isn't the final autoslice - this is JUST for deciding the wide range which is fed into the filter_range runction.

guinneak198: Thinking out loud so that you can correct me if I'm wrong: according to your later comment isn't this rather locking onto multiple peaks? e.g. if the noise is on scale with the signal and it considers that a peak? So lowering this threshold does not guarantee that it will actually select the signal

pyspecProcScripts/phasing.py
guinneak198: pushed a notebook FIDfromEcho_fail.py

.gitignore
```diff
@@ -17,3 +17,8 @@ doc_src/auto_api/
 doc_src/auto_examples/
 auto_figures/
 enhancement.png
+examples/data_files.csv
+examples/single_mutant_overlay.pdf
+pyspecProcScripts.egg-info/
+pyspecProcScripts/__init__.py.orig
```
jmfranck, line 23: just delete this file and remove from here (the .orig are created when resolving conflicts)

examples/FIDfromEcho_fail.py
```diff
@@ -0,0 +1,31 @@
+# In[1]:
```
jmfranck, line 1: Can this be integrated with an existing example? I thought there was one where you were looking over different files?

examples/FIDfromEcho_fail.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+    s.mean("nScans")
+    fl.next("raw")
+    fl.image(s)
+    fl.next("raw time domain")
+    s.ift('t2')
+    fl.image(s['t2':(None,80e-3)])
+    s.ft('t2')
+    # autoslice, phase and take FID slice
+    s = fid_from_echo(s, signal_pathway = s.get_prop('coherence_pathway'), fl=fl)
```
jmfranck, line 24: with comment above, can we do this for all?
Also, run black.

examples/FIDfromEcho_fail.py
```diff
@@ -0,0 +1,31 @@
...
+    s.ift('t2')
+    fl.image(s['t2':(None,80e-3)])
+    s.ft('t2')
+    # autoslice, phase and take FID slice
+    s = fid_from_echo(s, signal_pathway = s.get_prop('coherence_pathway'), fl=fl) 
+    s = select_pathway(s, s.get_prop('coherence_pathway'))
+    fl.next("phased and FID sliced")
+    fl.image(s)
+
+# In[2]:
```
jmfranck, line 29: even if too much of a pain to integrate with other examples, this should be removed

examples/FIDfromEcho_fail.py
```diff
@@ -0,0 +1,31 @@
...
+
```
jmfranck, line 30: ideally, then you can create a new PR, where you just add to the loop with your fussy nutation experiment, and we can make sure the whole example runs.

pyspecProcScripts/load_data.py
```diff
@@ -128,6 +128,9 @@ def proc_spincore_CPMG_v1(s, fl=None):
     s.setaxis("ph1", r_[0.0, 2.0] / 4)
     s.setaxis("tE", tE_axis)
     s.setaxis("t2", t2_axis)
+    s *= s.shape["nScans"]
+    s.squeeze()
+    s.set_prop('coherence_pathway',{'ph1':1})
```
jmfranck, line 133: this is nice.

run black, though

pyspecProcScripts/load_data.py
```diff
@@ -297,9 +300,14 @@ def proc_Hahn_echoph(s, fl=None):
...
     s.chunk("t", ["ph2", "ph1", "t2"], [2, 4, -1])
     s.labels({"ph2": r_[0.0, 2.0] / 4, "ph1": r_[0.0, 1.0, 2.0, 3.0] / 4})
+    s *= s.shape["nScans"]
+    s.squeeze()
+    s.set_prop('coherence_pathway',{'ph1':1,'ph2':-2})
     s.reorder(["ph2", "ph1"])
     s.setaxis("nScans", r_[0:nScans])
     s.reorder("t2", first=False)
+    s.set_units("t2", "s")
+    s.setaxis("t2", lambda x: x - s.get_prop("acq_params")["tau_us"] * 1e-6)
```
jmfranck, line 310:
```
s["t2"] -= s.get_prop("acq_params")["tau_us"]*1e-6
```
just because it's easier.

pyspecProcScripts/load_data.py
```diff
@@ -340,7 +353,12 @@ def proc_spincore_IR_v2(s, fl=None):
         s.chunk("t", ["ph2", "ph1", "t2"], [4, 4, -1])
     s.setaxis("ph1", r_[0, 1, 2, 3.0] / 4)
     s.setaxis("ph2", r_[0, 1, 2, 3.0] / 4)
+    s *= s.shape["nScans"]
+    s.squeeze()
+    s.set_prop('coherence_pathway',{'ph1':0,'ph2':-1})
     s.reorder(["ph1", "ph2"]).set_units("t2", "s")
+    s.set_units("t2", "s")
+    s.setaxis("t2", lambda x: x - s.get_prop("acq_params")["tau_us"] * 1e-6)
```
jmfranck, line 361: see comment above

pyspecProcScripts/load_data.py
```diff
@@ -466,8 +500,10 @@ def proc_spincore_ODNP_v1(s, fl=None):
     s.chunk("t", ["ph1", "t2"], [4, -1])
     s.set_units("t2", "s")
     s.labels({"ph1": r_[0.0, 1.0, 2.0, 3.0] / 4})
-    s.setaxis("t2", lambda x:
-            x-s.get_prop("acq_params")["tau_us"]*1e-6)
+    s.setaxis("t2", lambda x: x - s.get_prop("acq_params")["tau_us"] * 1e-6)
```
jmfranck, line 503: see above

pyspecProcScripts/phasing.py
```diff
@@ -283,14 +342,21 @@ def filter_ranges(B, A):
...
         d.setaxis(direct, lambda x: x - test_shift).register_axis({direct: 0})
-        ph0 = zeroth_order_ph(select_pathway(d,
-            signal_pathway)[direct:0.0],
-            fl=zeroth_fl)
+        if signal_pathway is not None:
+            ph0 = zeroth_order_ph(select_pathway(d,
+                signal_pathway)[direct:0.0],
+                fl=zeroth_fl)
+        else:
+            ph0 = zeroth_order_ph(d[direct:0.0],fl = zeroth_fl)
```
jmfranck, line 350: I think instead, just modify `select_pathway` so that it accepts `None` as an argument (and just returns the original data)

pyspecProcScripts/phasing.py
```diff
@@ -283,14 +342,21 @@ def filter_ranges(B, A):
...
         if fl is not None:
             t_start = d.getaxis(direct)[0]
-            d_sigcoh = select_pathway(d, signal_pathway)[direct : (t_start, -2 * t_start)]
-            d_sigcoh = select_pathway(d, signal_pathway).squeeze()
+            if signal_pathway is not None:
+                d_sigcoh = select_pathway(d, signal_pathway)[direct : (t_start, -2 * t_start)]
+                d_sigcoh = select_pathway(d, signal_pathway).squeeze()
+            else:
+                d_sigcoh = d[direct : (t_start, -2 * t_start)]
+                d_sigcoh = d.squeeze()
```
jmfranck, line 359: same comment as prev

jmfranck (changes requested):

examples/FIDfromEcho_fail.py
guinneak198: moved into FIDtoEcho_Actual_Challenging.py

examples/FIDfromEcho_fail.py
guinneak198: see above comment - this file no longer exists

examples/FIDfromEcho_fail.py
guinneak198: see above comment- this file no longer exists

pyspecProcScripts/load_data.py
```diff
@@ -317,7 +324,14 @@ def proc_spincore_IR(s, fl=None):
...
     s.setaxis("ph2", r_[0, 2.0] / 4)
-    s.reorder(["ph1", "ph2"]).set_units("t2", "s")
+    s.set_prop("coherence_pathway", {"ph1": 0, "ph2": -1})
+    s *= s.shape["nScans"]
+    s.squeeze()
+    if 'nScans' in s.dimlabels:
+        s.reorder(["ph1", "ph2", "nScans", "vd", "t2"]).set_units("t2", "s")
+    else:
+        s.reorder(["ph1", "ph2", "vd", "t2"]).set_units("t2", "s")
+    #s["t2"] -= s.get_prop("acq_params")["tau_us"] * 1e-6
```
jmfranck, line 334: don't include commented out lines

examples/FIDtoEcho_Actual_Challenging.py
```diff
@@ -18,41 +18,71 @@
 from collections import OrderedDict
 from numpy.random import normal
 from scipy.signal import tukey
-init_logging(level='debug')
+init_logging(level="debug")
```
jmfranck, line 21: you need to run `git merge --no-ff jf_review_actual` and resolve any conflicts.  I have run black on jf_review_actual, so that will clean up this diff, but there are other unmerged differences.

jmfranck: (after pulling jf_review_actual to update it)

examples/FIDtoEcho_Actual_Challenging.py
```diff
@@ -18,41 +18,70 @@
...
+        data = find_file(
+            filename,
+            exp_type=file_location,
+            expno=nodename,
+            postproc=postproc,
+            lookup=lookup_table,
+        )
         fl.basename = "(%s)" % label
-        fig, ax_list = subplots(1, 3, figsize = (7,7))
+        fig, ax_list = subplots(1, 4, figsize=(10, 7))
```
guinneak198, line 70: bigger sizing is easier to see

examples/FIDtoEcho_Actual_Challenging.py
```diff
@@ -18,41 +18,70 @@
...
-        fl.image(data['t2':(-1e3,1e3)], ax=ax_list[0])
+        fl.image(data["t2":(-1e3, 1e3)], ax=ax_list[0])
         ax_list[0].set_title("Raw Data")
-        data = data['t2':f_range]
-        fl.basename = "(%s)"%label
-        data = fid_from_echo(data, signal_pathway,
-                fl=fl)
-        fl.image(data['t2':(-1e3,1e3)], ax=ax_list[1],
-                human_units=False)
-        ax_list[1].set_title("Phased and centered (ν)")
```
guinneak198, line 54: we are making the second image the raw data in the t domain first before doing the fid_from_echo

pyspecProcScripts/__init__.py
```diff
@@ -22,12 +22,10 @@
 from .integral_w_error import integral_w_errors, active_propagation
 from .apod_matched_filter import apod_matched_filter
 from .envelope import L2G, fit_envelope
-from .first_level.QESR_rescale import QESR_scalefactor
```
guinneak198, line 25: this PR hasn't been merged into master yet so at this point this line shouldn't be here

pyspecProcScripts/phasing.py
```diff
@@ -222,6 +262,21 @@ def filter_ranges(B, A):
...
+        if any(np.array(range_gaps) > max_range_width):
+            if fl is not None:
+                fl.next("debug filter ranges")
+                fl.plot(freq_envelope, human_units=False)
+                for thisrange in peakrange:
+                    fl.plot(freq_envelope[direct:thisrange], human_units=False)
+            raise ValueError("finding more than one peak!")
+        else:
+            peakrange = [(peakrange[0][0],peakrange[-1][1])]
+        # }}}
```
guinneak198, line 279: handles when more than one peak is found by identifying "breaks" in the peak vs the actual large peak

pyspecProcScripts/phasing.py
```diff
@@ -245,7 +300,11 @@ def filter_ranges(B, A):
     else:
         thebasename = ""
     # {{{ sign flip and average input for hermitian
-    input_for_hermitian = select_pathway(d, signal_pathway).C
+    if signal_pathway is not None:
+        input_for_hermitian = select_pathway(d, signal_pathway).C
+    else:
+        logger.info(strm("You are telling me that you did not apply phase cycling, so I am not selecting a coherence pathway"))
+        input_for_hermitian = d.C
```
guinneak198, line 307: handle both data that has and has not been phase cycled

examples/FIDtoEcho_Actual_Challenging.py
```diff
@@ -18,41 +18,70 @@
...
-        ('tau_1000','ODNP_NMR_comp/Echoes','spincore_echo_v1',
-            'tau is 1 ms',1),
-        ('tau_3500','ODNP_NMR_comp/Echoes','spincore_echo_v1',
-            'tau is 3.5 ms',3),
-        ('tau_11135','ODNP_NMR_comp/Echoes','spincore_echo_v1',
-            'tau is 11.135 ms',3),
-            ]:
-        data = find_file(filename,exp_type=file_location,expno=nodename,
-                postproc=postproc,lookup=lookup_table)
+    for filename, nodename, file_location, postproc, label in [
```
guinneak198, line 32: added in an IR at no power so now filename is one of the arguments

examples/FIDtoEcho_Actual_Challenging.py
guinneak198: also made the raw t domain part of the subplots instead of it's own figure

pyspecProcScripts/phasing.py
```diff
@@ -186,17 +186,56 @@ def costfun(ph1):
...
+    Returns                                 
+    =======
+    d: nddata
+        FID of properly sliced and phased signal
+    """
     # {{{ autodetermine slice range
     freq_envelope = d.C
-    freq_envelope.ift('t2')
-    freq_envelope = freq_envelope['t2':(0,None)] # slice out rising echo estimate according to experimental tau in order to limit oscillations
-    freq_envelope.ft('t2')
```
guinneak198, line 199: see [thread](https://jmfrancklab.slack.com/archives/CV553TPRV/p1719501246917169) - this results in the rising time for some datasets to be thrown out when the correct time axis is applied

pyspecProcScripts/phasing.py
jmfranck: why?

jmfranck: I don't understand this -- it seems to me like if signal pathway is None, then you pull from the properties

jmfranck: so you would have a variable for your pathway, that would be either pulled from the arg, or else pulled from the properties

examples/FIDtoEcho_Actual_Challenging.py
```diff
@@ -70,11 +67,16 @@
...
         ax_list[0].set_title("Raw Data")
         data = data["t2":f_range]
+        fl.next("Data processing", fig=fig)
         fl.basename = "(%s)" % label
-        data = fid_from_echo(data, signal_pathway, fl=fl)
+        data = fid_from_echo(
+            data,
+            data.get_prop("coherence_pathway"),
+            fl=fl,
+        )
```
jmfranck, line 76: w/ my other comment, this would not change

jmfranck: for that data that's used here, you can't pull the coh pathway b/c it's not set in the data

examples/FIDtoEcho_Actual_Challenging.py
```diff
@@ -70,11 +67,16 @@
...
         fl.image(data["t2":(-1e3, 1e3)], ax=ax_list[1], human_units=False)
         ax_list[1].set_title("Phased and centered (ν)")
         data.ift("t2")
-        fl.image(data, ax=ax_list[2], human_units=False)
+        fl.image(data["t2":(None, 100e-3)], ax=ax_list[2], human_units=False)
```
jmfranck, line 80: I really don't think this change to a hard-coded 100 ms cutoff is justified

pyspecProcScripts/load_data.py
```diff
@@ -306,6 +309,10 @@ def proc_bruker_CPMG_v1(s, fl=None):
 
 def proc_spincore_SE_v1(s, fl=None):
     s.ft("ph1")
+    s.set_units('t2','s')
```
jmfranck, line 312: run black

pyspecProcScripts/load_data.py
```diff
@@ -363,7 +378,14 @@ def proc_spincore_IR(s, fl=None):
         s.chunk("t", ["ph2", "ph1", "t2"], [2, 2, -1])
     s.setaxis("ph1", r_[0, 2.0] / 4)
     s.setaxis("ph2", r_[0, 2.0] / 4)
-    s.reorder(["ph1", "ph2"]).set_units("t2", "s")
```
jmfranck, line 366: why is this removed??

jmfranck: I see what you're doing below, but it seems like a messier and more verbose way of doing the same thing

pyspecProcScripts/load_data.py
```diff
@@ -462,6 +500,26 @@ def proc_nutation_chunked(s, fl=None):
     return s
 
 
+def proc_nutation_v2(s, fl=None):
+    logging.info("loading pre-processing for nutation")
```
jmfranck, line 504: don't add `info` unless they're really well justified.  `debug` is fine.

pyspecProcScripts/load_data.py
```diff
@@ -462,6 +500,26 @@ def proc_nutation_chunked(s, fl=None):
...
+    s.set_units("indirect", "s")
+    s.ft(["ph1"], unitary=True)
```
jmfranck, line 506: you have to decide, in general, if this is going to be unitary or not, and be consistent across all postproc

pyspecProcScripts/load_data.py
```diff
@@ -471,6 +529,10 @@ def proc_nutation_v4(s, fl=None):
     s.rename("indirect", "p_90")
     s.set_units("t2", "s")
     s.set_units("p_90", "s")
+    s.set_prop("coherence_pathway", {"ph1": 1})
+    s["t2"] -= s.get_prop("acq_params")["tau_us"] * 1e-6
+    s *= s.shape["nScans"]
```
jmfranck, line 534: version 4 of the nutation doesn't have fixed nScans?

pyspecProcScripts/phasing.py
```diff
@@ -186,17 +186,56 @@ def costfun(ph1):
     # }}}
 
 
-def fid_from_echo(d, signal_pathway, fl=None, add_rising=False, fraction_nonnoise=0.1, direct="t2",
```
jmfranck, line 189: this `fraction_nonnoise` isn't used?

pyspecProcScripts/phasing.py
```diff
@@ -339,6 +404,7 @@ def filter_ranges(B, A):
             d[direct, 1 : N_rising + 1]
             + d_rising[direct, ::-1]
         ) / 2
+    d *= 2
```
jmfranck, line 407: in general, I think this PR is too big.  Try again

jmfranck:
You are doing multiple things at once, and a lot of it is not easy to sign off on, so please try again with more shorter PRs.

pyspecProcScripts/phasing.py
guinneak198: added the pulling from properties and an assert statement if that property is not set

examples/FIDtoEcho_Actual_Challenging.py
guinneak198: it's set in the preprocessing

pyspecProcScripts/phasing.py
guinneak198: not used

jmfranck: note that this PR has been closed

PR #94, opened 2024-03-29
add function for opening log that doesnt require Instruments

guinneak198:


jmfranck:
Can you add an example?  This looks good. Let's just make sure we can test it

examples/read_log.py
```diff
@@ -0,0 +1,53 @@
...
+myfilename = search_filename(
+    "230626_batch230515_E37_Ras_B10_ODNP_1.h5",
+    exp_type="ODNP_NMR_comp/ODNP",
+    unique=True,
+)
+with figlist_var() as fl:
+    # {{{ open h5 file to real log
+    with h5py.File(myfilename, "r") as f:
+        log_grp = f["log"]
+        thislog = logobj.from_group(log_grp)
```
jmfranck, line 17: Combine with last line

examples/read_log.py
```diff
@@ -0,0 +1,53 @@
...
+        read_array = thislog.total_log
+        read_dict = thislog.log_dict
```
jmfranck, line 19: These assignments don't help with clarity.  Don't introduce these variables.

examples/read_log.py
```diff
@@ -0,0 +1,53 @@
...
+    # }}}
+    # {{{ separate the time, RX, power, and commands from the
+    #     log array
+    for j in range(len(read_array)):
+        thistime, thisrx, thispower, thiscmd = read_array[j]
```
jmfranck, line 24: What is this? This loop isn't doing anything?

examples/read_log.py
```diff
@@ -0,0 +1,53 @@
...
+    # }}}
+    # In order to properly set the time axis to start at 0
+    # both the log's start time will be subtracted from the
+    # the relative time recorded
+    log_start_time = read_array["time"][0]
+    relative_time = read_array["time"]
+    time_axis = relative_time - log_start_time
```
jmfranck, line 31: Again, these three lines can be one line... you only want time_axis

pyspecProcScripts/phasing.py
```diff
@@ -423,38 +423,3 @@ def hermitian_function_test(
     return best_shift
 
 
-def determine_sign(s, direct="t2", fl=None):
```
jmfranck, line 426: What are you doing here? This doesn't have to do with the log, so why mess with this?

jmfranck (changes requested):

examples/read_log.py
guinneak198: didn't realize that none of these are used, my bad should have realized before asking you to review!

examples/read_log.py
guinneak198: I thought it would be more legible for someone reading this but if you think otherwise-I changed it

PR #95, opened 2024-04-08
there's no need to have two scripts that define the same function

guinneak198:


pyspecProcScripts/phasing.py
```diff
@@ -421,40 +421,3 @@ def hermitian_function_test(
...
-    """
-    assert s.get_ft_prop(
-        direct
-    ), "this only works on data that has been FT'd along the direct dimension"
-    if fl is not None:
-        fl.push_marker()
-        fl.next("selected pathway")
-        fl.image(s.C.setaxis("vd", "#").set_units("vd", "scan #"))
-    data_sgn = s.C.sum(direct)
-    data_sgn /= zeroth_order_ph(data_sgn)
```
jmfranck, line 454: I agree with what you're doing, but keep this code, which is more robust (compare this func to the one you are keeping... they are not the same)

jmfranck: Actually keep this.  It relies on phasing, so it's not a simple function

jmfranck (changes requested):

PR #96, opened 2024-06-20
Raw proc

jmfranck:
add example to proc scripts that uses the `postproc_type` and `coherence_pathway` properties to show the raw data.

### DIFF
--- a/examples/proc_raw.py
+++ b/examples/proc_raw.py
@@ -5,6 +5,15 @@ Show data with postproc
 
 Fourier transforms (and any needed data corrections for older data) are performed according to the `postproc_type` attribute of the data node.
 This script plots the result, as well as signal that's averaged along the `nScans` dimension.
+
+Tested with:
+
+``py proc_raw.py echo_6 240620_200uM_TEMPOL_pm_echo.h5 ODNP_NMR_comp/Echoes``
+
+``py proc_raw.py echo_8 240620_200uM_TEMPOL_pm_generic_echo.h5 ODNP_NMR_comp/Echoes``
+
+``py proc_raw.py CPMG_9 240620_200uM_TEMPOL_pm_generic_CPMG.h5 ODNP_NMR_comp/Echoes``
+
 """
 from pyspecProcScripts.load_data import lookup_table
 from pyspecProcScripts import select_pathway
--- a/examples/proc_raw.py
+++ b/examples/proc_raw.py
@@ -73,3 +73,6 @@ with figlist_var() as fl:
         d = select_pathway(d, d.get_prop("coherence_pathway"))
         fl.next("with coherence pathway selected")
         image_or_plot(d)
+        if len(d.dimlabels) == 2:
+            fl.next("pcolor of selected coherence pathway")
+            d.pcolor()
--- a/examples/proc_raw.py
+++ b/examples/proc_raw.py
@@ -65,10 +65,10 @@ with figlist_var() as fl:
         fl.next("signal averaged along nScans")
         image_or_plot(d)
     if d.get_prop("coherence_pathway") is not None:
-        forplot = abs(d)
-        for thisdim in d.get_prop("coherence_pathway").keys():
-            forplot.sum(thisdim)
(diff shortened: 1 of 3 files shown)


examples/proc_raw.py
```diff
@@ -0,0 +1,74 @@
...
+
+    image_or_plot(d)
+    if "nScans" in d.dimlabels:
+        d.mean("nScans")
+        fl.next("signal averaged along nScans")
+        image_or_plot(d)
+    if d.get_prop("coherence_pathway") is not None:
+        forplot = abs(d)
+        for thisdim in d.get_prop("coherence_pathway").keys():
+            forplot.sum(thisdim)
```
guinneak198, line 69: why/I don't understand why we would sum the coherence pathway? (there's no documentation description for nddata.sum in the documentation)

pyspecProcScripts/load_data.py
```diff
@@ -287,10 +287,24 @@ def proc_bruker_CPMG_v1(s, fl=None):
...
+    return s
+def proc_spincore_diffph_SE_v1(s, fl=None):
+    r"""this one uses a phase cycle where the overall phase and 90-180
+    phase difference are cycled in a nested way -- see the DCCT paper to
+    understand this!"""
+    s.ft(["ph2","ph_diff"], unitary=True)
+    # after the FT, these have a different meaning in terms of coherence
+    # pathways -- remember that when labeling, pySpecData will change the
+    # ph here to a δp
+    s.rename("ph2","ph_overall") # overall change in coherence
```
guinneak198, line 302: why not just call this ph_overall in the actual pulse sequence before saving?

guinneak198 (commented):
AG leaves questions

examples/proc_raw.py
```diff
@@ -0,0 +1,74 @@
...
+Show data with postproc
+====================
+`py proc_raw_data.py NODENAME FILENAME EXP_TYPE`
+
+Fourier transforms (and any needed data corrections for older data) are performed according to the `postproc_type` attribute of the data node.
+This script plots the result, as well as signal that's averaged along the `nScans` dimension.
+
+Tested with:
+
+``py proc_raw.py echo_6 240620_200uM_TEMPOL_pm_echo.h5 ODNP_NMR_comp/Echoes``
```
guinneak198, line 11: This doesn't work for me I get "I caught and error but am plotting anyways and nothing shows up in the resulting window

jmfranck: Then debug

jmfranck: note that this does work OK on my laptop (remotely) and on the NMR computer.

examples/proc_raw.py
jmfranck: this was a good comment -- this is fixed in the updated code.

examples/proc_raw.py
guinneak198: needed to pull pyspecdata updates

examples/proc_raw.py
```diff
@@ -0,0 +1,82 @@
...
+        fl.next("signal averaged along nScans")
+        image_or_plot(d)
+    if d.get_prop("coherence_pathway") is not None:
+        fl.next("sum of abs of all coherence pathways (for comparison)")
+        forplot = abs(d)
+        guess_direct = d.shape.max() # guess that the longest dimension is the direct
+        if guess_direct == 'nScans':
+            temp = d.shape
+            temp.pop('nScans')
+            guess_direct = temp.max()
```
guinneak198, line 77: how would it be possible for guess_direct to be nScans if in the block before we average that dimension?

guinneak198 (commented):
left a question for discussion

examples/proc_raw.py
jmfranck: Reviews are not appropriate forums for discussion

jmfranck: But I will say that the max method of the ndshape object is new, and you should go check it out, consider adding a PR with doc strings

jmfranck:
we reviewed and merged

PR #97, opened 2024-06-27
load data

guinneak198:
set t axis by subtracting tau, set coherence pathway property, and remove unitary from ph FT


pyspecProcScripts/load_data.py
```diff
@@ -38,8 +38,9 @@ def proc_bruker_deut_IR_withecho_mancyc(s, fl=None):
         fl.next("IR prior to FTing ph")
         fl.image(s.C.setaxis("indirect", "#").set_units("indirect", "scan #"))
     s.ft(
-        ["ph1", "ph2"], unitary=True
+        ["ph1", "ph2"]
     )  # fourier transforming from phase cycle dim to coherence dimension
+    s.set_prop("coherence_pathway", {"ph1": 0, "ph2": -1})
```
jmfranck, line 43: The name implies that it has an echo, so wouldn't Δp₂=+1?  Maybe just leave this out if we're not sure right now.

pyspecProcScripts/load_data.py
```diff
@@ -364,8 +384,13 @@ def proc_spincore_IR(s, fl=None):
     s.setaxis("ph1", r_[0, 2.0] / 4)
     s.setaxis("ph2", r_[0, 2.0] / 4)
     s.reorder(["ph1", "ph2"]).set_units("t2", "s")
+    s.set_units("t2", "s")
+    s.set_prop("coherence_pathway", {"ph1": 0, "ph2": -1})
```
jmfranck, line 388: again, where is this coming from? doesn't this have an echo?

pyspecProcScripts/load_data.py
```diff
@@ -38,8 +38,9 @@ def proc_bruker_deut_IR_withecho_mancyc(s, fl=None):
         fl.next("IR prior to FTing ph")
         fl.image(s.C.setaxis("indirect", "#").set_units("indirect", "scan #"))
     s.ft(
-        ["ph1", "ph2"], unitary=True
+        ["ph1", "ph2"]
```
jmfranck, line 41: can you explain in a comment maybe the first time that you are not doing unitary in order to match the amplitude of the coherence domain signal to its size in each transient

pyspecProcScripts/load_data.py
guinneak198: unsure so left a info(strm

pyspecProcScripts/load_data.py
guinneak198: added in def prc_spincore_SE_v1 (first popularly used processing in the file)

pyspecProcScripts/load_data.py
```diff
@@ -38,31 +38,24 @@ def proc_bruker_deut_IR_withecho_mancyc(s, fl=None):
...
     )  # fourier transforming from phase cycle dim to coherence dimension
+    logging.info(
+        strm(
+            "I didn't know what to set the coherence pathway to so you need to manually do this by setting s.set_prop('coherence_pathway') = {'ph1':x, 'ph2':y}"
```
jmfranck, line 45: this solution is not cool.  You are the one who wrote this module --> do you know what ppg it pertains to?  Why not just paste the main block of the ppg into the docstring of the function, as I ask elsewhere.

pyspecProcScripts/load_data.py
jmfranck: I don't see that here, and this is the first diff in the PR.

pyspecProcScripts/load_data.py
```diff
@@ -38,31 +38,24 @@ def proc_bruker_deut_IR_withecho_mancyc(s, fl=None):
...
-        )
+        fl.image(s_forplot.C.setaxis("indirect", "#").set_units("indirect", "scan #"))
         fl.next("frequency domain (all $\\Delta p$)")
         s_forplot.ft("t2", pad=4096)
-        fl.image(
-            s_forplot.C.setaxis("indirect", "#").set_units(
-                "indirect", "scan #"
-            )
-        )
+        fl.image(s_forplot.C.setaxis("indirect", "#").set_units("indirect", "scan #"))
```
jmfranck, line 58: I ran black on this file inside jf review actual -- you may need to merge to get rid of this black-related diffs.

pyspecProcScripts/load_data.py
```diff
@@ -84,8 +77,13 @@ def proc_bruker_deut_IR_mancyc(s, fl=None):
     # titling to coherence domain
     s.ft("t2", shift=True)  # fourier transform
     s.ft(
-        ["ph1", "ph2"], unitary=True
+        ["ph1", "ph2"]
     )  # fourier transforming from phase cycle dim to coherence dimension
+    logging.info(
+        strm(
+            "I didn't know what to set the coherence pathway to so you need to manually do this by setting s.set_prop('coherence_pathway') = {'ph1':x, 'ph2':y}"
```
jmfranck, line 84: again, not a great solution

pyspecProcScripts/load_data.py
guinneak198: I said it was in the most popularly used ppg but moved it here so you can see it

pyspecProcScripts/load_data.py
guinneak198: where else are you asking for this? I don't see - this is processing for a dataset run on the bruker, using AB's old old data.  I honestly don't know which ppg it pertains to if this was for the og 2H IR or what

pyspecProcScripts/load_data.py
jmfranck: add the main block of the ppg to the docstring of this function

jmfranck: please say here that you are not doing unitary in order to match the amplitude of the coherence domain signal to its size in each transient

pyspecProcScripts/load_data.py
jmfranck: if we don't know which ppg it pertains to, why are we keeping it around?

pyspecProcScripts/load_data.py
```diff
@@ -38,31 +38,25 @@ def proc_bruker_deut_IR_withecho_mancyc(s, fl=None):
...
+        )
+    )
     s.reorder(["indirect", "t2"], first=False)
     if fl is not None:
         s_forplot = s.C
         fl.next("FT")
-        fl.image(
```
jmfranck, line 47: you might need to merge from jf review actual so this doesn't show up as a diff.   jf review actual no longer differs here

I'm stopping reviewing this until this is fixed, because that seems wasteful

pyspecProcScripts/load_data.py
guinneak198: edited according to discussion [here](https://jmfrancklab.slack.com/archives/CTD29CZPE/p1719521574408469)

pyspecProcScripts/load_data.py
guinneak198: commented in spincore SE v1 since that is now the first function in the script

pyspecProcScripts/load_data.py
```diff
@@ -21,291 +21,15 @@
...
-        fl.next("freq domain coh domain")
-        fl.image(s)
-    s.ift("t2")
-    if fl is not None:
-        fl.next("t2 chunked", figsize=(5, 20))
-        fl.image(s)
-    s.ft("t2")
-    return s
-
-
```
jmfranck, line 306: what's the justification for removing these?

pyspecProcScripts/load_data.py
```diff
@@ -343,14 +71,19 @@ def proc_Hahn_echoph(s, fl=None):
...
+    s.squeeze()
     s.reorder(["ph2", "ph1"])
     s.setaxis("nScans", r_[0:nScans])
     s.reorder("t2", first=False)
     s.ft("t2", shift=True)
     if fl is not None:
         fl.next("raw data, chunked")
         fl.image(abs(s))
-    s.ft(["ph1", "ph2"], unitary=True)
+    s.ft(["ph1", "ph2"])
```
jmfranck, line 86: explain the first time around that this will preseve the amplitude between individual transients and the signal pathway in the coherence domain.

pyspecProcScripts/load_data.py
```diff
@@ -462,16 +217,40 @@ def proc_nutation_chunked(s, fl=None):
...
+
+
 def proc_nutation_v4(s, fl=None):
     if s.shape["indirect"] > s.shape["nScans"]:
         s.reorder(["ph1", "nScans", "indirect"])
     else:
         s.reorder(["ph1", "indirect", "nScans"])
     s["indirect"] *= 1e-6  # why is it labeled in μs??
     s.rename("indirect", "p_90")
+    s.set_prop("coherence_pathway", {"ph1": 1})
```
jmfranck, line 247: at some point, don't we start storing this in the file at the time of acquisition? since this is version 4, can you check if the version 4 code is yet doing that?

pyspecProcScripts/load_data.py
jmfranck: I know that for some, you lost track of which bruker ppts they pertained to.  However, I object to throwing out the files that pull the anavpt info, since that is tricky.

These files in the RM2D manuscript repo are doing something similar, and offer insight into which ppgs might be used for the functions that use anavpt:
```
proc_T1CPMG_Igepal.py
proc_T1CPMG_AOT_RM2.py
proc_T1CPMG_AOT_RM1.py
```

jmfranck (changes requested):

jmfranck:
I got rid of  stuff that didn't make sense and just merged

PR #99, opened 2024-07-01
WK add spincore_generalproc_v1 for run_largecyc_CPMG.py

wfkincaid:


pyspecProcScripts/load_data.py
```diff
@@ -629,6 +629,9 @@ def proc_spincore_ODNP_v4(s, fl=None):
         s.ft("t2")
     return s
 
+def proc_spincore_generalproc_v1(s, fl=None):
+    for j in [k for k in s.dimlabels if k.startswith('ph')]:
+        s.ft(j)
```
jmfranck, line 634: typically, I think we also want to:

- FT along `t2`
- `.squeeze()`

jmfranck: but discuss w/ AG

pyspecProcScripts/load_data.py
```diff
@@ -320,13 +320,15 @@ def proc_spincore_diffph_SE_v2(s, fl=None):
...
-    s.ft(["ph2", "ph_diff"])  # if we have used cycles for the axis
+    #s.ft(["ph2", "ph_diff"])  # if we have used cycles for the axis
     #                          coordinates, signal in the coherence
     #                          dimension will match the amplitude of signal
     #                          in a single transient if we do this
     # {{{ after the FT, these have a different meaning in terms of coherence
     #     pathways -- remember that when labeling, pySpecData will change
     #     the ph here to a δp
+    s.ft(['ph2'])
+    s.ft(['ph_diff'])
```
jmfranck, line 331: what is the point of this change?

pyspecProcScripts/load_data.py
guinneak198: for the rest of the preprocessing functions we save the squeeze for after - e.g. we don't include here

pyspecProcScripts/load_data.py
guinneak198: testing things sorry!

guinneak198 (approved):
looks good

PR #100, opened 2024-07-02
script to compare absolute time domain signal of CPMG and echoes

guinneak198:


examples/compare_CPMG_echo.py
```diff
@@ -0,0 +1,44 @@
...
+        thisd.squeeze()
+        fl.next("raw data for %s" % thislabel)
+        rows = np.prod([thisd.shape[j] for j in thisd.dimlabels[:-1]])
+        if rows < 500:
+            fl.image(thisd)
+        else:
+            fl.image(thisd, interpolation="bilinear")
+        thisd.ift("t2")
+        fl.next("abs(t domain) comparison")
+        thisd = select_pathway(thisd, thisd.get_prop("coherence_pathway"))
```
jmfranck, line 38: But you need to grab alternating pathways for alternating echoes for the 4x4x4 cpmg. It's not just one pathway.

If you're using the data with the incorrectly assigned axis, you will also want to divide by 4. 

I would recommend that the easy solution to this is to pass an extra argument in the list of tuples at the top, which is a function that either does what you are doing here, or does what you need to do for the 4x4x4 cpmg

jmfranck (changes requested):

examples/compare_CPMG_echo.py
guinneak198: okay I added odd and even which I think is what we want to do?

examples/compare_CPMG_echo.py
```diff
@@ -0,0 +1,69 @@
...
+        #    False,
+        #    "CPMG simple cyc",
+        #),
+        #(
+        #    "240702_13p5mM_TEMPOL_CPMG.h5",
+        #    "ODNP_NMR_comp/CPMG",
+        #    "CPMG_17",
+        #    True,
+        #    "CPMG large cyc SW = 2.0",
+        #),
```
jmfranck, line 20: do not commit commented out code

jmfranck, line 20: do not commit commented out code

examples/compare_CPMG_echo.py
```diff
@@ -0,0 +1,69 @@
...
+        rows = np.prod([thisd.shape[j] for j in thisd.dimlabels[:-1]])
+        if rows < 500:
+            fl.image(thisd)
+        else:
+            fl.image(thisd, interpolation="bilinear")
+        thisd.ift("t2")
+        fl.next("abs(t domain) comparison")
+        if complex_cpmg:
+            odd = select_pathway(thisd,thisd.get_prop('coherence_pathway'))
+            even_pw = {'ph1':-1,'ph2':-2,'ph_overall':-1}
```
jmfranck, line 57: you need to run black -- I updated the flake command so this will hopefully flag in the future.

examples/compare_CPMG_echo.py
```diff
@@ -0,0 +1,69 @@
...
+        thisd.squeeze()
+        fl.next("raw data for %s" % thislabel)
+        rows = np.prod([thisd.shape[j] for j in thisd.dimlabels[:-1]])
+        if rows < 500:
+            fl.image(thisd)
+        else:
+            fl.image(thisd, interpolation="bilinear")
+        thisd.ift("t2")
+        fl.next("abs(t domain) comparison")
+        if complex_cpmg:
```
jmfranck, line 55: can we call this `manystep_cpmg` instead? I just don't like "complex" b/c of double meaning

examples/compare_CPMG_echo.py
```diff
@@ -0,0 +1,69 @@
...
+        thisd = find_file(
+            thisfile, exp_type=exp_type, expno=nodename, lookup=lookup_table
+        )
+        thisd.squeeze()
+        fl.next("raw data for %s" % thislabel)
+        rows = np.prod([thisd.shape[j] for j in thisd.dimlabels[:-1]])
+        if rows < 500:
+            fl.image(thisd)
+        else:
+            fl.image(thisd, interpolation="bilinear")
```
jmfranck, line 52: I just updated pyspecdata -- it should be possible to replace all of this with `fl.image(thisd, interpolation='auto')`

examples/compare_CPMG_echo.py
jmfranck: it gives the same result, but for clarity, ph2 should be +2.

examples/compare_CPMG_echo.py
```diff
@@ -0,0 +1,69 @@
...
+            even_pw = {'ph1':-1,'ph2':-2,'ph_overall':-1}
+            even = select_pathway(thisd.C,even_pw)
+            thisd = odd + even
+        else:
+            thisd = select_pathway(thisd, thisd.get_prop("coherence_pathway"))
+        if "nEcho" in thisd.dimlabels:
+            thisd.smoosh(["nEcho", "t2"], "t2")
+            acq = thisd.get_prop("acq_params")
+            echo_time = 1e-6 * 2 * (acq["tau_us"] + acq["p90_us"])
+            thisd["t2"] = (thisd["t2"]["nEcho"]) * echo_time + thisd["t2"]["t2"]
```
jmfranck, line 66: if I understand what you're doing, I think you want `p180_us` divided by two (which should be negligibly different than what you have here, but maybe more legible)

jmfranck (changes requested):
this was good, thanks!   comments should be easy

jmfranck:
what this shows for the CPMG is very confusing -- the first step to diagnose would be to see how much signal is in all the pathways:
```
# do this after squashing t2 and echoes:
s_cohsize = abs(s).mean_all_but(['t2']+[j for j in s.dimlabels if j.startswith('ph')])
```
and also select out just the two pathways and see what they look like:
```
s['ph_overall',-1]['ph2',2]['ph1',r_[1,-1]].reorder('nEchoes',first=False) # put echoes on the inside so they show up along x
```

Did you really run with only 4 echoes?  That's all that was allowed b/c of the phase cycling?

examples/compare_CPMG_echo.py
```diff
@@ -141,18 +141,21 @@
...
         thisd.squeeze()
         fl.next("raw data for %s" % thislabel)
-        rows = np.prod([thisd.shape[j] for j in thisd.dimlabels[:-1]])
-        if rows < 500:
-            fl.image(thisd)
-        else:
-            fl.image(thisd, interpolation="bilinear")
+        fl.image(thisd, interpolation='auto')
+        fl.next("raw data for %s, slice overall and signal average" % thislabel)
+        fl.image(thisd['ph_overall',-1].sum('nScans').ift('t2'), interpolation='auto')
```
jmfranck, line 146: since `ph_overall` will always be -1

examples/compare_CPMG_echo.py
```diff
@@ -141,18 +141,21 @@
         )
         thisd.squeeze()
         fl.next("raw data for %s" % thislabel)
-        rows = np.prod([thisd.shape[j] for j in thisd.dimlabels[:-1]])
-        if rows < 500:
-            fl.image(thisd)
-        else:
-            fl.image(thisd, interpolation="bilinear")
+        fl.image(thisd, interpolation='auto')
```
jmfranck, line 144: implemented my earlier comment

examples/compare_CPMG_echo.py
```diff
@@ -143,7 +143,10 @@
         fl.next("raw data for %s" % thislabel)
         fl.image(thisd, interpolation='auto')
         fl.next("raw data for %s, slice overall and signal average" % thislabel)
-        fl.image(thisd['ph_overall',-1].sum('nScans').ift('t2'), interpolation='auto')
+        forplot = thisd['ph_overall',-1].sum('nScans').ift('t2')
+        if 'nEcho' in forplot.dimlabels:
+            forplot.reorder('nEcho', first=False)
+        fl.image(forplot)
```
jmfranck, line 149: this plot allows me to see where the artefacts are.

examples/compare_CPMG_echo.py
wfkincaid: I'm a little confused, there is no `p180_us` parameter to be pulled from the ini

jmfranck: that's weird -- I think that means that the ppg is not storing it but should be.

pyspecProcScripts/load_data.py
```diff
@@ -311,18 +311,13 @@ def proc_spincore_diffph_SE_v2(s, fl=None):
     r"""this one uses a phase cycle where the overall phase and 90-180
     phase difference are cycled in a nested way -- see the DCCT paper to
     understand this!"""
-    s.ft(["ph2", "ph_diff"])  # if we have used cycles for the axis
-    #                          coordinates, signal in the coherence
-    #                          dimension will match the amplitude of signal
-    #                          in a single transient if we do this
+    s = proc_spincore_generalproc_v1(s, fl=fl)
```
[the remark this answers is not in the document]
jmfranck: resolve when read

examples/compare_CPMG_echo.py
```diff
@@ -141,18 +141,21 @@
...
+            odd_pw = {'ph1':1,'ph2':-2,'ph_overall':-1}
+            odd = select_pathway(thisd,odd_pw)
+            even_pw = {'ph1':-1,'ph2':+2,'ph_overall':-1}
             even = select_pathway(thisd.C,even_pw)
-            thisd = odd + even
+            # {{{ interleave the echoes
+            thisd = odd
+            # "even" starts w/ 1 b/c that's the second echo
+            thisd['nEcho',1::2] = even['nEcho',1::2]
+            # }}}
```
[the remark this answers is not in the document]
jmfranck: you can resolve this comment once you read it

examples/compare_CPMG_echo.py
jmfranck: I actually ended  up undoing this, b/c I wanted to see if there were artefacts for other ph_overall 

you can resolve this comment once you read it

examples/compare_CPMG_echo.py
jmfranck: you can resolve this comment once you read it

examples/compare_CPMG_echo.py
jmfranck: you can resolve this comment once you read it

jmfranck:
note that I deliberately didn't run black on the whole file -- leading to these flake errors.  That was to keep the diffs minimal so you could see what I'm doing.

Obviously, you'll want to run black when getting it ready to merge

examples/compare_CPMG_echo.py
```diff
@@ -0,0 +1,64 @@
...
+            "echo large cyc SW = 10.0",
+        ),
+    ]:
+        thisd = psd.find_file(
+            thisfile, exp_type=exp_type, expno=nodename, lookup=lookup_table
+        )
+        forplot = thisd.C.sum("nScans").ift("t2").run(abs)
+        thisd.squeeze()
+        fl.next("raw data for %s" % thislabel)
+        fl.image(thisd, interpolation="none")
```
jmfranck, line 46: note that one of these plots doesn't show up properly.  you need to use the new `"auto"` interpolation

examples/compare_CPMG_echo.py
```diff
@@ -0,0 +1,68 @@
...
+            "no power",
+        ),
+        (
+            "240702_13p5mM_TEMPOL_echo.h5",
+            "ODNP_NMR_comp/Echoes",
+            "echo_12",
+            False,
+            "echo large cyc SW = 10.0",
+            "no power",
+        ),
```
jmfranck, line 41: can you add all the powers that you acquired?  I'm assuming you can diff against the earlier version where they were commented out.

PR #101, opened 2024-07-03
WK adding in proc script from JF notebook to look at T2 diffs with mu…

wfkincaid:
…ltiple echoes

### DIFF
--- a/examples/proc_cpmg_gen_echo_extensive.py
+++ b/examples/proc_cpmg_gen_echo_extensive.py
@@ -1,0 +1,123 @@
+import numpy as np
+import pyspecdata as psd
+import matplotlib.pylab as plt
+from itertools import cycle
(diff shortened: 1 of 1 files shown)


jmfranck:
You have nicely identified a difference where the more rigorously cycled CPMG [here](https://jmfrancklab.slack.com/files/U02QBNNKX1N/F07B639QPV2/image.png?origin_team=TK0FA3GM7&origin_channel=CTD29CZPE) is different from the CP-less CPMG.

As discussed [here](https://jmfrancklab.slack.com/archives/CTD29CZPE/p1720119569878569), I think that you can pull code from my most recent notebook to show exactly *why*. 

Certainly, that should go in your notebook, but it's not clear to me yet if we want to make that into an example or not, so we can leave any new PR generation for the discussion.

PR #104, opened 2024-07-16
some issue with setting wrong units in f domain

guinneak198:
units set wrong


jmfranck (changes requested):
Just set the units to Hertz

PR #108, opened 2024-07-23
processing to calculate sqrt(P) of pulse capture

guinneak198:


examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,121 @@
...
+Assuming the data is the capture of the pulse sequence as seen on the GDS
+oscilloscope, here the data is converted to analytic power, frequency filtered
+and the absolute is taken prior to integrating to return the beta value (us
+sqrt(W)). 
+"""
+import pyspecdata as psd
+import matplotlib.pyplot as plt
+from numpy import sqrt, array
+
+show_all = False
```
guinneak198, line 13: diagnostic option

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,121 @@
...
+with psd.figlist_var() as fl:
+    for filename, nodename, beta90_range, beta180_range in [
+        (
+            "240729_test_p90_pulse_calib.h5",
+            "pulse_calib_2",
+            (None, 34e-6),
+            (34e-6, None),
```
guinneak198, line 20: need to select wide range so that this works for shorter and longer pulses. here tau = 20 us

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,121 @@
...
+        ),
+    ]:
+        d = psd.nddata_hdf5(
+            filename + f"/{nodename}",
+            directory=psd.getDATADIR(exp_type="ODNP_NMR_comp/test_equipment"),
+        )
+        beta_1 = []
+        beta_2 = []
```
guinneak198, line 28: initiate list for the beta values of the 90 pulse and the 180 pulse

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,121 @@
...
+                fig.suptitle(fl.basename)
+                # {{{ Plot raw capture
+                ax_list[1].set_ylabel(r"$\sqrt{P_{pulse}}$")
+                ax_list[0].set_ylabel(r"$\sqrt{P_{pulse}}$")
+                ax_list[1].set_ylabel(None)
+                ax_list[1].set_title("180 Pulse")
+                fl.next("Pulse Calibration")
+                fl.plot(s["t":beta90_range], alpha=0.2, color="blue", ax=ax_list[0])
+                fl.plot(s["t":beta180_range], alpha=0.2, color="blue", ax=ax_list[1])
+                # }}}
```
guinneak198, line 49: again diagnostic with the option set to False as I assume we don't want to see all 50 pulse settings

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,121 @@
...
+            if show_all:
+                # {{{ plot and integrate 90 pulse
+                fl.plot(abs(s["t":beta90_range]), ax=ax_list[0], color="red")
+                fl.plot(abs(s["t":beta180_range]), ax=ax_list[1], color="red")
+                # }}}
+            beta90_int = abs(s["t":beta90_range]).contiguous(
+                lambda x: x > 0.01 * s["t":beta90_range].max()
+            )[0]
+            beta180_int = abs(s["t":beta180_range]).contiguous(
+                lambda x: x > 0.01 * s["t":beta180_range].max()
```
guinneak198, line 77: select integration range based on lambda function (anythin higher than 0.01 of the max in the selected region). Selected region needed to separate the 90 pulse from the 180 pulse

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,120 @@
...
+            "240729_test_p90_pulse_calib.h5",
+            "pulse_calib_2",
+            (None, 34e-6),
+            (34e-6, None),
+        ),
+    ]:
+        d = psd.nddata_hdf5(
+            filename + f"/{nodename}",
+            directory=psd.getDATADIR(exp_type="ODNP_NMR_comp/test_equipment"),
+        )
```
jmfranck, line 26: Use find file instead

examples/proc_GDS_pulse_capture.py
jmfranck: This doesn't make sense, and it is over complicated because this should be for only one pulse, as we have discussed

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,120 @@
...
+        beta_1 = []
+        beta_2 = []
+        for j in range(len(d["p_90"])):
+            s = d["p_90", j].C
+            s.set_units("t", "s")
+            s *= 101.35  # attenutation ratio
```
jmfranck, line 32: This should be set at the top of the script

jmfranck: Or in your list of tuples

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,120 @@
...
+                    s.get_prop("set_beta")[j]
+                )
+                fig, ax_list = plt.subplots(1, 2, figsize=(10, 4))
+                fig.suptitle(fl.basename)
+                # {{{ Plot raw capture
+                ax_list[1].set_ylabel(r"$\sqrt{P_{pulse}}$")
+                ax_list[0].set_ylabel(r"$\sqrt{P_{pulse}}$")
+                ax_list[1].set_title("180 Pulse")
+                fl.next("Pulse Calibration")
+                fl.plot(s["t":beta90_range], alpha=0.2, color="blue", ax=ax_list[0])
```
jmfranck, line 46: By only doing one pulse, there is no need for this manual range

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,120 @@
...
+            s.ift("t")
+            if show_all:
+                # {{{ plot and integrate 90 pulse
+                fl.plot(abs(s["t":beta90_range]), ax=ax_list[0], color="red")
+                fl.plot(abs(s["t":beta180_range]), ax=ax_list[1], color="red")
+                # }}}
+            beta90_int = abs(s["t":beta90_range]).contiguous(
+                lambda x: x > 0.01 * s["t":beta90_range].max()
+            )[0]
+            beta180_int = abs(s["t":beta180_range]).contiguous(
```
jmfranck, line 75: This is good, but after you do this, you should expand the range slightly so that you capture the edges of the pulse

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,120 @@
...
+                )
+                ax_list[1].axvline(beta180_int[0] * 1e6, ls=":", alpha=0.2)
+                ax_list[1].axvline(beta180_int[1] * 1e6, ls=":", alpha=0.2)
+                # }}}
+        # {{{ make nddata for beta of 90 pulse and 180 pulse
+        if show_all:
+            fl.basename = None
+        fig2, axs = plt.subplots(1, 2, figsize=(10, 4))
+        fig2.suptitle(r"Measured $\beta$ vs $t_{pulse}$")
+        fl.next(r"Measured $\beta$ vs $t_{pulse}$")
```
jmfranck, line 108: Yes, this is what you want

examples/proc_GDS_pulse_capture.py
jmfranck: Also, I do not understand why there is a list here the best way to do that is to pull the ND shape, and then pop the direct dimension.

Also, it occurs to me that you can use the same code both to get the data that you need for your calibration, as well as to verify the calibration once it's done. In the first case, your indirect dimension would be the programmed pulse length, and in the second case, your indirect dimension would be the programmed beta value. The strategy I describe above enables that with the same code.

jmfranck (changes requested):

examples/proc_GDS_pulse_capture.py
jmfranck: (Play with an example in I python to see what I'm talking about0)

examples/proc_GDS_pulse_capture.py
jmfranck: again, don't put things in a PR which belong as code comments.  If you feel that you need a PR comment to explain the code, you're probably missing a code comment

examples/proc_GDS_pulse_capture.py
jmfranck: same as above + see comment about expanding

jmfranck: (although this what does "based on lambda function" mean? You're just using the `contiguous` function as it's intended to be used)

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+show_all = False
+with psd.figlist_var() as fl:
+    for filename, nodename, label in [
+        ("240730_test_amp1_fin_pulse_calib.h5", "pulse_calib_1", "amplitude = 1"),
+        ("240730_test_amp0p1_fin_pulse_calib.h5", "pulse_calib_1", "amplitude = 0.1"),
+    ]:
+        d = psd.find_file(
+            filename, expno=nodename, exp_type="ODNP_NMR_comp/test_equipment"
+        )
+        beta = psd.ndshape(d).pop("t").alloc().rename("p_90", "prog_beta")
```
jmfranck, line 23: no -- you are not programming β.  You set/program `t_p` or `t_pulse` as well as `amplitude` and you measure the β value that you got for those settings!

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
+"""
+Calculate actual beta as a function of pulse length
+===================================================
+Assuming the data is the capture of the pulse sequence as seen on the GDS
+oscilloscope, here the data is converted to analytic power, frequency filtered
+and the absolute is taken prior to integrating to return the beta value (us
+sqrt(W)).
```
jmfranck, line 7:
```
:math:`\beta = \int \sqrt{P(t)} dt`
```

examples/proc_GDS_pulse_capture.py
jmfranck: Again, consider code comment!!

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+show_all = False
+with psd.figlist_var() as fl:
+    for filename, nodename, label in [
+        ("240730_test_amp1_fin_pulse_calib.h5", "pulse_calib_1", "amplitude = 1"),
+        ("240730_test_amp0p1_fin_pulse_calib.h5", "pulse_calib_1", "amplitude = 0.1"),
+    ]:
+        d = psd.find_file(
+            filename, expno=nodename, exp_type="ODNP_NMR_comp/test_equipment"
+        )
+        beta = psd.ndshape(d).pop("t").alloc().rename("p_90", "desired_beta")
```
jmfranck, line 23: No, the axis coordinates are t_pulse!

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+import pyspecdata as psd
+import matplotlib.pyplot as plt
+from numpy import sqrt
+
+atten_ratio = 101.35  # attenutation ratio
+show_all = False
+with psd.figlist_var() as fl:
+    for filename, nodename, label in [
+        ("240730_test_amp1_fin_pulse_calib.h5", "pulse_calib_1", "amplitude = 1"),
+        ("240730_test_amp0p1_fin_pulse_calib.h5", "pulse_calib_1", "amplitude = 0.1"),
```
jmfranck, line 18: You need a comment explaining what program was used to acquire this data

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+    ]:
+        d = psd.find_file(
+            filename, expno=nodename, exp_type="ODNP_NMR_comp/test_equipment"
+        )
+        beta = psd.ndshape(d).pop("t").alloc().rename("p_90", "desired_beta")
+        beta.setaxis("desired_beta", d.get_prop("desired_betas"))
+        t_v_beta = psd.ndshape(d).pop("t").alloc().rename("p_90", "At_p")
```
jmfranck, line 25: I don't understand what this is supposed to be, and I don't think that it's necessary

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+            list(
+                d.get_prop("acq_params")["amplitude"] * d.get_prop("set_p90s")[j]
+                for j in range(len(d["p_90"]))
+            ),
+        )
+        for j in range(len(d["p_90"])):
+            s = d["p_90", j].C
+            s.set_units("t", "s")
+            s *= atten_ratio
+            s /= sqrt(2)  # Vrms
```
jmfranck, line 37: I'm not sure if we want this, but we can figure that out based on whether or not the amplitude looks correct

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+            s /= sqrt(50)  # V/sqrt(R) = sqrt(P)
+            if show_all:
+                fl.basename = r"programmed $\beta$ = %s" % str(
```
jmfranck, line 40: There is no such thing as a programmed beta for this program

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+                    s.get_prop("set_betas")[j]
```
jmfranck, line 41: There is also no such thing as a set beta.. there are only the pulse links that you program and the values of beta that you observe

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+            # {{{ frequency filter
+            s["t":(0, 11e6)] *= 0
+            s["t":(24e6, None)] *= 0
+            # }}}
+            s.ift("t")
+            if show_all:
+                # {{{ plot and integrate 90 pulse
+                fl.plot(abs(s), color="red")
+                # }}}
+            beta90_int = abs(s).contiguous(lambda x: x > 0.01 * s.max())[0]
```
jmfranck, line 68: You need a better variable name here... This is just the range of times over which you want to integrate the pulse

jmfranck: Also, as previously stated, you typically want to expand this a little bit so that you actually integrate over the whole pulse

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+                s.ft("t")
+                # }}}
+            # {{{ frequency filter
+            s["t":(0, 11e6)] *= 0
+            s["t":(24e6, None)] *= 0
+            # }}}
+            s.ift("t")
+            if show_all:
+                # {{{ plot and integrate 90 pulse
+                fl.plot(abs(s), color="red")
```
jmfranck, line 66: I would move this down, and only show the analytic signal over the integration range that you determine in the next step

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+                # }}}
+            beta90_int = abs(s).contiguous(lambda x: x > 0.01 * s.max())[0]
+            beta1 = abs(s["t":beta90_int]).integrate("t").data.item() * 1e6
+            beta["desired_beta", j] = beta1
```
jmfranck, line 70: I think that if you just replace this desired beta string with t_ pulse this would make sense

jmfranck: Just search and replace desired_beta with t_ pulse before doing anything else

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+            t_v_beta["At_p", j] = beta1
```
jmfranck, line 71: This is not useful for anything

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+            if show_all:
+                plt.ylabel(r"$\sqrt{P_{pulse}}$")
+                plt.text(
+                    beta90_int[0] * 1e6 - 1,
+                    -1,
+                    r"$t_{90} \sqrt{P_{tx}} = %f s \sqrt{W}$" % beta1,
+                )
+                plt.axvline(beta90_int[0] * 1e6, ls=":", alpha=0.2)
```
jmfranck, line 79: See my suggestion above, which is simpler

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+                plt.axvline(beta90_int[1] * 1e6, ls=":", alpha=0.2)
+        # {{{ make nddata for beta of 90 pulse and 180 pulse
+        fl.next(r"Measured $\beta$ vs programmed $\beta$")
```
jmfranck, line 82: It's measured beta versus programmed pulse length

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+"""
+import pyspecdata as psd
+import matplotlib.pyplot as plt
+from numpy import sqrt
+
+atten_ratio = 101.35  # attenutation ratio
+show_all = False
+with psd.figlist_var() as fl:
+    for filename, nodename, label in [
+        ("240730_test_amp1_fin_pulse_calib.h5", "pulse_calib_1", "amplitude = 1"),
```
jmfranck, line 17: The amplitude parameter here should be a number, since you're going to want to use that later

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,91 @@
...
+                    -1,
+                    r"$t_{90} \sqrt{P_{tx}} = %f s \sqrt{W}$" % beta1,
+                )
+                plt.axvline(beta90_int[0] * 1e6, ls=":", alpha=0.2)
+                plt.axvline(beta90_int[1] * 1e6, ls=":", alpha=0.2)
+        # {{{ make nddata for beta of 90 pulse and 180 pulse
+        fl.next(r"Measured $\beta$ vs programmed $\beta$")
+        fl.plot(beta, "o", label=label)
+        psd.gridandtick(plt.gca())
+        plt.ylabel(r"measured $\beta$ / $\mathrm{\mu s \sqrt{W}}$")
```
jmfranck, line 85: I'm going to ignore from here to the end because I would end up repeating a lot of the comments from above. 

Here I understand what you're trying to do with this extra ND data that you've defined, but I still point out that it's not necessary.

All that you're trying to do, which makes sense, is to multiply your amplitude by your pulse length, so that you can put both the low power and the high power data on the same plot.

jmfranck: I would point out that this plot that you are generating is not exactly any of the three plots that are part of our main goal right now. That is okay. 

It is the first plot turned on its side (with x and y interchanged).

I still hold that you want to add to the end of this a section where you flip the axis coordinates and the values of the data, and then determine your polynomial coefficients and your linear threshold in this code, so that you do not need to copy and paste entire calibration tables into the function that does the interpolation for you.

This is tied into a discussion that is on slack, or in the pr for the acquisition, or both.

jmfranck (changes requested):

examples/proc_GDS_pulse_capture.py
guinneak198: over phone we discussed showing plots for both t_v_beta and beta_v_t

jmfranck: No, we didn't. T_v_beta as used and defined here is very different than what we discussed over the phone, which is simply flipping the coordinate axis and the data at the end

examples/proc_GDS_pulse_capture.py
```diff
@@ -12,147 +12,160 @@
...
+        d.set_units("t", "s")  # why isn't this done already??
+        d.rename(
+            "p_90", "t_pulse"
+        )  # just b/c existing name doesn't make sense
+        d["t_pulse"] = np.float64(
+            d["t_pulse"]
+        )  # PR COMMENT: these seem to be integers.  Since the pulse lengths should have been created with linspace, that is super super suspect
+        d *= atten_ratio
+        d /= np.sqrt(2)  # Vrms
+        d /= np.sqrt(50)  # V/sqrt(R) = sqrt(P)
```
jmfranck, line 40: the only part that needs to be looped is the part where you determine the slices.

examples/proc_GDS_pulse_capture.py
```diff
@@ -12,147 +12,160 @@
...
-                for j in range(len(d["p_90"]))
-            ),
-        )
+        d.set_units("t", "s")  # why isn't this done already??
+        d.rename(
+            "p_90", "t_pulse"
+        )  # just b/c existing name doesn't make sense
+        d["t_pulse"] = np.float64(
+            d["t_pulse"]
+        )  # PR COMMENT: these seem to be integers.  Since the pulse lengths should have been created with linspace, that is super super suspect
```
jmfranck, line 37: see and remove PR comments after dealing with them.

The data needs to be reacquired b/c:

- why is this axis labeled with integers? Very suspicious.
- The raw data has a very stepped characteristic -- it looks like you are zoomed way out on the y-axis.
- I would just change the name of the indirect dimension to t_pulse so we don't have to do this.
- Is there an open PR for saving the units to the file?

examples/proc_GDS_pulse_capture.py
```diff
@@ -12,147 +12,160 @@
...
+                if j % skip_plots == 0:
+                    switch_to_plot(d, j)
+                    fl.plot(
+                        d["t_pulse", j],
+                        alpha=0.2,
+                        color=thiscolor,
+                        label=thislabel,
+                    )
+
+        indiv_plots(d, "raw", "blue")
```
jmfranck, line 59: since this is repeated over an over, should be made into a function.

examples/proc_GDS_pulse_capture.py
```diff
@@ -12,147 +12,160 @@
...
             int_range[-1] += 1e-6
-            beta1 = abs(s["t":int_range]).integrate("t").data.item() * 1e6
-            beta_v_t["At_p", j] = beta1
-            if show_all:
-                # {{{ plot and integrate 90 pulse
-                fl.plot(abs(s["t":int_range]), color="red")
-                # }}}
+            beta["t_pulse", j] = (
+                abs(s["t":int_range]).integrate("t").data.item() * 1e6
+            )
```
jmfranck, line 87: note that this does not make sense, which again makes me suspicious that your axis labels are incorrect.

jmfranck (changes requested):

examples/proc_GDS_pulse_capture.py
guinneak198: realized this just now in processing and am fixing

examples/proc_GDS_pulse_capture.py
```diff
@@ -0,0 +1,171 @@
...
+            # {{{ t vs beta
+            if amplitude > 1:
+                linear_threshold = 30
+                curve_threshold = 35
+            else:
+                linear_threshold = 10
+                curve_threshol = 11
+            t_v_beta = beta.shape.alloc().rename("At_p", "beta")
+            t_v_beta.setaxis("beta", beta.data)
+            t_v_beta.data[:] = beta.getaxis("At_p")
```
jmfranck, line 148: this is the code to flip the function -- this should go up, before all the fitting stuff.  You want to fit functions for t_pulse(β).  Also feed `alloc(` `dtype=np.float64` as I do above (so that it's real and not complex)

jmfranck: you should just be able to do something like  this:
```
c_nonlinear = t_v_beta.polyfit('t_pulse', order=10)
c_linear = t_v_beta['t_pulse':range_over_which_its_linear].polyfit('t_pulse',order=1)
```
that's it.

jmfranck: As noted in the other PR, the `prog_plen` function is messed up in your PR.  I would not worry about that right now.   Just work it out here.

jmfranck: Specifically, use this here:

```
def prog_plen(desired):
    def zonefit(desired):
        if desired > linear_threshold:
            return np.polyval(c_linear[::-1], desired)
        else:
            return np.polyval(c_nonlinear[::-1], desired)
    ret_val = np.vectorize(zonefit)(desired)
    if ret_val.size > 1:
        return ret_val
    else:
        return ret_val.item()
```
if you show here that it works (and show that the linear extension beyond the data works), then that basically gives you the code for the other PR!  You just need to hard-code in the linear_threshold and the coefficients for each power.

jmfranck: I actually think that my comments above, and the lines that I'm highlighting, should form the basis of what you need to finish off this script.  I don't understand what all the rest of the lines of code in this `codehasbeenreviewed=False` block are for.

examples/proc_GDS_pulse_capture.py
guinneak198: no longer applicable so resolving

jmfranck: do not resolve comments please!

examples/proc_GDS_pulse_capture.py
guinneak198: discussed [here](https://jmfrancklab.slack.com/archives/C06LNFNP2HM/p1722446119339489)

examples/proc_GDS_pulse_capture.py
```diff
@@ -21,15 +21,17 @@
 skip_plots = 10  # diagnostic -- set this to None, and there will be no plots
 with psd.figlist_var() as fl:
     for filename, nodename, amplitude in [
-            #("240730_test_amp1_fin_pulse_calib.h5", "pulse_calib_1", 1.0),
-        ("240731_amp0p1_calib_fin_pulse_calib.h5", "pulse_calib_1", 0.1),
+        ("240730_test_amp1_fin_pulse_calib.h5", "pulse_calib_1", 1.0),
+        # PR COMMENT: the other set of data was too zoomed out along the x axis, and was being aliased
```
jmfranck, line 25: please read PR comments in code

examples/proc_GDS_pulse_capture.py
```diff
@@ -115,50 +117,18 @@ def indiv_plots(d, thislabel, thiscolor):
...
-            )
-            fl.plot(fit, color=thiscolor, ls=":", alpha=0.5)
-            psd.gridandtick(plt.gca())
-            plt.ylabel(r"measured $\beta$ / $\mathrm{\mu s \sqrt{W}}$")
-            plt.xlabel(r"Amplitude*$t_{pulse}$ / $\mu$s")
-            # }}}
-            # {{{ t vs beta
-            fl.next(r"Amplitude*$t_{pulse}$ vs Measured $\beta$")
-            fl.plot(t_v_beta, "o", color=thiscolor, label=thislabel)
-            if amplitude > 1:
```
jmfranck: do not resuscitate this stuff that I'm deleting.  all the info you need is in the code that I added and in my PR comments.  This stuff is going in a bad direction.

examples/proc_GDS_pulse_capture.py
guinneak198: I think we are on the same page here?

examples/proc_GDS_pulse_capture.py
guinneak198: This has since been addressed

examples/proc_GDS_pulse_capture.py
guinneak198: not sure if still applicable so ignoring

examples/proc_GDS_pulse_capture.py
guinneak198: PR [here](https://github.com/jmfrancklab/pyspecdata/pull/50)

examples/proc_GDS_pulse_capture.py
guinneak198: fixed

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,212 @@
...
+with psd.figlist_var() as fl:
+    for filename, nodename in [
+        ("240805_calib_amp1_pulse_calib.h5", "pulse_calib_1"),  # high power
+        ("240805_calib_amp0p1_a_pulse_calib.h5", "pulse_calib_3"),  # low power
+        ("240805_calib_amp0p2_a_pulse_calib.h5", "pulse_calib_1"),  # low power
+        ("240805_calib_amp0p05_pulse_calib.h5", "pulse_calib_5"),  # low power
+    ]:
+        d = psd.find_file(
+            filename, expno=nodename, exp_type="ODNP_NMR_comp/test_equipment"
+        )
```
jmfranck, line 53: the ppg doesn't have a postproc type.  You should set a postproc_type and add an assert statement here

you don't have to do anything, but want to make sure you understand the data that you're reading

jmfranck: for existing dataset, you can hack in the postproc type (there is an example of how to do this or something similar for the coherence pathway on slack)

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,212 @@
...
+        indiv_plots(d, "raw", "blue")
+        # {{{ data is already analytic, and downsampled to below 24 MHz
+        indiv_plots(abs(d), "analytic", "orange")
+        d.ft("t")
+        # {{{ apply frequency filter
+        d.ift("t")
+        dt = d["t"][1] - d["t"][0]
+        SW = 1 / dt
+        carrier = d.get_prop("acq_params")["carrierFreq_MHz"] * 1e6
+        if int(carrier / SW) % 2 == 0:
```
jmfranck, line 90: this doesn't make sense to me -- you  at least need a comment, but I think it's incorrect

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,212 @@
...
+            center = carrier % SW
```
jmfranck, line 91: the math for this and the next are not quite right.  You are not accounting for the fact that it's shifted, so it runs from -SW/2 to SW/2.  I think you wan to first add SW/2, then modulo with SW, then subtract back out, or something like that, but it should be possible to do this with one expression

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,212 @@
...
+        else:
+            center = SW - (carrier % SW)
+        d.ft("t")
+        left = center - 1e6
+        right = center + 1e6
+        d["t":(0, left)] *= 0
+        d["t":(right, None)] *= 0
```
jmfranck, line 98: please just combine these into two lines

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,212 @@
...
+                    )
+
+        indiv_plots(d, "raw", "blue")
+        # {{{ data is already analytic, and downsampled to below 24 MHz
+        indiv_plots(abs(d), "analytic", "orange")
+        d.ft("t")
+        # {{{ apply frequency filter
+        d.ift("t")
+        dt = d["t"][1] - d["t"][0]
+        SW = 1 / dt
```
jmfranck, line 88: please combine into two lines

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,212 @@
...
+            center = carrier % SW
+        else:
+            center = SW - (carrier % SW)
+        d.ft("t")
+        left = center - 1e6
+        right = center + 1e6
+        d["t":(0, left)] *= 0
+        d["t":(right, None)] *= 0
+        plt.axvline(left * 1e-6)
+        plt.axvline(right * 1e-6)
```
jmfranck, line 100: there is no need for this -- it should be obvious by looking in the frequency domain

jmfranck (changes requested):
I only reviewed up to my last comment.  Please fix and drop a comment when it's ready for review again

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,212 @@
...
+
+def prog_plen(desired):
+    """function that takes the coefficients of the linear and nonlinear
+    regions and applies the fit respectively to calculate the pulse time that
+    will return the desired beta value
+    """
+
+    def zonefit(desired):
+        if desired > linear_threshold:
+            return np.polyval(c_linear[::-1], desired)
```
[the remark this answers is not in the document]
guinneak198: moved back into script where c_linear was defined and I thought I had tested and passed but makes sense to put after c_linear is defined anyway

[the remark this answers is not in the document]
jmfranck: I don't think this will work as-is.

examples/calibrate_tsqrtP.py
guinneak198: took inspiration from [here](https://www.ni.com/en/shop/data-acquisition/measurement-fundamentals/analog-fundamentals/acquiring-an-analog-signal--bandwidth--nyquist-sampling-theorem-.html) and added comments

examples/calibrate_tsqrtP.py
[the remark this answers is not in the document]
guinneak198: I've tested and it runs

examples/calibrate_tsqrtP.py
guinneak198: now updated pulling equation directly from cavanagh

jmfranck: Please don't waste time by ignoring the code and comments that I wrote:

If your carrier is slightly greater than your spectrum width, this will not trigger the if statement. But more than that, this is not taking into account the fact that your FT is shifted. We are not starting from zero. We are starting from SW/2. 

Please roll back to the version that I reviewed, test that code in IPython, and debug that code if need it.

examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,103 @@
...
+        for_L *= L
+        for_L /= s_max
+        # }}}
+        # {{{ heaviside hat functions
+        # {{{ determine center frequency
+        s.ift("t")
+        dt = s["t"][1] - s["t"][0]
+        SW = 1 / dt
+        fn = SW / 2
+        m = np.round(carrier / SW)
```
jmfranck, line 67: This is very similar to what I had written, but you need to think it through and add a comment to explain it.

I am fairly sure that this should be floor instead of round.

Seriously consider going back to my code, which was written in a way that is easier to understand.

jmfranck (changes requested):

examples/calibrate_tsqrtP.py
guinneak198: rolled back and debugged

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,215 @@
...
+        d.ift("t")
+        SW = 1 / (d["t"][1] - d["t"][0])  # sample rate
+        carrier = (
+            d.get_prop("acq_params")["carrierFreq_MHz"] * 1e6
+        )  # signal frequency
+        n = np.floor(
+            (carrier + SW / 2) / SW
+        )  # how far is the carrier from the left side of the spectrum (which is at SW/2), in integral multiples of SW
+        nu_a = (
+            -SW / 2 + (carrier + SW / 2) - n * SW
```
jmfranck, line 80: The SW/2 here cancel, and it makes equal sense without (you are just subtracting an integral SW, which is fine)

jmfranck (changes requested):

pyspecProcScripts/load_data.py
```diff
@@ -838,6 +838,7 @@ def proc_field_sweep_v2(s):
     "ag_CPMG_strob": proc_bruker_CPMG_v1,
     "ag_T1CPMG_2h": proc_bruker_T1CPMG_v1,
     "chirp": proc_capture,
+    "GDS_capture_v1": proc_capture,
```
jmfranck, line 841: I don't think that you want this b/c you don't want to do anything to it, and don't need to pass the lookup table to find file.

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,216 @@
...
+        # ("240805_calib_amp0p1_a_pulse_calib.h5", "pulse_calib_3"),  # low power
+        # ("240805_calib_amp0p2_a_pulse_calib.h5", "pulse_calib_1"),  # low power
+        ("240805_calib_amp0p05_pulse_calib.h5", "pulse_calib_5"),  # low power
+    ]:
+        d = psd.find_file(
+            filename, expno=nodename, exp_type="ODNP_NMR_comp/test_equipment"
+        )
+        assert (
+            d.get_prop("postproc_type") is not None
+        ), "No postproc type was set upon acquisition"
```
jmfranck, line 38: no, assert that the postproc type is the right one

pyspecProcScripts/load_data.py
jmfranck: the way that you have the code written above (which is correct) this will never get used

jmfranck (changes requested):
some changes to make.  Also, can you please separate the PR so that we are doing one file at a time.  I need to read the whole file every time I review as it is.

jmfranck:
To separate the PR, I would just close this PR, and then make the new branches that are 1 file each.

guinneak198:
I incorporated your edits and made 3 separate PRs:

[- calibrating pulses](https://github.com/jmfrancklab/proc_scripts/pull/112)
[- verifying calibration](https://github.com/jmfrancklab/proc_scripts/pull/113)
[- caluclate integral of indiv capture](https://github.com/jmfrancklab/proc_scripts/actions/runs/10455030646/job/28949063663?pr=114)

jmfranck:
these have all been reviewed.

PR #109, opened 2024-08-07
Gds nutation combo

guinneak198:
processing AG is using for absolute noise measurements


jmfranck:
This seems to involve the capture of two pulses, rather than just one

PR #110, opened 2024-08-12
Proc fid nutation

guinneak198:
processing for nutations (FID and regular)


examples/proc_nutation.py
```diff
@@ -0,0 +1,103 @@
...
+    if "nScans" in s.dimlabels:
+        s.mean("nScans")
+    mysgn = prscr.determine_sign(
+        prscr.select_pathway(s["t2":signal_range], signal_pathway)
+    )
+    # {{{ apply overall zeroth order correction
+    s /= prscr.zeroth_order_ph(prscr.select_pathway(s["t2":0], signal_pathway))
+    d_raw = s.C
+    fl.image(
+        prscr.select_pathway(s["t2":signal_range].C, signal_pathway), ax=ax1
```
jmfranck, line 45: Making copies is not clean code

examples/proc_nutation.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+    ax1.set_title("Raw Data")
+    d_unc = prscr.select_pathway(
+        s["t2":signal_range].C.real, signal_pathway
+    ).integrate("t2")
```
jmfranck, line 50: The real property already makes a copy. No need for for the extra copy here

examples/proc_nutation.py
```diff
@@ -0,0 +1,103 @@
...
+    # }}}
+    # {{{ define mysgn based on phase difference between d_unc and d0
+    mysign = (d0 / d_unc).angle / np.pi
+    mysign = np.exp(1j * np.pi * mysign.run(np.round))
+    # }}} 
+    d_raw *= mysign
+    fl.image(prscr.select_pathway(d_raw, signal_pathway), ax=ax2)
+    ax2.set_title("Same Sign")
+    s = d_raw
```
jmfranck, line 71: Not clean

examples/proc_nutation.py
```diff
@@ -0,0 +1,103 @@
...
+    s.set_units("t2", "s")
+    s.set_error(None)
+    s = prscr.fid_from_echo(s, signal_pathway)
```
jmfranck, line 74: If this is a single pulse mutation, it doesn't make sense to be calling this function

pyspecProcScripts/load_data.py
```diff
@@ -346,6 +346,11 @@ def proc_Hahn_echoph(s, fl=None):
     return s
 
 
+def proc_spincore_FID_nutation_v1(s, fl=None):
+    s = proc_spincore_generalproc_v1(s, fl=fl)
```
jmfranck, line 350: You can just use the dictionary to point straight to the general proc function, since you're not doing anything in addition

jmfranck (changes requested):

examples/proc_nutation.py
guinneak198: added separate script for FID nutations

examples/proc_FID_nutation.py
```diff
@@ -0,0 +1,77 @@
...
+    fl.image(
+        prscr.select_pathway(s["t2":signal_range], signal_pathway), ax=ax1
+    )
+    ax1.set_title("Raw Data")
+    # }}}
+    # {{{ FID slice
+    s.ift("t2")
+    s = s["t2":(0, None)]
+    s *= 2
+    s["t2":0] *= 0.5
```
jmfranck, line 48: it is already an FID!

examples/proc_FID_nutation.py
```diff
@@ -0,0 +1,77 @@
...
+    # }}}
+    s.ft("t2")
+    fl.image(
+        prscr.select_pathway(s, signal_pathway), human_units=False, ax=ax2
+    )
+    ax2.set_title("Phased and FID sliced")
+    s = prscr.select_pathway(s["t2":signal_range].real, signal_pathway)
```
jmfranck, line 55: two calls to select pathway are redundant

examples/proc_FID_nutation.py
```diff
@@ -0,0 +1,77 @@
...
+with psd.figlist_var() as fl:
+    # {{{ set up subplots
+    fl.next("Raw Data with averaged scans", fig=fig)
+    fig.suptitle("FID Nutation %s" % sys.argv[2])
+    # }}}
+    signal_pathway = s.get_prop("coherence_pathway")
+    if "nScans" in s.dimlabels:
+        s.mean("nScans")
+    # {{{ apply overall zeroth order correction
+    s /= prscr.zeroth_order_ph(prscr.select_pathway(s["t2":0], signal_pathway))
```
jmfranck, line 38: I'm confused -- I thought you ended up saying that you  needed to do some weirdness with the phasing, but here you are just zeroth order phasing everything together? This seems inconsistent with your slides

examples/proc_FID_nutation.py
```diff
@@ -0,0 +1,77 @@
...
+            min=s.data.max() * 0.8,
+            max=s.data.max() * 1.5,
+        ),
+        R=dict(value=3e3, min=0, max=3e4),
+        beta_ninety=dict(value=2e-5, min=0, max=1),
+    )
+    f.fit()
+    fit = f.eval(100)
+    fl.plot(fit)
+    psd.gridandtick(plt.gca())
```
jmfranck, line 77: give the beta_90 on the plot?

examples/proc_nutation.py
```diff
@@ -0,0 +1,102 @@
...
+    # }}}
+    signal_pathway = s.get_prop("coherence_pathway")
+    if "nScans" in s.dimlabels:
+        s.mean("nScans")
+    mysgn = prscr.determine_sign(
+        prscr.select_pathway(s["t2":signal_range], signal_pathway)
+    )
+    # {{{ apply overall zeroth order correction
+    s /= prscr.zeroth_order_ph(prscr.select_pathway(s["t2":0], signal_pathway))
+    d_raw = s.C
```
jmfranck, line 43: why are you making many copies

examples/proc_nutation.py
```diff
@@ -0,0 +1,102 @@
...
+    fl.image(
+        prscr.select_pathway(s["t2":signal_range], signal_pathway), ax=ax1
+    )
+    ax1.set_title("Raw Data")
+    d_unc = prscr.select_pathway(
```
jmfranck, line 48: it's not clear why this is called d -- I think this is basically your signal integral, so maybe call it something based on that?

examples/proc_nutation.py
```diff
@@ -0,0 +1,102 @@
...
+        s["t2":signal_range].real, signal_pathway
+    ).integrate("t2")
+    # }}}
+    # {{{ zeroth_order on individual betas
+    for j in range(len(s.getaxis("beta"))):
+        ph0 = prscr.zeroth_order_ph(
+            prscr.select_pathway(
+                s["beta", j]["t2":signal_range], signal_pathway
+            )
+        )
```
jmfranck, line 58: you shouldn't be using a loop here and you shouldn't be using the zeroth order  ph function

examples/proc_nutation.py
```diff
@@ -0,0 +1,102 @@
...
+        s["beta", j] /= ph0
+    d0 = prscr.select_pathway(
```
jmfranck, line 60: as with above, rename this something that's not d0

examples/proc_nutation.py
```diff
@@ -0,0 +1,102 @@
...
+        s["t2":signal_range].real, signal_pathway
+    ).integrate("t2")
+    # }}}
+    # {{{ define mysgn based on phase difference between d_unc and d0
+    mysign = (d0 / d_unc).angle / np.pi
+    mysign = np.exp(1j * np.pi * mysign.run(np.round))
```
jmfranck, line 66: why not just use all the above to overwrite the existing function that determines the sign?

jmfranck: also, why is this used for the 2 pulse variant, but  not for the one pulse variant?

jmfranck (changes requested):

examples/proc_FID_nutation.py
guinneak198: I got confused when you said to not use fid_from_echo (thinking you meant not using any of it's functionalities but now I think I see you are saying the FID portion is unnecessary

examples/proc_nutation.py
guinneak198: edited for simplicity

examples/proc_nutation.py
guinneak198: edited routine to be more straightforward

examples/proc_nutation.py
guinneak198: Here I was applying the phase correction based on the differences (we discussed this in a meeting) but I am realizing that was a diagnostic and shouldnt be applied to actual processing scripts

examples/proc_nutation.py
guinneak198: I removed this mysign stuff as I'm fairly sure it was diagnostic and what I have now works well just using the fid_from_echo. I agree with overwriting the determine_sign but that should be a separate PR I think?

guinneak198:
There was confusion on implementing the phase difference definition of mysign that we discussed in a meeting about the nutations. But I am realizing (I think) that that was simply a diagnostic test and not meant to be part of the actual phasing. Also some confusion on [your comment ](https://github.com/jmfrancklab/proc_scripts/pull/110#discussion_r1714490930) about not using fid_from_echo for the single pulse nutation. I edited and simplified and both run

examples/proc_FID_nutation.py
```diff
@@ -31,18 +31,18 @@
...
-    signal_pathway = s.get_prop("coherence_pathway")
     if "nScans" in s.dimlabels:
         s.mean("nScans")
     # {{{ apply overall zeroth order correction
-    s = prscr.select_pathway(s, signal_pathway)
+    s = prscr.select_pathway(s, s.get_prop("coherence_pathway"))
     s /= prscr.zeroth_order_ph(s["t2":0])
     fl.image(s["t2":signal_range], ax=ax1)
-    ax1.set_title("Raw Data")
+    ax1.set_title("Signal pathway / ph0")
```
jmfranck, line 40: see that you're OK with this

examples/proc_FID_nutation.py
```diff
@@ -31,18 +31,18 @@
...
     # }}}
     s.ift("t2")
     # {{{ Apply phasing
     s["t2"] -= s.getaxis("t2")[0]
+    # PR I'm stopping here -- this is for an FID!!!!
```
jmfranck, line 45: deal with this

proc_nutation.py
```diff
@@ -16,6 +16,7 @@
 logger = init_logging("info")
 max_kHz = 200
 for searchstr, exp_type, nodename, postproc, freq_slice, t_slice in [
+        # PR: either choose to process all, or only one
```
jmfranck, line 19: deal with this

proc_nutation.py
```diff
@@ -33,13 +34,14 @@
         searchstr,
         exp_type=exp_type,
         expno=nodename,
-        postproc=postproc,
+        postproc=postproc,# PR why do you need to manually specify the postproc? that should not be needed!
```
jmfranck, line 37: deal with this

proc_nutation.py
```diff
@@ -33,13 +34,14 @@
...
         lookup=lookup_table,
         fl=fl,
     )
     s = s["t2":freq_slice]
     fl.next("data")
     fl.image(s)
+    # {{{ PR why are we doing the following -- is it needed?? If it is, comment on this, and also include a slide in your pptx
```
jmfranck, line 44: deal with this (applies to entire bracketed region)

proc_nutation.py
```diff
@@ -49,10 +51,12 @@
     s.ft("t2")
     fl.next("After rx offset correction")
     fl.image(s)
+    # }}}
     if "amp" in s.dimlabels:
         plen = s.get_prop("acq_params")["p90_us"] * 1e-6
         logger.info(strm("pulse length is:", plen))
     s.ift("t2")
+    # {{{ PR why would you not just use the fid_from_echo here??
```
jmfranck, line 59: deal with this (applies to entire bracketed region)

jmfranck (changes requested):
this one will require further review after most glaring issues are dealt with

examples/proc_FID_nutation.py
```diff
@@ -44,31 +44,31 @@
...
     s.ft("t2")
     fl.image(s, ax=ax1)
-    d_raw = s.C
     ax1.set_title("Signal pathway / ph0")
     # }}}
     # {{{ Look at phase variation
-    phase_var_s = s.C
-    d_uncorrected = phase_var_s.C.real.integrate("t2")
+    d_uncorrected = s.C.real.integrate("t2")
+    # PR again, you need to rename these
```
jmfranck, line 51: I have gone through multiple review cycles asking for this.  Even if it's on a different PR, you should have seen this when reading through the diff (are you proofing it before saying it's ready for review?)

examples/proc_FID_nutation.py
```diff
@@ -44,31 +44,31 @@
...
     for j in range(len(s.getaxis("beta"))):
         ph0 = prscr.zeroth_order_ph(phase_var_s["beta", j])
         phase_var_s["beta", j] /= ph0
     d_ind_ph0 = phase_var_s.real.integrate("t2")
     mysign = (d_ind_ph0 / d_uncorrected).angle / np.pi
     mysign = np.exp(1j * np.pi * mysign.run(np.round))
-    d_raw *= mysign
+    s *= mysign
```
jmfranck, line 58: I just went ahead and made this change b/c I've asked for this many times, as well.

examples/proc_FID_nutation.py
```diff
@@ -44,31 +44,31 @@
...
     # }}}
+    # {{{ generate the table of integrals and fit
     s = s.real.integrate("t2")
     s.set_error(None)
     A, R, beta_ninety, beta = sp.symbols("A R beta_ninety beta", real=True)
     fl.plot(s, "o", ax=ax3, human_units=False)
-    f = psd.lmfitdata(s)
-    f.functional_form = (
+    s = psd.lmfitdata(s)
+    s.functional_form = (
```
jmfranck, line 68: I didn't comment on this, but it's the same story.  The data changes as it is processed.  There is no need to spin up multiple copies of the data.  This is what we set out as one of the fundamental principles of clean code.

jmfranck (changes requested):
problems with proc_FID_nutation.  Not looking at the other diffs, since it seems like this was not proofed.

proc_nutation.py
guinneak198: wrong script has since been deleted (was not in examples repo and very old)

proc_nutation.py
guinneak198: again wrong script outside of examples not sure what this is so deleted. Correct version inside examples repo

proc_nutation.py
guinneak198: again this was an old old version not in the examples repo and has since been deleted. The up to date version is inside the examples directory

proc_nutation.py
guinneak198: again this was an old old version not in the examples repo and has since been deleted. The up to date version is inside the examples directory

examples/proc_nutation.py
guinneak198: this was not the case and mysign / looking at phase variation was added back in

examples/proc_FID_nutation.py
```diff
@@ -0,0 +1,97 @@
...
+    # {{{ Look at phase variation
+    d_integral = s.C.real.integrate("t2")
+    phase_ind_beta = (
+        s.C
+    )  # copy for applying zeroth order to individual indirect indices
+    for j in range(len(s.getaxis("beta"))):
+        phase_ind_beta["beta", j] /= prscr.zeroth_order_ph(
+            phase_ind_beta["beta", j]
+        )
+    phase_ind_beta.real.integrate("t2")
```
jmfranck, line 58: This doesn't do anything, because the real property creates a copy. If you want to process the data in place, you can use the run function, which you can pass np.real.

examples/proc_FID_nutation.py
```diff
@@ -0,0 +1,97 @@
...
+    mysign = (phase_ind_beta / d_integral).angle / np.pi
+    mysign = np.exp(1j * np.pi * mysign.run(np.round))
```
jmfranck, line 60: Your proposed method for determining the sign, which is what I think this is, should replace the existing function in the package

jmfranck (changes requested):

pyspecProcScripts/phasing.py
```diff
@@ -753,41 +752,46 @@ def hermitian_function_test(
...
-    data_sgn.run(np.real).run(lambda x: np.sign(x))
+    d = select_pathway(s, signal_pathway)
+    d_integral = d.C.real.integrate(direct)
+    ind_phase_ind = d.C
+    for j in range(len(s.getaxis(indirect))):
+        ind_phase_ind[indirect, j] /= zeroth_order_ph(
+            ind_phase_ind[indirect, j]
+        )
+    ind_phase_ind.run(np.real).integrate(direct)
+    mysign = (ind_phase_ind / d_integral).angle / np.pi
```
jmfranck, line 792: none of this makes much sense.  What we want is just this:
```
d = select_pathway(d,signal_pathway) # signal pathway kwargs should be None, and pull from the property if that's true
d = d[direct:signal_frq_range] # because we would like to use this before phasing, it might actually make sense to take the signal frequency range as an argument.
d /= zeroth_order_ph(d)
# we talked about how the previous function is only for signals where we have mixed negative and positive signals.
# for a simple 1D zeroth order, we can use the following two lines to determine the phase.
ph0 = d.C.integrate(direct) 
ph0 /= abs(ph0)
if fl is not None:
    d_indivphased = d/ph0 # individually phased
    d_indivphased /= abs(d_indivphased) # because we only want to subtract phase angle on the next line
    fl.image(d/d_indivphased)
    # we had initially talked about taking the phase of the thing we plot there, and doing the calculation I do on the final line
    # but note that it's mathematically ph0/d*abs(d)*d = ph0*abs(d), whose phase is just given by ph0
return ph0.angle.run(lambda x: np.exp(1j*np.round(x/pi)*pi) # here I just do the whole calculation inside the lambda function
                                         ).run(np.sign) # note that the sign is just to essentially round to -1 or +1

**Review by jmfranck — CHANGES_REQUESTED** [review]
needs significant edits and re-review.

File: pyspecProcScripts/phasing.py (hunk 26 above)
Comment (jmfranck) [reply pyspecProcScripts/phasing.py:792]: note how the loop above is not vectorized --> that was a red flag that something was goofy

File: pyspecProcScripts/phasing.py [hunk 27]
```diff
@@ -427,7 +427,6 @@ def filter_ranges(B, A):
                 N_ratio /= (
                     for_resid.data.size
                 )  # the signal this has been plotted against is signal averaged by N_ratio
-                resi_sum = for_resid[direct, 7:-7].mean(direct).item()
```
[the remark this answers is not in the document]
guinneak198: flake says it's unused/never called

pyspecProcScripts/phasing.py
```diff
@@ -753,41 +752,46 @@ def hermitian_function_test(
...
+    ind_phase_ind.run(np.real).integrate(direct)
+    mysign = (ind_phase_ind / d_integral).angle / np.pi
+    mysign = np.exp(1j * np.pi * mysign.run(np.round))
     if fl is not None:
         fl.next("check sign")
-        fl.image((s.C.setaxis("vd", "#").set_units("vd", "scan #")) * data_sgn)
-        fl.pop_marker()
-    return data_sgn
+        fl.image(d * mysign)
+    return mysign.data
```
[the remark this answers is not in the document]
guinneak198: I think that's linked to what you say [here](https://github.com/jmfrancklab/proc_scripts/pull/110#discussion_r1728030610)? but fixed

examples/proc_FID_nutation.py
```diff
@@ -35,16 +37,28 @@
...
+    # {{{ Apply overall zeroth order correction, select the pathway, and apply a rudimentary alignment
     s /= prscr.zeroth_order_ph(
-        prscr.select_pathway(s["t2":0], s.get_prop("coherence_pathway"))
+        prscr.select_pathway(
+            s["t2":signal_range].sum("t2"), s.get_prop("coherence_pathway")
+        )
     )
+    s = prscr.select_pathway(
+        s["t2":signal_range], s.get_prop("coherence_pathway")
+    )
```
jmfranck, line 48: b/c this gets done over and over below

jmfranck: just select the signal of interest and move on

examples/proc_FID_nutation.py
```diff
@@ -35,16 +37,28 @@
...
+    #     I want to see the signal resonance frequency
+    s.ift("t2")
+    shift = s * np.exp(-(pi**2) * s.fromaxis("t2") ** 2 * (2 * 50**2))
+    shift.ft("t2")
+    shift = shift["t2":signal_range].real.run(abs).argmax("t2")
+    shift.set_error(None)
+    # interestingly, if we were correcting here, we would multiply by the phase
+    # here to perform the shift, which would be efficient!
     s.ft("t2")
+    # }}}
```
jmfranck, line 59: just to improve robustness of integral

pyspecProcScripts/phasing.py
```diff
@@ -753,7 +753,7 @@ def hermitian_function_test(
 
 
 def determine_sign(
-    s, indirect, signal_freq_range, direct="t2", signal_pathway=None, fl=None
+    s, signal_freq_range, direct="t2", signal_pathway=None, fl=None
```
jmfranck, line 756: Here, I just implemented several of the corrections, since I needed to edit for other purposes.

jmfranck (changes requested):
still needs some work.

examples/proc_nutation.py
```diff
@@ -0,0 +1,105 @@
+"""
+Process nutation data
```
jmfranck, line 2: for this one, see the new `rough_table_of_integrals` function that I made.  The only thing that you need to do is to add the `fid_from_echo` function in the right place if the data is `echo_like`

jmfranck: (that replaces most of the processing present in this file)

examples/proc_FID_nutation.py
guinneak198: I think the new rough_table_integrals takes care of this now

examples/proc_FID_nutation.py
guinneak198: you moved this to rough_table_integrals which I like

examples/proc_nutation.py
```diff
@@ -50,7 +49,7 @@
     )
     s.fit()
     fit = s.eval(500)
-    fl.plot(fit, human_units=False)
+    fl.plot(fit, human_units=False, ax=ax3)
```
jmfranck, line 52: this is why we pass back the Axes object from the function

proc_Hahn_echoph.py
```diff
@@ -10,6 +10,8 @@
 
 
 def select_pathway(s, pathway):
+    if len(pathway) == 0:
+        return s
```
jmfranck, line 14: this is the easiest solution to the thing that you are noting below:
if the pathway has already been selected, just pass `{}` -- then this doesn't do anything

pyspecProcScripts/third_level/rough_table_of_integrals.py
```diff
@@ -58,6 +59,7 @@ def rough_table_of_integrals(
         select_pathway(s[direct:signal_range].sum(direct), signal_pathway)
     )
     if echo_like:
+        # PR why are you making copies here?  This doesn't make sense.
```
jmfranck, line 62: deal w/ comment please

pyspecProcScripts/third_level/rough_table_of_integrals.py
```diff
@@ -95,6 +97,9 @@ def rough_table_of_integrals(
     ax2.set_title("Check phase variation along indirect")
     # }}}
     if echo_like:
+        # PR -- the comment below implies that at this point, the sign
+        # has been flipped, so we probably don't want the next line.
+        # You can plot to diagnose if you want ot check
```
jmfranck, line 102: deal w/ comment please

jmfranck (changes requested):
needs further review -- note that I'm first comparing to 18835

examples/proc_nutation.py
guinneak198: I don't want the integral to be part of the subplots because I'd rather show the FID sliced and phased be the third subplot so that the user can see the final phasing before integrating

proc_Hahn_echoph.py
guinneak198: removed function as discussed [here](https://jmfrancklab.slack.com/archives/CV553TPRV/p1724779149443699?thread_ts=1724778674.063839&cid=CV553TPRV)

examples/proc_FID_nutation.py
```diff
@@ -47,21 +47,15 @@
...
-    ax3.set_xlabel(r"$\beta$ / $\mathrm{\mu s \sqrt{W}}$")
-    beta_90 = s.output("beta_ninety") / 1e-6
-    ax3.axvline(beta_90)
-    ax3.text(
+    fl.plot(s, "o", ax=ax_last)
+    fl.plot(fit, ax=ax_last)
+    ax_last.set_title("Integrated and fit")
+    beta_90 = s.output("beta_ninety")
+    ax_last.axvline(beta_90)
+    ax_last.text(
```
jmfranck, line 55: this was a problem that needed to be fixed in pyspecdata.  The figure list was really set up for a one dataset per figure model, which we're obviously messing with here.

pyspecProcScripts/third_level/rough_table_of_integrals.py
```diff
@@ -54,6 +54,7 @@ def rough_table_of_integrals(
     fig.set_figheight(6)
     # {{{ set up subplots
     fl.next("Raw Data with averaged scans", fig=fig)
+    fl.skip_units_check()
```
jmfranck, line 57: you need to update pyspecdata to get this -- we'll want to do this whenever we load multiple subplots.

pyspecProcScripts/third_level/rough_table_of_integrals.py
```diff
@@ -80,7 +81,6 @@ def rough_table_of_integrals(
         ax=ax1,
     )
     ax1.set_title("Signal pathway / ph0")
-    ax1.set_ylabel(r"$\beta$ / $\mathrm{\mu s \sqrt{W}}$")
```
jmfranck: this ruins the generality of this function (which should work for all sorts of stuff, and should be (is) handled by pyspecdata)

jmfranck (changes requested):
I see and liked your solution with 4 vs. 3 plots.  I cleaned this up a little and had to add a hack for pyspecdata to get everything to run smoothly.  Unfortunately, I'm out of time for this now.  Please do triple check if you can.

pyspecProcScripts/third_level/rough_table_of_integrals.py
guinneak198: was trying to get the y axis labels to use the latex symbol but figured out how to do symbols with latex symbols so like you said no need for this!

proc_Hahn_echoph.py
```diff
@@ -1,100 +1,72 @@
...
-    )
-    s.mean("nScans")
-    # }}}
-    # {{{rough centering of sliced data
-    s = s["t2":slice_f]
-    s.ift("t2")
-    rough_center = (
-        abs(s).convolve("t2", 0.01).mean_all_but("t2").argmax("t2").item()
-    )
-    s.setaxis(t2 - rough_center)
```
guinneak198, line 47: a lot of this is now done by fid_from_echo so I updated accordingly and made sure this runs

proc_Hahn_echoph.py
```diff
@@ -1,100 +1,72 @@
...
+            ph0 = prscr.zeroth_order_ph(ph0, fl=fl)
+            psd.logger.info(psd.strm("phasing dimension as one"))
+        else:
+            psd.logger.info(
+                psd.strm(
+                    "there is only one dimension left -- standard 1D zeroth order phasing"
+                )
             )
+            ph0 = ph0 / abs(ph0)
+        s /= ph0
```
guinneak198, line 52: taken from original script - trying to make the least changes as possible from og

pyspecProcScripts/load_data.py
```diff
@@ -333,7 +333,8 @@ def proc_Hahn_echoph(s, fl=None):
     s *= s.shape["nScans"]
     s.squeeze()
     s.reorder(["ph2", "ph1"])
-    s.setaxis("nScans", r_[0:nScans])
+    if "nScans" in s.dimlabels:
+        s.setaxis("nScans", r_[0:nScans])
```
guinneak198, line 337: proc_Hahn_echoph.py was complaining about this so I edited so it runs

guinneak198: specifically the squeeze above gets rid of the nScans dimension if there is only 1 scan

examples/proc_FID_nutation.py
```diff
@@ -27,11 +27,12 @@
 with psd.figlist_var() as fl:
     if "nScans" in s.dimlabels:
         s.mean("nScans")
+    s.rename("beta", r"$\beta$")  # label beta with latex
     s, ax_last = prscr.rough_table_of_integrals(
         s, signal_range, fl=fl, title=sys.argv[2], echo_like=False
     )
     # {{{ generate the table of integrals and fit
-    A, R, beta_ninety, beta = sp.symbols("A R beta_ninety beta", real=True)
+    A, R, beta_ninety, beta = sp.symbols("A R beta_ninety $\\beta$", real=True)
```
jmfranck, line 35: this is not a good change b/c:
```
In [58]: import sympy as sp

In [59]: b = sp.symbols('beta')

In [60]: sp.latex(b)
Out[60]: '\\beta'
```

examples/proc_nutation.py
```diff
@@ -27,11 +27,12 @@
 with psd.figlist_var() as fl:
     if "nScans" in s.dimlabels:
         s.mean("nScans")
+    s.rename("beta", r"$\beta$")
     s, ax_last = prscr.rough_table_of_integrals(
         s, signal_range, fl=fl, title=sys.argv[2], echo_like=True
     )
     # {{{ generate the table of integrals and fit
-    A, R, beta_ninety, beta = sp.symbols("A R beta_ninety beta", real=True)
+    A, R, beta_ninety, beta = sp.symbols("A R beta_ninety $\\beta$", real=True)
```
jmfranck, line 35: see above

examples/proc_nutation.py
```diff
@@ -27,11 +27,12 @@
 with psd.figlist_var() as fl:
     if "nScans" in s.dimlabels:
         s.mean("nScans")
+    s.rename("beta", r"$\beta$")
```
jmfranck, line 30: just leave this alone -- the solution would be to make a PR in pyspecdata to `unitify_axis`

proc_Hahn_echoph.py
```diff
@@ -7,13 +7,13 @@
...
-with psd.fl_mod() as fl:
+with psd.figlist_var() as fl:
     for searchstr, exp_type, nodename, postproc, label_str, slice_f in [
         (
             "200302_alex_probe_water",
-            "test_equip",
+            "ODNP_NMR_comp/ODNP/old/2020",
             "signal",
-            "spincore_Hahn_echoph_v1",
+            "proc_Hahn_echoph",
```
jmfranck: why are you pointing to different data?

proc_Hahn_echoph.py
```diff
@@ -7,13 +7,13 @@
 
 
 signal_pathway = {"ph1": 1, "ph2": 0}
-with psd.fl_mod() as fl:
```
jmfranck: `fl_mod` is actually defined and belongs to proc scripts -- search for `class fl_mod` --> it supplies other plotting styles.

proc_Hahn_echoph.py
```diff
@@ -27,26 +27,13 @@
...
         # }}}
-        # {{{ apply phase corrections
-        best_shift = prscr.hermitian_function_test(
-            prscr.select_pathway(s, signal_pathway)
-        )
-        psd.logger.info(psd.strm("best shift is", best_shift))
         s_uncorrected = s.C.ft("t2")
-        s.setaxis("t2", lambda x: x - best_shift).register_axis(
-            {"t2": 0}, nearest=False
-        )
```
jmfranck: I don't understand why you're making these modifications

proc_Hahn_echoph.py
```diff
@@ -65,29 +52,20 @@
...
-        fl.image(s.C.convolve("t2", 10))  # so that resolution of
-        # plot isn't higher than that of screen -- note that we do
-        # this on a copy of the data, since we don't actually want
-        # to alter the data here
         # }}}
         # {{{slice out FID from echo
-        s.ift("t2")
-        s = prscr.slice_FID_from_echo(s)
-        s.ft("t2")
+        s = prscr.fid_from_echo(s, signal_pathway)
```
jmfranck: I don't understand why you're making these modifications

jmfranck: this might be a case of genuinely just explain

pyspecProcScripts/load_data.py
jmfranck:
```
s.setaxis('nScans','#')
```

pyspecProcScripts/simple_functions.py
```diff
@@ -89,3 +89,97 @@ def select_pathway(*args, **kwargs):
...
+    if fl is not None:
+        fl.next("check sign")
+        if "vd" in s.dimlabels:
+            fl.image(
+                s.C.setaxis("vd", "#").set_units("vd", "scan #") * data_sgn
+            )
+        else:
+            fl.image(s * data_sgn)
+        fl.pop_marker()
+    return data_sgn
```
jmfranck: why are you adding this when you're spinning up a whole new function for this purpose?

proc_Hahn_echoph.py
jmfranck: I don't understand why you are modifying this script.

proc_Hahn_echoph.py
guinneak198: 1) we moved the data to the old folder so I just redirected it 2) I comment that the load data applies a squeeze which removes the nScans dimension and so the averaging throws an error 3) updated fid_from_echo replaces everything

proc_Hahn_echoph.py
guinneak198: see [here](https://jmfrancklab.slack.com/archives/CV553TPRV/p1724856423822879?thread_ts=1724855644.907539&cid=CV553TPRV)

proc_Hahn_echoph.py
guinneak198: we had moved the data to this folder a while ago and the version is titled wrong as is in the load_data. but we decided to delete this script anyway

proc_Hahn_echoph.py
guinneak198: thank you! was wondering why I got the issue with fl_mod not belonging to pyspecdata

pyspecProcScripts/simple_functions.py
guinneak198: this must've been a messy commit sorry not sure how this got added!

guinneak198: but I fixed

PR #112, opened 2024-08-19
script to process calibration of pulses

guinneak198:


jmfranck (changes requested):
quick comments -- needs re-review

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,271 @@
...
+        carrier = (
+            d.get_prop("acq_params")["carrierFreq_MHz"] * 1e6
+        )  # signal frequency
+        n = np.floor(
+            (carrier + SW / 2) / SW
+        )  # nearest integer multiple of sampling frequency
+        nu_a = carrier - n * SW  # aliasing frequency
+        if nu_a < 0:
+            perceived_nu = abs(nu_a)
+            # we removed the negative frequencies so it's counterpart is the abs
```
jmfranck, line 93: This is not an explanation for why you're just pulling the abs.  Draw yourself a sketch, and come up with an explanation based on the sketch.  Btw, I do not know if this is actually correct!

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,271 @@
...
+        # {{{ apply frequency filter
+        d.ift("t")
+        SW = 1 / (d["t"][1] - d["t"][0])  # sample rate
+        carrier = (
+            d.get_prop("acq_params")["carrierFreq_MHz"] * 1e6
+        )  # signal frequency
+        n = np.floor(
+            (carrier + SW / 2) / SW
+        )  # nearest integer multiple of sampling frequency
+        nu_a = carrier - n * SW  # aliasing frequency
```
jmfranck, line 90: It would seem to me that if this is negative, you would want -carrier plus an integer SW instead

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,271 @@
...
+        if nu_a < 0:
+            perceived_nu = abs(nu_a)
+            # we removed the negative frequencies so it's counterpart is the abs
+            d.run(np.conj)
+        else:  # otherwise subtract the difference between the signal and nyquist from nyquist's
+            perceived_nu = SW - nu_a
```
jmfranck, line 96: What is this?

jmfranck: As in why?

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,271 @@
...
+        d.ft("t")
+        d["t" : (0, perceived_nu - 0.5 * slicewidth)] *= 0
```
jmfranck, line 98: It does make sense to copy the data for this, because you're trying different filters

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,271 @@
...
+        d["t" : (perceived_nu + 0.5 * slicewidth, None)] *= 0
+        # }}}
+        d.ift("t")
+        indiv_plots(abs(d), "filtered analytic", "red")
+        fl.next("collect filtered analytic", legend=True)
+        for j in range(d.shape["t_pulse"]):
```
jmfranck, line 104: It seems like this for block mostly does the same thing that you do further down. Why not just move this code in with the stuff down below

jmfranck: And remove the redundancy

jmfranck: I also don't understand what you are doing here that is not handled by the indiv_plots function. 

Overall, it's very likely that the code needs to be cleaned up, or else you need in line code comments to explain what is going on

examples/calibrate_tsqrtP.py
guinneak198: addressed with new function

examples/calibrate_tsqrtP.py
guinneak198: addressed in new function

examples/calibrate_tsqrtP.py
guinneak198: you are thinking of the wrong PR this script only uses the HH for filtering (we found this is the optimal filter and works better than the lorentzian

examples/calibrate_tsqrtP.py
guinneak198: knowing this is rhetorical I've added comments in the script for clarification

jmfranck:
I'm putting off dealing with this one until we merge the lawn that has the sign function in it.

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,322 @@
...
+        ), "The wrong postproc_type was set so you most likely used the wrong script for acquisition"
+        amplitude = d.get_prop("acq_params")["amplitude"]
+        fl.basename = f"amplitude = {amplitude:.2f}"
+        # {{{ ensure units are set, the preprocessing will output the statement
+        #     that they were not set but to still be able to run this script we
+        #     include the following
+        if not d.get_units("t") == "s":
+            d.set_units("t", "s")
+        if not d.get_units("t_pulse") == "s":
+            d.set_units("t_pulse", "s")
```
jmfranck, line 68: this implies the units might be set to something else, and gives no warning.
We want:

- if the units are not set, something is wrong --> issue a warning
- only set the units if this is `None` --> if it's set to something else, don't mess with it.

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,322 @@
...
+            fl.push_marker()
+            fl.basename = None
+            fl.next("collect filtered analytic")
+            fl.plot(abs(s["t":int_range]), alpha=0.3)
+            fl.pop_marker()
+            # }}}
+            beta_v_t["t_pulse", j] = (
+                abs(s["t":int_range]).integrate("t").data.item()
+            )  # t * sqrt(P_amp)
+            beta_v_t["t_pulse", j] /= np.sqrt(2)  # t * sqrt(P_rms)
```
jmfranck, line 142: just include in previous line

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,322 @@
...
+        # {{{ Add labels and grids to the fit plots
+        for j in (
+            r"Amplitude*$t_{pulse}$ vs $\beta$",
+            r"Amplitude*$t_{pulse}$ vs $\beta$, zoomed",
+        ):
+            fl.next(j)
+            psd.gridandtick(plt.gca())
+            plt.ylabel(r"$At_{pulse}$ / $\mathrm{\mu s}$")
+            plt.xlabel(r"$\beta$ / $\mathrm{\mu s \sqrt{W}}$")
+        # }}}
```
jmfranck, line 322: after other PRs are merged, compare to other functions with `switch_to_plot`

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,322 @@
...
+        ] *= amplitude  # we only t_pulse to be multiplied by amplitude for plotting purposes only!
+        fl.plot(
+            (beta_v_t.C / 1e-6),
+            color=thiscolor,
+            label="amplitude = %f" % amplitude,
+        )
+        beta_v_t[
+            "$A t_{pulse}$"
+        ] /= amplitude  # need to divide back out for the determination of the coefficients below
+        psd.gridandtick(plt.gca())
```
jmfranck, line 176: the ordering of the lines makes this very hard to read -- move the gridandtick up to after the plot, and leave these two lines that return the data to its previous state together

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,322 @@
...
+            beta_v_t["t_pulse"].copy() / 1e-6
+        )  # because our ppg wants μs
+        t_v_beta.set_units("μs").set_units(r"$\beta$", "s√W")
+        # use as temp for ultimate coeff
+        # }}}
+        # {{{ Determine linear and nonliear coefficients
+        c_nonlinear = t_v_beta[
+            r"$\beta$":(None, linear_threshold)
+        ].C  # make copy to allow the next
+        #                                                               line not to cause the data to
```
jmfranck, line 214: comment is weirdly formatted.  Leave the hash at the start, but put all the letters on the second like so that it's aligned under make -- but the entire rest of the comment here.  Then highlight and `gq`

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,322 @@
...
+                    return np.polyval(
+                        c_nonlinear[::-1], desired - linear_threshold
+                    )
+
+            ret_val = np.vectorize(zonefit)(desired)
+            if ret_val.size > 1:
+                return ret_val
+            else:
+                return ret_val.item()
+
```
jmfranck, line 255: JF reviewed this file to here -- note other comments.

jmfranck (changes requested):

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,322 @@
...
+        fl.next(r"Amplitude*$t_{pulse}$ vs $\beta$", legend=True)
+        t_v_beta.set_plot_color_next()
```
jmfranck, line 257: since you are already using this, and it should be simpler -- can you use this instead of manually pulling the color cycler above?

examples/calibrate_tsqrtP.py
guinneak198: realized these lines weren't even needed to begin with and just extra garbage so just deleted no need for switch_to_plot

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,328 @@
...
+                    int_range[0] * 1e6 - 1,
+                    0.25,
+                    r"$t_{pulse} \sqrt{P_{tx}} = %f \mathrm{μs} \sqrt{\mathrm{W}}$"
+                    % (beta_v_t["t_pulse", j].item() / 1e-6),
+                )
+            # }}}
+        # {{{ show what we observe -- how does β vary with the programmed pulse length
+        fl.basename = None  # reset so all amplitudes are on same plots
+        beta_v_t.set_plot_color_next() #cycle color list to next color
+        thiscolor = beta_v_t.get_plot_color() # set thiscolor to the above
```
guinneak198, line 163: need both lines, if I set thiscolor to `beta_v_t.set_plot_color_next()` then everytime this color is called it would cycle the color leading to all lines being a different color

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,328 @@
...
+            beta_v_t["t_pulse"].copy() / 1e-6
+        )  # because our ppg wants μs
+        t_v_beta.set_units("μs").set_units(r"$\beta$", "s√W")
+        # use as temp for ultimate coeff
+        # }}}
+        # {{{ Determine linear and nonliear coefficients
+        c_nonlinear = t_v_beta[
+            r"$\beta$":(None, linear_threshold)
+        ].C  # make copy to allow the next
+        #      line not to cause the data to actually shift
```
guinneak198, line 217: fixed comment format

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,328 @@
...
+            color=thiscolor,
+            label="amplitude = %f" % amplitude,
+        )
+        plt.axhline(  # the linear threshold is the threshold above which beta is linear
+            y=linear_threshold / 1e-6,
+            color=thiscolor,
+            alpha=0.8,
+            label=f"linear threshold for amp={amplitude}",
+        )
+        psd.gridandtick(plt.gca())
```
guinneak198, line 181: move gridandtick to section with actual plotting

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,328 @@
...
+            fl.push_marker()
+            fl.basename = None
+            fl.next("collect filtered analytic")
+            fl.plot(abs(s["t":int_range]), alpha=0.3)
+            fl.pop_marker()
+            # }}}
+            beta_v_t["t_pulse", j] = abs(s["t":int_range]).integrate(
+                "t"
+            ).data.item() / np.sqrt(
+                2
```
guinneak198, line 141: include sqrt(2) in this line

examples/calibrate_tsqrtP.py
```diff
@@ -0,0 +1,328 @@
...
+        if d.get_units("t") is None:
+            logging.info(
+                "Units for your t axis weren't set to anything so I am setting them to s"
+            )
+            d.set_units("t", "s")
+        if d.get_units("t_pulse") is None:
+            logging.info(
+                "Units for your t_pulse axis weren't set to anything so I am setting them to s"
+            )
+            d.set_units("t_pulse", "s")
```
guinneak198, line 68: "if the units are not set, something is wrong --> issue a warning
only set the units if this is None --> if it's set to something else, don't mess with it"

guinneak198 (commented):
reviewed and implemented JF comments

PR #113, opened 2024-08-19
script to verify calibration of pulses

guinneak198:


examples/verify_tsqrtP_calibration.py
```diff
@@ -0,0 +1,150 @@
...
+        #    "pulse_calib_2",
+        #),  # low power
+        #(
+        #    "240805_test_calib_amp0p2_a_pulse_calib.h5",
+        #    "pulse_calib_1",
+        #),  # low power
+        #(
+        #    "240805_test_calib_amp1_pulse_calib.h5",
+        #    "pulse_calib_1",
+        #),  # low power
```
jmfranck, line 44: don't include commented out lines -- you probably want to uncomment them

examples/verify_tsqrtP_calibration.py
```diff
@@ -0,0 +1,150 @@
...
+        n = np.floor(
+            (carrier + fn) / SW
+        )  # how far is the carrier from the left side of the spectrum (which is at SW/2), in integral multiples of SW
+        nu_a = (
+            carrier - n * SW
+        )  # find the aliased peak -- again, measuring from the left side
+        center = SW - abs(nu_a)
+        d.ft("t")
+        d["t" : (0, center - 0.5 * slicewidth)] *= 0
+        d["t" : (center + 0.5 * slicewidth, None)] *= 0
```
jmfranck, line 100: as with the other PR, don't do this center thing.  If the signal aliases to negative, then it aliases to negative.

jmfranck (changes requested):
quick comments -- needs re-review

examples/verify_tsqrtP_calibration.py
guinneak198: we discussed that we had thrown out negative frequencies and so this will not work

examples/verify_tsqrtP_calibration.py
```diff
@@ -0,0 +1,148 @@
...
+            (carrier + SW / 2) / SW
+        )  # nearest integer multiple of sampling frequency
+        nu_a = carrier - n * SW  # find the aliasing frequency
+        if nu_a < 0:
+            perceived_nu = abs(
+                nu_a
+            )  # we removed the negative frequencies so it's counterpart is the abs
+            d.run(np.conj)
+        else:  # otherwise subtract the difference between the signal and nyquist from nyquist's
+            perceived_nu = SW - nu_a
```
jmfranck, line 95: Same comments, as in other PR's about the fact that you have two variables that appear to be filling the same role and that the absolute value is not actually explained.  Based on comments you provide in the code here, it should be possible to sketch yourself a figure of the aliasing and see why it's justified to simply take the absolute value here, if in fact, it is justified.  


I would suggest hashing this out **here** since this is the simplest poll request that involves this issue

jmfranck (changes requested):

guinneak198:
Added in the function that we made in the proc_ind_GDS_capture to identify the frequency of the signal just so that it runs but once that branch is merged the diff will only be the one script

examples/verify_tsqrtP_calibration.py
guinneak198: handled with new funciton

examples/verify_tsqrtP_calibration.py
```diff
@@ -0,0 +1,150 @@
...
+        amplitude = d.get_prop("acq_params")["amplitude"]
+        fl.basename = f"amplitude = {amplitude}"
+        if not d.get_units("t") == "s":
+            print(
+                "units weren't set for the t axis or else I can't read them from the hdf5 file!"
+            )
+            d.set_units("t", "s")
+        d *= V_atten_ratio
+        d /= np.sqrt(50)  # V/sqrt(R) = sqrt(P)
+        pulse_lengths = d.get_prop("programmed_t_pulse_us")
```
jmfranck, line 57: Convert to SI without prefix or change the name of the variable

examples/verify_tsqrtP_calibration.py
```diff
@@ -0,0 +1,150 @@
...
+        d["t" : (nu_a + 0.5 * HH_width, None)] *= 0
+        # }}}
+        d.ift("t")
+        indiv_plots(abs(d), "filtered analytic", "red")
+        # {{{ Diagnostic to plot all pulses on the same plot to see none were skipped
+        fl.next("collect filtered analytic", legend=True)
+        for j in range(d.shape["beta"]):
+            s = d["beta", j].C
+            s["t"] -= abs(s).contiguous(lambda x: x > 0.03 * s.max())[0][0]
+            fl.plot(abs(s), alpha=0.3, label=f"{j}")
```
jmfranck, line 104: There are a few problems here: 
- this is an unnecessary copy 
- you are repeating code that you do below 
- just doesn't make sense not to do this plot below, with the data that you actually use

jmfranck: I've made this comment before!

examples/verify_tsqrtP_calibration.py
```diff
@@ -0,0 +1,150 @@
...
+                    0.25,
+                    r"$t_{90} \sqrt{P_{tx}} = %f \mathrm{μs} \sqrt{\mathrm{W}}$"
+                    % (beta["beta", j].item() / 1e-6),
+                )
+                # }}}
+        # {{{ show what we observe -- how does β vary with the programmed pulse length
+        fl.basename = None  # we want to plot all amplitudes together now
+        fl.next(r"Measured $\beta$ vs programmed $\beta$")
+        fl.plot(
+            (beta.C / 1e-6).set_units("μs√W"),
```
jmfranck, line 143: Unnecessary copy, and also it should handle the units for you.

jmfranck (changes requested):

jmfranck:
With above review, the verify... script probably needs further review

examples/verify_tsqrtP_calibration.py
guinneak198: at this point thinking it over, now that we have the decreasing_ind thing this plot becomes obsolete and just extra noise so I am getting rid of it entirely

guinneak198:
removed all instances of copy and unused variables and removed unneeded diagnostic

examples/verify_tsqrtP_calibration.py
```diff
@@ -49,35 +49,30 @@
         fl.basename = f"amplitude = {amplitude:.1f}"
         if not d.get_units("t") == "s":
             psd.logger.info(
-                psd.strm(
-                    "units weren't set for the t axis or else I can't read them from the hdf5 file!"
-                )
+                "units weren't set for the t axis or else I can't read them from the hdf5 file!"
```
jmfranck: strm is not needed here

examples/verify_tsqrtP_calibration.py
```diff
@@ -49,35 +49,30 @@
...
             )
             d.set_units("t", "s")
         d *= V_atten_ratio  # V at output of amplifier
         d /= np.sqrt(50)  # V/sqrt(R) = sqrt(P)
 
         # {{{ functions that streamline plotting the desired pulse length datasets
-        def switch_to_plot(d, thisj):
-            thislen = d.get_prop("programmed_t_pulse")[thisj] / 1e-6
+        def switch_to_plot(d, j):
+            thislen = d.get_prop("programmed_t_pulse")[j] / 1e-6
```
jmfranck: I rolled back -- `j` or `k` etc are by definition an index you're iterating over, so the `this` is a little weird

examples/verify_tsqrtP_calibration.py
```diff
@@ -49,35 +49,30 @@
...
             fl.next(f"pulse length = {thislen:.2f} μs")
 
-        def indiv_plots(d, thislabel, thiscolor, thisj=None):
+        # PR this change vs. c772298 was not good  -- it's a bit forced, and as noted below, you're leaving out stuff that was done before.  So I rolled it back
+        def indiv_plots(d, thislabel, thiscolor):
```
jmfranck: see (and remove) comment.  Please check that everything still works the way you want

jmfranck: by "a bit force" I mean that you're switching from something fairly simple to something more complicated.  That's not bad, but it doesn't fulfill the end purpose (read the comment surrounding the code you're trying to replace below)

jmfranck: (the PR comment refers to the entire function)

examples/verify_tsqrtP_calibration.py
jmfranck: you're not really responding to the comment here -- I'm just saying it's in the wrong place

jmfranck: I just did it myself.

examples/verify_tsqrtP_calibration.py
```diff
@@ -71,6 +71,7 @@ def indiv_plots(d, thislabel, thiscolor):
                         d["beta", j],
                         alpha=0.2,
                         color=thiscolor,
+                        # PR the label here is weird -- I think you should give it a more meaningful label
```
jmfranck, line 74: see (+ remove) comment

examples/verify_tsqrtP_calibration.py
```diff
@@ -118,6 +112,14 @@ def indiv_plots(d, thislabel, thiscolor):
...
             int_range[0] -= 2e-6
             int_range[-1] += 2e-6
+            # {{{ plot the data that we're integrating all together
+            # PR -- this is the correct choice based on what I'm saying in the PR comments
+            fl.push_marker()
+            fl.basename = None
+            fl.next("collect filtered analytic")
+            fl.plot(abs(s["t":int_range]), alpha=0.3)
+            fl.pop_marker()
+            # }}}
```
jmfranck, line 122: here I fixed the issues with the diagnostic plot -- note that there is no copy.
I chose to set basename to None to just put everything on the same plot.  The push/pop marker are not the only way of doing this -- if you know the name of the next plot after this, you can just put a "next" after this, instead (or, more a next statement above down, if that makes sense)

jmfranck (changes requested):
still needs to review. I implemented various changes myself to help out.

examples/verify_tsqrtP_calibration.py
guinneak198: I think this section ended up removing [this guy](https://github.com/jmfrancklab/proc_scripts/pull/113#discussion_r1733182045) which makes sense so that the different amplitudes aren't all on top of eachother

pyspecProcScripts/simple_functions.py
```diff
@@ -158,8 +159,13 @@ def find_apparent_anal_freq(s):
     """
     carrier = s.get_prop("acq_params")["carrierFreq_MHz"] * 1e6
     dt = s["t"][1] - s["t"][0]
+    # In analytic signal the negative frequencies were tossed.
+    # So we first need to check if the carrier is in the SW of
+    # the analytic signal or if the signal was aliased
     if carrier < 1 / dt:
-        print("You are in the clear and no aliasing took place!")
+        psd.logger.info(
```
jmfranck, line 166: this is no good --> logging should be imported independently -- it doesn't belong to pyspecdata

pyspecProcScripts/simple_functions.py
```diff
@@ -1,6 +1,9 @@
 "First order functions for very simple (a few lines) data manipulation"
 import numpy as np
+import pyspecdata as psd
+import logging
 
+logging.basicConfig(format="%(levelname)s:%(message)s", level=logging.INFO)
```
guinneak198, line 6: If I just used logging the message wouldn't show up in the terminal (even if I used psd.strm) so I follow [this](https://docs.python.org/3/howto/logging.html)

guinneak198 (commented):
went line by line and tested

pyspecProcScripts/simple_functions.py
jmfranck: you don't want this here.  Logging will not show up unless you turn it on *in the script*.  So whatever example you have should have a `psd.init_logging()` at the top.

examples/verify_tsqrtP_calibration.py
```diff
@@ -0,0 +1,150 @@
...
+            verify_beta["beta", j] = (
+                abs(s["t":int_range]).integrate("t").data.item()
+            )  # tp * sqrt(P_amp)
+            verify_beta["beta", j] /= np.sqrt(2)  # tp*sqrt(P_rms)
```
jmfranck, line 121: this is a bit odd -- just include the sqrt on the first line.

examples/verify_tsqrtP_calibration.py
```diff
@@ -0,0 +1,150 @@
...
+        fl.next(r"Measured $\beta$ vs programmed $\beta$")
+        fl.plot(
+            (verify_beta / 1e-6).set_units("μs√W"),
+            color=thiscolor,
+            label="Amplitude = %0.2f" % amplitude,
+        )
+        plt.xlabel(r"Programmed $\beta$ / $\mathrm{\mu s \sqrt{W}}$")
+        plt.ylabel(r"Measured $\beta$ / $\mathrm{\mu s \sqrt{W}}$")
+        psd.gridandtick(plt.gca())
+        # }}}
```
jmfranck, line 150: aside from comments above, this file should be good to go.

pyspecProcScripts/simple_functions.py
jmfranck: btw `psd.init_logging` includes something like the config, and particularly points the log file to a particular location.

The reason why we prefer logging over printing for anything that gets committed, is that it's much easier to read through the log to figure out what's up vs. looking on the command line.

jmfranck: along those lines, I would actually make the statement that says everything is  good a debug, not info --> we don't need to see that everything is OK, only be alerted if something is funky

jmfranck:
@guinneak198 note that I looked at this, and the most recent comments are not dealt with

PR #114, opened 2024-08-19
script to calculate integral of GDS pulse capture

guinneak198:


examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,108 @@
...
+        raw_color = next(color_cycle)
+        fl.plot(s, color=raw_color, label="raw analytic")
+        abs_color = next(color_cycle)
+        fl.plot(abs(s), color=abs_color, label="abs(analytic)")
+        # {{{ apply frequency filter
+        SW = 1 / (s["t"][1] - s["t"][0])  # sample rate
+        carrier = (
+            s.get_prop("acq_params")["carrierFreq_MHz"] * 1e6
+        )  # signal frequency
+        fn = SW / 2  # nyquist frequency
```
jmfranck, line 45: this is not the nyquist frequency.  Please substitute this below.  It's more legible that way

examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,108 @@
...
+        n = np.floor(
+            (carrier + fn) / SW
+        )  # how far is the carrier from the left side of the spectrum (which is at SW/2), in integral multiples of SW
+        nu_a = (
+            carrier - n * SW
+        )  # find the aliased peak -- again, measuring from the left side
+        center = SW - abs(nu_a)
```
jmfranck, line 52: you don't want this -- I will comment below

examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,108 @@
...
+        s.ft("t")
+        fl.next("Frequency Domain")
+        fl.plot(s, color=raw_color, label="raw analytic")
+        fl.plot(abs(s), color=abs_color, label="abs(analytic)")
+        # {{{ lorentzian filter
+        x = s.C.getaxis("t")
+        for_L = s.C  # make copy for lorentzian filter
+        Lambda = 15.19e6 - 14.61e6
```
jmfranck, line 60: call this Delta_nu -- note how it's in a different color here b/c it's (similar to) a python keyword

examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,108 @@
...
+            carrier - n * SW
+        )  # find the aliased peak -- again, measuring from the left side
+        center = SW - abs(nu_a)
+        s.ft("t")
+        fl.next("Frequency Domain")
+        fl.plot(s, color=raw_color, label="raw analytic")
+        fl.plot(abs(s), color=abs_color, label="abs(analytic)")
+        # {{{ lorentzian filter
+        x = s.C.getaxis("t")
+        for_L = s.C  # make copy for lorentzian filter
```
jmfranck, line 59: this line is not needed

examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,108 @@
...
+            )
+            .setaxis("t", x)
+            .set_units("t", for_L.get_units("t"))
+        )
+        for_L *= L
+        for_L /= s_max
+        # }}}
+        # {{{ heaviside hat functions
+        # {{{ determine center frequency
+        s.ift("t")
```
jmfranck, line 74: all of this gets deleted.

examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,108 @@
...
+        # }}}
+        s.ft("t")
+        wide_HH = s.C
+        wide_HH["t" : (0, center - loose_HH_width)] *= 0
+        wide_HH["t" : (center + loose_HH_width, None)] *= 0
```
jmfranck, line 79: can we rename this as `Heaviside_filtered`

The Lorentzian can be acheived with a very simple one-line thing:
```
Lorentzian_filtered = s * 1/(1+(s.fromaxis('t')-nu0)**2*...whatever you need to get the width you want goes here...
```
```

File: examples/proc_GDS_capture_int.py [hunk 7]
```diff
@@ -0,0 +1,108 @@
...
+        for_L *= L
+        for_L /= s_max
+        # }}}
+        # {{{ heaviside hat functions
+        # {{{ determine center frequency
+        s.ift("t")
+        # }}}
+        s.ft("t")
+        wide_HH = s.C
+        wide_HH["t" : (0, center - loose_HH_width)] *= 0
```
jmfranck, line 78: do not use this "center" thing.  Use nu_a.  It might be a negative number, and that's OK -- that will happen if you alias your signal to negative.

Also, add a line here -- if nu_a is negative, check that plus or minus HH width doesn't include zero.  Also, you need to multiply HH width by 0.5 here, or else it's not a width.

jmfranck (changes requested):
line 45 is straightforward.

I need to re-review from line 52 on after your edits

examples/proc_GDS_capture_int.py
jmfranck: for the "add a line" I'm just saying
```
assert (
             0 > nu_a + 0.5*HH_width ) or ( 0 < nu_a - 0.5*HH_width), "unfortunately the region I want to filter includes DC -- this is probably not good, and means you should pick a different timescale for your scope so this doesn't happen"
             ```

Comment (guinneak198) [reply examples/proc_GDS_capture_int.py]: https://jmfrancklab.slack.com/archives/CV553TPRV/p1724163442806419

Comment (jmfranck) [reply examples/proc_GDS_capture_int.py]: we discussed this -- sounds like either comments or change are in progress

File: examples/proc_GDS_capture_int.py [hunk 8]
```diff
@@ -0,0 +1,108 @@
...
+HH_width = 2e6
+int_slop = 1e-6
+
+with psd.figlist_var() as fl:
+    for filename, nodename in [
+        (
+            "240819_amp0p05_beta_max_pulse_capture.h5",
+            "pulse_capture_9",
+        ),
+        # (
```
jmfranck, line 19: Probably uncomment?

examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,108 @@
...
+        carrier = (
+            s.get_prop("acq_params")["carrierFreq_MHz"] * 1e6
+        )  # signal frequency
+        n = np.floor(
+            (carrier + SW / 2) / SW
+        )  # nearest integer multiple of sampling frequency
+        nu_a = carrier - n * SW  # aliasing frequency
+        if nu_a < 0:
+            perceived_nu = abs(
+                nu_a
```
jmfranck, line 55: This is not explained.  If you are claiming you want the abs, you need to explain why.  Why is the negative peak located here?

jmfranck: Also `nu_a` is the perceived frequency, so the multiple variables flying around are confusing

jmfranck (changes requested):

jmfranck:
Still need to re-review

pyspecProcScripts/simple_functions.py
```diff
@@ -137,25 +137,30 @@ def determine_sign(s, direct="t2", fl=None):
...
-            # from below or above SW / 2
-            nu_a = abs(carrier - n * SW)
-    return s, nu_a
+            # when stored, we threw out the negative frequencies
+            # so that means that we captured an aliased copy of the
+            # negative frequency
+            nu_a = -carrier + n * SW
+            # we need to make note of the fact that the phase of our final time domain signal
+            # (after filtering and up-conversion) will be flipped
+            isflipped = True
```
jmfranck, line 165: a lot of this was repeated, and this is equivalent to the abs, but only because of the conditional, so it actually makes more sense this way

jmfranck (changes requested):
please review comments and implement docstring as noted

examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,105 @@
...
+        ),
+    ]:
+        s = psd.find_file(
+            filename, expno=nodename, exp_type="ODNP_NMR_comp/test_equipment"
+        )
+        # {{{ define basename
+        amplitude = s.get_prop("acq_params")["amplitude"]
+        beta_us_sqrt_W = s.get_prop("acq_params")["beta_90_s_sqrtW"] * 1e6
+        fl.basename = (
+            f"amplitude = {amplitude}, $\\beta$ = {beta_us_sqrt_W} \n"
```
jmfranck, line 38: Give units

examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,105 @@
...
+            filtered_data.ft("t")
+            fl.next("Frequency Domain")
+            fl.plot(
+                abs(filtered_data), color=thiscolor, alpha=0.5, label=label
+            )
+            filtered_data.ift("t")
+            beta = (
+                abs(filtered_data).integrate("t").data.item()
+                / np.sqrt(2)
+                * 1e6
```
jmfranck, line 95: This is a style/ legibility thing, but do not manipulate the order of magnitude here.. do it when you print out the value.

** All quantities should be in SI base units unless the name of the variable specifically indicates otherwise**

examples/proc_GDS_capture_int.py
```diff
@@ -0,0 +1,105 @@
...
+            )
+            fl.next(r"$\sqrt{P_{analytic}}$ vs $t_{pulse}$")
+            plt.text(
+                (s["t"].max() * 1e6) / 2 - 50,
```
jmfranck, line 99: Same as above

jmfranck (changes requested):
Directly addressing these comments should do the trick

examples/proc_GDS_capture_int.py
```diff
@@ -1,3 +1,4 @@
+# PR docstring needed
```
jmfranck, line 1: Sorry -- I missed this before -- you just need a description.

examples/proc_GDS_capture_int.py
```diff
@@ -33,9 +34,7 @@
         )
         # {{{ define basename
         amplitude = s.get_prop("acq_params")["amplitude"]
-        # Only adjusting beta to us for basename (it fits/looks better)
-        beta_us_sqrt_W = s.get_prop("acq_params")["beta_90_s_sqrtW"] * 1e6
-        fl.basename = f"amplitude = {amplitude}, ${{\\beta}}$ = {beta_us_sqrt_W} ${{\\mathrm{{\\mu s \\sqrt{{W}}}}}}$ \n"
+        fl.basename = f"amplitude = {amplitude}, ${{\\beta}}$ = {s.get_prop('acq_params')['beta_90_s_sqrtW'] / 1e-6} ${{\\mathrm{{\\mu s \\sqrt{{W}}}}}}$ \n"
```
jmfranck, line 37: I'm just compacting

jmfranck (changes requested):
just need a docstring, then that should do it.

examples/proc_GDS_capture_int.py
jmfranck: Is this the processing for your "simple capture"? Or what is this?

guinneak198: yes processing for individual capture of single pulse length - mentioned acq file in docstring

PR #115, opened 2024-08-19
version where RX scans has been dealt with prior to saving

guinneak198:


pyspecProcScripts/load_data.py
```diff
@@ -840,6 +845,7 @@ def proc_field_sweep_v2(s):
     "chirp": proc_capture,
     "spincore_CPMG_v1": proc_spincore_CPMG_v1,
     "spincore_SE_v1": proc_spincore_SE_v1,
+    "spincore_SE_v2": proc_spincore_SE_v2,
```
jmfranck, line 848: Just feed the general proc as the value. Your function is only calling the general proc.

PR #116, opened 2024-08-30
preprocessing for current run_nutation.py

guinneak198:


guinneak198:
I was going to process the newly acquired nutation data and got the error there is no v6 this PR addresses that

jmfranck (approved):

PR #119, opened 2024-09-18
fix ODNP postproc

jmfranck:
The previous postproc functions relevant to ODNP seem to be obsolete stuff.  It seem to run fine if I just use the generalproc function.  I update the proc_raw documentation to include an example where I read both the E(p) and the T1.


pyspecProcScripts/load_data.py
```diff
@@ -349,28 +349,9 @@ def proc_Hahn_echoph(s, fl=None):
...
-    if fl is not None:
-        fl.next("time domain (all $\\Delta p$)")
-        fl.image(s.C.setaxis("vd", "#").set_units("vd", "scan #"))
-    s.ft("t2")
-    if fl is not None:
-        fl.next("frequency domain (all $\\Delta p$)")
-        fl.image(s.C.setaxis("vd", "#").set_units("vd", "scan #"), black=False)
-    return s
+    if s.get_prop("coherence_pathway") is None:
+        s.set_prop("coherence_pathway", {"ph1": 0, "ph2": +1})
```
guinneak198, line 353: shouldn't ph2 be -1?

.github/workflows/flake8.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         run: |
           if [ -n "$CHANGED_FILES" ]; then
             echo "Running Flake8 on changed files: $CHANGED_FILES"
-            flake8 --inline-quotes='"' --max-line-length=88 --ignore=E231,E501,E225,E226,E261,E401,E306,E293,W293,E741,W291,E265,E302,E305,E203,W503,Q002 $CHANGED_FILES
+            flake8 --inline-quotes='"' --max-line-length=79 --ignore=E203,E225,E226,E231,E261,E265,E293,E302,E305,E306,E401,E731,E741,Q002,W291,W293,W503 $CHANGED_FILES
```
jmfranck, line 60: please leave this alone -- I realized this got committed with the other stuff rather than pushed to master, but it is a needed update.

examples/proc_raw.py
```diff
@@ -22,6 +26,9 @@
 import sys
 import os
 from itertools import cycle
+import pyspecdata as psd
+
+psd.init_logging(level="debug")
```
guinneak198, line 31: is this necessary for the script to run?

jmfranck, line 31: this is just useful

pyspecProcScripts/load_data.py
jmfranck, line 353: here it seemed it might be wise to add info that was missing before, but I don't see anything at a glance in the stuff that I'm deleting that's not handled by the generalproc function

pyspecProcScripts/load_data.py
```diff
@@ -656,31 +637,6 @@ def proc_spincore_ODNP_v3(s, fl=None):
...
-    if fl is not None:
-        fl.next("Raw Data \n Frequency Domain")
-        fl.image(s)
-        s.ift("t2")
-        fl.next("Raw Data \n Time Domain")
-        fl.image(s)
-        s.ft("t2")
-    return s
-
-
```
jmfranck, line 683: this actually didn't run b/c the data doesn't have a `power` dimension, so it dies on line 663.  But also, I don't see any added value here on top of the generalproc function -- at all.

This does beg the question of whether we want to handle the business of turning the log into a power axis as part of the postproc, here, or if we want to handle it independently.  That could be a point for discussion

jmfranck: but for now, it's clear this function offers no value any longer, and should die.

guinneak198: not sure why your examples would be calling this version of postproc since the most recent version in examples/combined_ODNP calls v3 of IR and v5 of ODNP

pyspecProcScripts/load_data.py
guinneak198: I fixed

jmfranck: no, it should be +1, because there is a third pulse that inverts

pyspecProcScripts/load_data.py
```diff
@@ -349,28 +353,9 @@ def proc_Hahn_echoph(s, fl=None):
...
-    if fl is not None:
-        fl.next("time domain (all $\\Delta p$)")
-        fl.image(s.C.setaxis("vd", "#").set_units("vd", "scan #"))
-    s.ft("t2")
-    if fl is not None:
-        fl.next("frequency domain (all $\\Delta p$)")
-        fl.image(s.C.setaxis("vd", "#").set_units("vd", "scan #"), black=False)
-    return s
+    if s.get_prop("coherence_pathway") is None:
+        s.set_prop("coherence_pathway", {"ph1": 0, "ph2": -1})
```
jmfranck, line 357: it's +1

pyspecProcScripts/load_data.py
```diff
@@ -701,7 +672,8 @@ def proc_spincore_generalproc_v1(s, include_tau_sub=True, fl=None):
         neworder += [nonphdims[j] for j in sizeidx]
     # }}}
     s.reorder(neworder)
-    # {{{ put ph_overall outside, if it exists, since there should be nothing outside that
+    # {{{ put ph_overall outside, if it exists, since there should be nothing
+    # outside that
```
jmfranck, line 676: for all comments that are wrapped, please put second line of text under the first line of text (not including curly brackets)

examples/proc_raw.py
```diff
@@ -18,6 +18,12 @@
 ``py proc_raw.py CPMG_9 240620_200uM_TEMPOL_pm_generic_CPMG.h5
 ODNP_NMR_comp/Echoes``
 
+# PR -- this field sweep is not good field sweep
+# data.  (1) The acquisition length is small.  (2)
+# Phase cycling noise and difference in signal
+# amplitude across nScans indicates bad repetition
+# delay.  (3) You can't see the rise and fall of
+# the ESR resonance.
```
jmfranck, line 26: please see and deal with comment

pyspecProcScripts/load_data.py
```diff
@@ -1,16 +1,11 @@
...
 altered in anyway.
->>>>>>> master
-
-Parameters
-==========
-self:   nddata or h5 file
-
-Returns
-=======
-nddata that has been FTed and in coherence domain
```
jmfranck, line 12: I cleaned this up -- it's a module docstring, not a function docstring

examples/proc_raw.py
jmfranck: it's needed to understand what is going on.

examples/proc_raw.py
guinneak198: replaced with wider and better resolved with newer data

PR #122, opened 2024-09-26
Use rough table

wfkincaid:


examples/test_field_sweep.py
```diff
@@ -0,0 +1,81 @@
...
+
+signal_pathway = {"ph1": 1}
+with psd.figlist_var() as fl:
+    for thisfile, exptype, nodename, post_proc, label_str in [
+        (
+            "240924_13p5mM_TEMPOL_field.h5",
+            "ODNP_NMR_comp/field_dependent",
+            "field_1",
+            "field_sweep_v4",
+            "9-24-24 13.5 mM TEMPOL sweep",
```
jmfranck, line 14: I would argue this can/should be constructed from the date and chemical name -- something like
```
label_str = f"{s.get_prop('acq_params')['date']} {s.get_prop('acq_params')['chemical_name']} sweep"
```

examples/test_field_sweep.py
```diff
@@ -0,0 +1,81 @@
...
+        ###{{{ Raw data
+        label = label_str
+        fl.basename = "(%s)" % label
+        fl.next("raw data")
+        fl.image(s, interpolation="auto")
+        ###}}}
+        ###{{{ Time Domain
+        field_axis = []
+        for indir_val in range(np.shape(s.getaxis("indirect"))[0]):
+            field_axis.append(s.getaxis("indirect")[indir_val][0])
```
jmfranck, line 34: this is not good b/c it's not `vectorized` code (look that up)

I am pretty sure that the axis is a `structured array` and that you're trying to pull a particular `field` -- you can see `proc_raw.py` for an example of how we do this.

examples/test_field_sweep.py
```diff
@@ -0,0 +1,81 @@
...
+        rx_offset_corr = s["t2" : (t2_max * 0.75, None)]
+        rx_offset_corr = rx_offset_corr.mean(["t2"])
+        s -= rx_offset_corr
+        fl.next("time domain")
+        s.ft(["ph1"])
+        fl.image(s["t2":(None, 0.05)])
+        ###}}}
+        ###{{{ Before Phasing
+        # print(np.shape(s['indirect','data{0}','carrierFreq']))
+        s = s["t2":(-800, 800)]
```
jmfranck, line 53: this is a hard-coded manual slice -- this is what we are trying to avoid.

examples/test_field_sweep.py
```diff
@@ -0,0 +1,81 @@
...
+            fl.plot(prscr.select_pathway(s, signal_pathway))
+        ###}}}
+        # fl.show();quit()
+        # s.ift('t2')
+        # print(s.getaxis('t2')[0])
+        # fl.show();quit()
+        # s['t2'] -= s.getaxis('t2')[0]
+        s.set_units("t2","s")
+        s.ft("t2")
+        s = prscr.fid_from_echo(s, signal_pathway, fl=fl)
```
jmfranck, line 68: but `rough_table_of_integrals` includes `fid_from_echo`, so there should be no need for this

jmfranck (changes requested):
I know that you were not intending for this to be done yet, but I looked at it, and it seemed that a few quick comments could be helpful

examples/test_field_sweep.py
jmfranck: here "this" is the label

examples/test_field_sweep.py
jmfranck: here the words that are in `this format` (aside from the file name) are "vocab words" that you can look up  -- all of them are relevant to numpy

examples/test_field_sweep.py
jmfranck: so -- I think that this and the following comment fall under the general umbrella of *almost everything before the `rough_table_of_integrals` function should be unnecessary

examples/test_field_sweep.py
```diff
@@ -0,0 +1,81 @@
...
+            fl.plot(prscr.select_pathway(s.C.mean('nScans'),signal_pathway))
+        else:
+            fl.plot(prscr.select_pathway(s, signal_pathway))
+        ###}}}
+        # fl.show();quit()
+        # s.ift('t2')
+        # print(s.getaxis('t2')[0])
+        # fl.show();quit()
+        # s['t2'] -= s.getaxis('t2')[0]
+        s.set_units("t2","s")
```
jmfranck, line 66: setting the units should never be necessary.  There was a bug where the units were not being saved that should be fixed in the most recent versions of pyspecdata.  But even when not fixed, the postproc functions hsould be taking care of this.

examples/test_field_sweep.py
```diff
@@ -0,0 +1,81 @@
...
+        ###}}}
+        ###{{{ Before Phasing
+        # print(np.shape(s['indirect','data{0}','carrierFreq']))
+        s = s["t2":(-800, 800)]
+        fl.next("before phasing")
+        print(psd.ndshape(s))
+        if "nScans" in s.dimlabels:
+            fl.plot(prscr.select_pathway(s.C.mean('nScans'),signal_pathway))
+        else:
+            fl.plot(prscr.select_pathway(s, signal_pathway))
```
jmfranck, line 59: I would be careful that you're not duplicating diagnostic plots that are already available from `rough_table_of_integrals` and from `fid_from_echo` which is inside it.

examples/test_field_sweep.py
wfkincaid: I dont see how i can do this above where s is initialized

jmfranck (changes requested):
the comments for the previous review haven't been addressed yet.  Please address those first, then re-request review (and don't forget to comment show it shows up on my slack!!)

jmfranck:
> the comments for the previous review haven't been addressed yet. Please address those first, then re-request review (and don't forget to comment show it shows up on my slack!!)

This was in response to the fact that you requested a review.

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,29 @@
-"""Check NMR/ESR resonance ratio using a field sweep
-====================================================
-
-Analyzes field sweep data.  Determines the resonance frequency from the carrier
-frequency and the offset of the signal, converts to MHz, then divide by the
-Bridge 12 μw frequency stored in the file to get the resonance ratio of MHz/GHz
-"""
```
jmfranck, line 7: let's keep the docstring

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,29 @@
...
-]:
-    s = find_file(
+        "field_1",
+        "field_sweep_v4",
+        "240924 13.5 mM TEMPOL field sweep",
+    )
+    s = psd.find_file(
         thisfile,
-        exp_type=exp_type,
+        exp_type=exptype,
```
jmfranck, line 15: keep the naming with the underscore

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,29 @@
...
-    s["t2", 0] *= 0.5
-    s.ft("t2")
-    s = select_pathway(s, signal_pathway)
-    s /= zeroth_order_ph(s.C.mean("t2"))
-    nu_NMR = []
-    assert set(s.getaxis("indirect").dtype.names) == {
-        "Field",
-        "carrierFreq",
-    }, "'indirect' axis should be a structured array that stores the carrier frequency and the field"
-    # {{{ the following works fine for all the cases that we've tested, and so
```
jmfranck, line 75: this block could be re-used without much modification

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,29 @@
...
-    # the end.
-    all_offsets = zeros(len(s.getaxis("indirect")))
-    for z in range(len(s.getaxis("indirect"))):
-        fl.plot(s["indirect", z], ax=ax_list[1])
-        if z == 0:
-            ax_list[1].axvline(color="k", x=freq_slice[0])
-            ax_list[1].axvline(color="k", x=freq_slice[-1])
-        offset = s["indirect", z].C.mean("nScans").argmax("t2").item()
-        all_offsets[z] = offset
-        carrier_freq_MHz = s.getaxis("indirect")[z]["carrierFreq"]
```
jmfranck, line 89: to use this, you would choose the option to use the carrier frequency rather than the field, and then, you could remove the final square bracket

In looking at this, I'm realizing that this loop can probably be vectorized, but we can deal with that after getting a loop that works.

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,29 @@
...
-        field = s.getaxis("indirect")[z]["Field"]
```
jmfranck, line 90: I don't see the point of using the field here, at all -- we are not interested in the field -- only in the NMR frequency.

The only reason I pull the field in my modification is because the carrier frequency is stored incorrectly by the pulse sequence.

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,29 @@
...
-        -1j
-        * 2
-        * pi
-        * nddata(all_offsets, [-1], ["indirect"])
-        * s.fromaxis("t2")
-    )
-    s.ft("t2")
-    # }}}
-    fl.plot(s.C.mean("nScans"), ax=ax_list[2])
-    s = s["t2":freq_slice].mean("nScans").integrate("t2")
```
jmfranck, line 110: lines 98-110 here are not useful to us -- this type of stuff is done inside the rough table of integrals function

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,29 @@
...
-            s.getaxis("resonance ratio")[0] : s.getaxis("resonance ratio")[
-                -1
-            ] : 100j
-        ],
-        "resonance ratio",
-    )
-    polyline = fl.plot(
-        field_fine.eval_poly(fitting, "resonance ratio"),
-        label="fit",
-        ax=ax_list[3],
```
jmfranck, line 133: this block again is going to be useful to us

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,29 @@
...
+    use_freq = False
+    if use_freq:
+        s["indirect"] = s["indirect"]["carrierFreq"] * 1e6
+        s.set_units("indirect", "Hz")
+    else:
+        # I wanted to use the carrier (use_frep=True), but it seems like the
+        # first point isn't stored properly
+        s["indirect"] = s["indirect"]["Field"]
+        s.set_units("indirect", "G")
+    prscr.rough_table_of_integrals(s, fl=fl)
```
jmfranck, line 29: note that I'm making comments about stuff that gets removed currently before this, but this block would actually come before that.

pyspecProcScripts/phasing.py
```diff
@@ -318,7 +318,7 @@ def fid_from_echo(
     d.ift(direct)
     # {{{ apply phasing, and check the residual
     d[direct] -= d.getaxis(direct)[0]
-    if fl is not None:
+    if fl is not None and fl.basename is not None:
```
jmfranck, line 321: when I tried to show the diagnostic plots, I needed this

pyspecProcScripts/third_level/rough_table_of_integrals.py
```diff
@@ -87,7 +102,7 @@ def rough_table_of_integrals(
     # and handle newer data where it is set I have the following if/else
     signal_pathway = signal_pathway or s.get_prop("coherence_pathway")
     # {{{ Apply overall zeroth order correction, select the pathway, and apply
-    # a rudimentary alignment
+    #     a rudimentary alignment
```
jmfranck, line 105: note how wrapped comments should line up the text, like this

jmfranck (changes requested):
here I make comments while reviewing the diff.  I think it makes sense to leave this PR open until we finish both of our discussion targets relevant to this file.  Here I comments on the parts of the file that are currently delete (and that should be!) but that help us to achieve the second of the two discussion targets that involves changing this file.

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,132 @@
...
+    frq_slice = peak_range
+    fl.next("phased data")
+    fl.plot(s)
+    plt.axvline(x=frq_slice[0])
+    plt.axvline(x=frq_slice[-1])
+    print("t2 as a dimlabel is ", s["t2"])
+    if "nScans" in s.dimlabels:
+        s = s["t2":frq_slice].C.mean("nScans").integrate("t2")
+    else:
+        s = s["t2":frq_slice].C.integrate("t2")
```
wfkincaid, line 112: this fails here now

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,132 @@
...
-        ],
-        "resonance ratio",
-    )
-    polyline = fl.plot(
-        field_fine.eval_poly(fitting, "resonance ratio"),
-        label="fit",
-        ax=ax_list[3],
+    peak_range = nu_B12_GHz
+    peak_range = 1.2 * np.r_[-0.5, 0.5] * peak_range + np.mean(peak_range)
+    frq_slice = peak_range
```
wfkincaid, line 103: this needs to be changed to take the new peak_range (not this manually determined one)

examples/proc_fieldSweep.py
```diff
@@ -1,142 +1,75 @@
...
-    ax_list[3].set_title("Field Sweep ppt")
-    fitting = s.polyfit("resonance ratio", order=2)
-    field_fine = nddata(
-        r_[
-            s.getaxis("resonance ratio")[0] : s.getaxis("resonance ratio")[
-                -1
-            ] : 100j
-        ],
-        "resonance ratio",
+    c_poly = s.polyfit("indirect", 4)
```
wfkincaid, line 53: I was having trouble with this and earlier versions of this line but resolved this (needed to pull recent changes in pyspecdata)

PR #123, opened 2024-09-26
Hack acq params

jmfranck:


guinneak198:
it seems like this is a GUI? I see widget and buttons, etc. Reading over however it's not clear to me how I would open the GUI. For example, FLInst MWtune opens the GUI for microwave tuning and I'm pretty sure this is explained in the documentation. What would be the terminal command here?

jmfranck:
Run the script. This is used to edit hdf5 files after they have been created.

examples/hack_acq_params.py
```diff
@@ -0,0 +1,259 @@
...
+            if key not in self.node:
+                self.node.create_group(key)  # Create group if it doesn't exist
+            group = self.node[key]
+            node_dict = NodeAsDict(group)
+            for subkey, subvalue in value.items():
+                node_dict[
+                    subkey
+                ] = subvalue  # Recursively assign subkeys and values
+        elif isinstance(value,bytes):
+            self.node.attrs[key] = value.decode( "utf-8")
```
guinneak198, line 65: below in save_changes (line 226) we convert to bytes so this is needed to decode

examples/hack_acq_params.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def save_changes(self):
+        self.hdf_file = h5py.File(self.hdf_filename, "r+")
+        if self.nodename in self.hdf_file:
+            self.node = self.hdf_file[self.nodename]
+        else:
+            raise ValueError(
+                f"Expno '{self.nodename}' not found in the HDF5 file."
+            )
+        self.read_acq_params()
```
guinneak198, line 219: see debugging thread `[here](https://jmfrancklab.slack.com/archives/CV553TPRV/p1735668659397929)`

PR #125, opened 2024-09-27
E(p) -- Use rough table ep

wfkincaid:


jmfranck:
there's a lot of (figurative) noise in this, which at this point should really be just a call to rough_table_of_integrals.  So, I'm just going to close it.

PR #126, opened 2024-09-30
Proc ep rtoi func

wfkincaid:


wfkincaid:
I think I did this right, it should be the use_rough_table PR + the new file `proc_Ep.py`

examples/proc_Ep.py
```diff
@@ -27,5 +27,3 @@
     s["indirect"] = s["indirect"]["start_times"]
     s.set_units("indirect", "s")
     prscr.rough_table_of_integrals(s, fl=fl)
-    fl.show()
```
jmfranck: the with block handles show.

wfkincaid:
I added a lot to proc_Ep trying to incorp in the power log stuff but it further made me feel that the power log should be handled by a function. One reason for this is that some of the info from the power log will be needed for fitting and plotting along the power axis with proc_FIR

examples/proc_Ep.py
```diff
@@ -32,7 +32,6 @@
     )
     orig_axis = s["indirect"]  # let's save this so we
     #                           can pass it to the log
-    s = prscr.convert_to_power(thisfile, exptype, s, fl=fl)
```
jmfranck: I just moved this to the end so that we could see the older plot (which I thought was kind of nice) as well as the final plot.

examples/proc_Ep.py
```diff
@@ -52,3 +51,8 @@
         plt.FuncFormatter(lambda x, _: str(datetime.timedelta(seconds=x)))
     )
     # }}}
+    # (delete when read) just note that I moved this to the end
```
jmfranck, line 54: yes, just delete.

pyspecProcScripts/convert_to_power.py
```diff
@@ -8,33 +8,48 @@
...
     power log to useable axis
     Parameters
     ===========
-    filename: thisfile,
-
-    exptype: exp_type
-
     s: nddata
-
+        This is an nddata whose "indirect"
```
jmfranck, line 17: I added to the docstring a bit

pyspecProcScripts/convert_to_power.py
```diff
@@ -8,33 +8,48 @@
...
+        (modified in-place)
+        When we're done, the axis coordinates for "indirect" are a (not
+        structured) array containing the average power recorded on the
+        log between each start_time and stop_time.
+        The axis coordinate errors are the standard deviations of the
+        same.
     """
     my_filename = psd.search_filename(
-        re.escape(filename), exp_type=exptype, unique=True
+        re.escape(filename), exp_type=exp_type, unique=True
```
jmfranck, line 39: just using consistent naming

pyspecProcScripts/convert_to_power.py
```diff
@@ -8,33 +8,48 @@
...
-        thislog = logobj.from_group(f["log"])
+        thislog = logobj.from_group(f[node_name])
         log_array = thislog.total_log
     assert log_array.dtype.names == ("time", "Rx", "power", "cmd"), str(
         log_array.dtype.names
     )
+    assert s["indirect"].dtype.names == (
+        "start_times",
+        "stop_times",
+    ), str(s["indirect"].dtype.names)
```
jmfranck, line 50: just checking that the data is in the expected format

pyspecProcScripts/convert_to_power.py
```diff
@@ -50,31 +65,39 @@ def convert_to_power(filename, exptype, s, fl=None):
...
         )
         .set_units("time", "s")
     )
-    log_vs_time = prscr.dBm2power(log_vs_time)
+    # TODO ☐: these are the power readings on the gigatronics.  They
+    # need to be converted to actual powers by adding in a constant dB
+    # value that represents our directional coupler and the associated
+    # losses.  This number is about (but not exactly) 20.  Alex should
+    # be able to tell you what it is.
+    log_vs_time = prscr.dBm2power(log_vs_time + 20)
```
jmfranck, line 77: I posted a message about this in the processing scripts channel

jmfranck: as I see it, this is the only thing that needs to be taken care of before closing the PR.

pyspecProcScripts/convert_to_power.py
```diff
@@ -108,12 +134,15 @@ def convert_to_power(filename, exptype, s, fl=None):
...
+            mean_power_vs_time, "o"
+        )  # this  should be a *single* o at the center of each power step.
+        #    Its y value should be the avaerage power for that step, and its
+        #    error bars should give the standard deviation of the power over
+        #    the step
     # }}}
-    s["indirect"].data = mean_power_vs_time.data
+    s.setaxis("indirect", mean_power_vs_time.data).set_error(
+        "indirect", mean_power_vs_time.get_error()
+    ).set_units("indirect", "W")
```
jmfranck, line 147: the `s['indirect']` is a numpy array.  It has a `data` attribute, but I think that's not what you're going for here. 

btw, what I'm doing in the setaxis call is the same as `s['indirect'] = mean_power_vs_time.data` I just like stringing the calls together.

jmfranck (changes requested):

PR #127, opened 2024-09-30
Proc fir rtoi

wfkincaid:


examples/proc_Ep.py
```diff
@@ -13,7 +13,7 @@
     thisfile, exptype, nodename, post_proc, lookup = (
         "240924_13p5mM_TEMPOL_ODNP_1.h5",
         "ODNP_NMR_comp/ODNP",
-        "ODNP",
+        "FIR_34dBm",
```
jmfranck, line 16: You should be pulling an IR dataset -- not the ODNP dataset from the same file.

examples/proc_Ep.py
```diff
@@ -27,5 +27,3 @@
     s["indirect"] = s["indirect"]["start_times"]
     s.set_units("indirect", "s")
     prscr.rough_table_of_integrals(s, fl=fl)
-    fl.show()
```
jmfranck: again -- `with` handles the showing

examples/proc_raw.py
```diff
@@ -114,4 +114,5 @@ def image_or_plot(d):
         image_or_plot(forplot)
         d = select_pathway(d, d.get_prop("coherence_pathway"))
         fl.next("with coherence pathway selected")
+        plt.gca().set_title("select " + str(d.get_prop("coherence_pathway")))
```
jmfranck, line 117: I added this to be more informative

pyspecProcScripts/load_data.py
```diff
@@ -668,6 +668,7 @@ def proc_spincore_generalproc_v1(s, include_tau_sub=True, fl=None):
         #            this
     # {{{ always put the phase cycling dimensions on the outside
     neworder = [j for j in s.dimlabels if j.startswith("ph")]
+    neworder.sort() # it's confusing if the pulses don't come in order
```
jmfranck, line 671: I added this, which we will generally want.

examples/proc_raw.py
jmfranck: note that this is the wrong pathway for the IR -- Δp_1 is the first pulse, which inverts the population (magnetization along z) so it should be 0 (population has a coherence order of 0).  Δp_2 is the 90 pulse, which is followed by an echo, so it's +1 (echo converts to -1)

I do not have time right now to track if this is set in the data or in the preproc.  In either case, hacking the preproc function so that it gets set to the correct value is probably the way to go, though if it's set in the ppg, should write a small PR to just change the value set there and increment the version number.

jmfranck: (please pull in @guinneak198 for help on that as needed!)

jmfranck (changes requested):

examples/proc_Ep.py
wfkincaid: This was ODNP, i think it just got changed in the wrong file

examples/proc_raw.py
wfkincaid: Do I still need to do this? it looks like you made the requested change in load_data ("hack_..._IR_v3")

examples/proc_FIR.py
```diff
@@ -23,21 +24,28 @@
...
     )
     for nodename in [
-        # "FIR_noPower",
+        "FIR_noPower",
         "FIR_34dBm",
     ]:
-        print(nodename)
+        fl.basename = nodename # this is a good example of how we can
+        #                        use basename to make it easy to deal
+        #                        with multiple datasets
```
jmfranck, line 32: note comment

examples/proc_FIR.py
```diff
@@ -23,21 +24,28 @@
...
             expno=nodename,
             postproc=post_proc,
             lookup=prscr.lookup_table,
         )
+        # TODO ☐: if the units are not already set, that should be handled in
+        # the postproc.  (This was a problem with pyspecdata, and in any newer
+        # pulse sequence, the units should be set)
+        # NOTE: you can just delete "TODO" items as you resolve them -- no need
+        # to check them off.
         s.set_units("vd", "s")
```
jmfranck, line 45: please address

examples/proc_FIR.py
```diff
@@ -23,21 +24,28 @@
...
-        clock_correction = True
```
jmfranck: just moved up top

examples/proc_FIR.py
```diff
@@ -23,21 +24,28 @@
...
         indirect = "vd"
         direct = "t2"
+        # TODO ☐: clock correction should be a new function in the library
```
jmfranck, line 48: Please just move the code into a different module -- just like you recently did for `convert_to_power`

examples/proc_FIR.py
```diff
@@ -67,51 +75,72 @@
...
+        # before you actually take the integral over range of frequencies.
+        # But, I think it's useful for you to think about this as well and
+        # implement a solution.
         if "nScans" in s.dimlabels:
-            s_int = s.C.mean("nScans")
-        else:
-            s_int = s.C
-        print(s_int)
-        s.set_plot_color_next()
+            s.mean("nScans")
```
jmfranck, line 88: the TODO always applies to the stuff below -- here the averaging over nScans

examples/proc_FIR.py
```diff
@@ -67,51 +75,72 @@
...
+        # TODO ☐: below, you are tearing apart the nddata object and putting it
+        # back together -- that's always a bad sign
```
jmfranck, line 90: this comment I largely take care of and break down next steps below -- but this is a good rule of thumb

examples/proc_FIR.py
```diff
@@ -67,51 +75,72 @@
                 fl.next("after auto-clock correction")
                 fl.image(s)
             # }}}
-        np.squeeze(s)
-        s, _ = prscr.rough_table_of_integrals(s, fl=fl)
+        s.squeeze()
```
jmfranck, line 78: squeeze affects the dimensions, so we need to use the pyspecdata method.  I'm actually surprised this didn't give an error, but hopefullly you can see how squeeze is different from e.g. sqrt or log, etc.

examples/proc_FIR.py
```diff
@@ -67,51 +75,72 @@
...
         W = (
             s.get_prop("acq_params")["FIR_rep"] * 1e-6
             + s.get_prop("acq_params")["acq_time_ms"] * 1e-3
         )
-        functional_form = Mi * (
+        # (delete when done) the code didn't match our definition of "clean"
+        # because you're creating many variable names for the same dataset.
+        # Also, you are defining variables in order to use them only on one
+        # other line of code, rather than combining.  The way around a lot of
+        # this is to think about how you can *reorder* the code.
```
jmfranck, line 101: see this

examples/proc_FIR.py
```diff
@@ -67,51 +75,72 @@
...
+        s.functional_form = Mi * (
             1 - (2 - sympy.exp(-W * R1)) * sympy.exp(-vd * R1)
         )
-        IR_data = psd.nddata(s_int.C.data, ["vd"])
-        IR_data.setaxis("vd", s_int.getaxis("vd"))
-        f = psd.lmfitdata(IR_data)
-        f.functional_form = functional_form
-        f.set_guess(
+        # TODO ☐: the M_inf on the next line should be set based on the data
+        # (some multiple of the max and min)
```
jmfranck, line 107: please address

examples/proc_FIR.py
```diff
@@ -67,51 +75,72 @@
...
-        )
-        ax = plt.gca()
+        s.fit()
+        s_fit = s.eval(200)
+        psd.plot(s_fit,ax=ax_last, alpha=0.5) # here, we plot the fit
+        #                                       together with the table of
+        #                                       integrals.
+        # TODO ☐: next pull from code below to collect the normalized fits into a single plot.
+        # I don't really see the value of the unnormalized anymore because we have each
+        # individual curve together with the table of integrals, now.
```
jmfranck, line 119: please address

pyspecProcScripts/third_level/rough_table_of_integrals.py
```diff
@@ -162,7 +166,9 @@ def rough_table_of_integrals(
     )
     # }}}
     s = s[direct:signal_range].real.integrate(direct).set_error(None)
-    psd.plot(s, "-o", ax=ax4)
+    if inc_plot_color:
+        s.set_plot_color_next()
+    psd.plot(s, "o", ax=ax4)
```
jmfranck, line 171: there is a plot color property of nddata that you can get and set with `set_plot_color` and `get_plot_color`.  Any copies/etc of your nddata then inherit that property, making it easy to color a single dataset in the same way.  This function here sets the property based on the default matplotlib color cycle.

jmfranck (changes requested):

examples/proc_FIR.py
wfkincaid: what do you mean by this? should I be returning the list of dropped dimensions?

examples/proc_FIR.py
```diff
@@ -15,16 +15,15 @@
...
-
+# TODO ☐: to directly answer your question, the thing you get from
+#         matplotlib isn't a cycle, but a "cycler", which is different.
+#         You need to feed it to "cycle" from "itertools" in order to
+#         make it into a "cycle" and be able to use "next" on it.
+#         BUT you should't do that, anyways.  The nddata object that
+#         comes out of rough_table_of_integrals will always plot with
+#         the same color (as will copies of it).  You can also get the
+#         color of that object (e.g. for use in an unrelated dataset
+#         axvline, etc) with `s.get_plot_color()`
```
jmfranck, line 40: do read this explanation -- the long story short is that because of this, I'm able to just delete several keyword arguments below, and everything works.

examples/proc_FIR.py
```diff
@@ -50,9 +49,8 @@
         indirect = "vd"
         direct = "t2"
         if clock_correction:
-            # {{{ clock correction
-            s = prscr.clock_correct(s, fl=fl)
-            # }}}
+            s = prscr.clock_correct(s)
+        # TODO ☐: why do you return dropped?
```
jmfranck, line 58: just address -- to me, it seems like `s.squeeze()` should be sufficient.

examples/proc_FIR.py
```diff
@@ -11,10 +11,24 @@
...
+thisfile, exptype, post_proc, lookup = (
+    "240924_13p5mM_TEMPOL_ODNP_1.h5",
+    "ODNP_NMR_comp/ODNP",
+    "spincore_IR_v3",
+    prscr.lookup_table,
+)
+R1nodenames = [
+    "FIR_noPower",
+    "FIR_34dBm",
+]
```
jmfranck, line 27: I moved these up b/c this is the stuff that you would change when using this script as an example.

pyspecProcScripts/clock_correct.py
```diff
@@ -38,7 +38,8 @@ def clock_correct(s, fl=None):
         fl.next("clock correction")
         fl.plot(s_clock, ".", alpha=0.7)
     clock_corr = s_clock.argmax("clock_corr").item()
-    plt.axvline(x=clock_corr, alpha=0.5, color="r")
+    if fl is not None:
+        plt.axvline(x=clock_corr, alpha=0.5, color="r")
```
jmfranck, line 42: I turned off the diagnostic plot for clock correction, which I think we usually don't want to see, and this line kept showing up.

jmfranck (changes requested):

examples/proc_FIR.py
```diff
@@ -0,0 +1,107 @@
...
+# Because we are going ot want to get both R1 fit values as well as the
+# associated errors, we collect the results in an nddata rather than
+# just e.g. a list
+R1data = psd.ndshape([("power", len(R1nodenames))]).alloc(dtype=np.float64)
+# TODO ☐: to directly answer your question, the thing you get from
```
wfkincaid, line 32: What is my 'to do' for this?

examples/proc_FIR.py
wfkincaid: I had `s.squeeze()` but you were suprised this worked. I think `s = s.squeeze()` is the correct way to do this. I returned to list of dropped dimmensions as I wasn't sure if this was what you wanted by 'implement pyspecdata module'

examples/proc_FIR.py
```diff
@@ -0,0 +1,107 @@
...
+                max=1.5 * s.max().item(),
+            ),
+            R_1=dict(value=0.8, min=0.01, max=100),
+        )
+        s.fit()
+        s_fit = s.eval(200)
+        psd.plot(s_fit, ax=ax_last, alpha=0.5)  # here, we plot the fit
+        #                                         together with the
+        #                                         table of integrals.
+        # TODO ☐: pyspecdata gives an example of how to show the fit equation on
```
wfkincaid, line 75: I moved your todo up to where I am trying to address it for the table of integrals plot, I didnt delete the 'to do' as I still have an error to address

PR #129, opened 2024-10-07
general proc for spincore data (including psd)

guinneak198:
a postproc for SpinCore PSDs and include running the conj on the t axis for spincore data since for some reason SC flips the axis

### DIFF
--- a/pyspecProcScripts/load_data.py
+++ b/pyspecProcScripts/load_data.py
@@ -671,8 +671,8 @@ def proc_spincore_generalproc_v1(
         #            the amplitude of signal in a single transient if we do
         #            this
         s[j] = -s[j]/4 + 1 # invert (b/c of conj), convert to cyc and add 1 full cyc
+        s[j] %= 1
         s.sort(j)
-        s[j] *= 4
     # {{{ always put the phase cycling dimensions on the outside
     neworder = [j for j in s.dimlabels if j.startswith("ph")]
     # }}}
@@ -688,11 +688,9 @@ def proc_spincore_generalproc_v1(
     if "ph_overall" in s.dimlabels:
         s.reorder("ph_overall")
         s["ph_overall"] = -s["ph_overall"]/4 + 1
+        s["ph_overall"] %= 1
         s.sort("ph_overall")
-        s["ph_overall"] *= 4
     # }}}
-    for k in s.get_prop("coherence_pathway"):
-        s.get_prop("coherence_pathway")[k] -= 1
     # {{{ apply the receiver response
     s.set_prop(
         "dig_filter",
--- a/pyspecProcScripts/load_data.py
+++ b/pyspecProcScripts/load_data.py
@@ -665,6 +665,7 @@ def proc_spincore_generalproc_v1(
     s, direct="t2", include_tau_sub=True, fl=None
 ):
     s.fromaxis(direct).run(np.conj)  # SC flips data in a weird way, this
+    s
     #             corrects for that
     if include_tau_sub:
         if "tau_us" in s.get_prop("acq_params").keys():
--- a/pyspecProcScripts/load_data.py
+++ b/pyspecProcScripts/load_data.py
@@ -323,11 +323,6 @@ def proc_spincore_diffph_SE_v2(s, fl=None):
     return s
 
 
-def proc_spincore_general_v1(s, fl=None):
-    s = proc_spincore_generalproc_v1(s, include_tau_sub=False, direct="t")
-    return s
-
-
 def proc_Hahn_echoph(s, fl=None):
     logging.debug("loading pre-processing for Hahn_echoph")
     s.reorder("t", first=True)
@@ -664,7 +659,7 @@ def proc_spincore_ODNP_v4(s, fl=None):
 def proc_spincore_generalproc_v1(
     s, direct="t2", include_tau_sub=True, fl=None
 ):
-    s.fromaxis(direct).run(np.conj)  # SC flips data in a weird way, this
+    s.run(np.conj)  # SC flips data in a weird way, this
     #                                  corrects for that
     if include_tau_sub:
         if "tau_us" in s.get_prop("acq_params").keys():
@@ -675,6 +670,9 @@ def proc_spincore_generalproc_v1(
         #            coordinates, signal in the coherence dimension will match
         #            the amplitude of signal in a single transient if we do
         #            this
+        s[j] = -s[j]/4 + 1 # invert (b/c of conj), convert to cyc and add 1 full cyc
+        s.sort(j)
+        s[j] *= 4
     # {{{ always put the phase cycling dimensions on the outside
     neworder = [j for j in s.dimlabels if j.startswith("ph")]
     # }}}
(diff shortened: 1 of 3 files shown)


pyspecProcScripts/load_data.py
```diff
@@ -323,6 +323,11 @@ def proc_spincore_diffph_SE_v2(s, fl=None):
     return s
 
 
+def proc_spincore_general_v1(s, fl=None):
+    s = proc_spincore_generalproc_v1(s, include_tau_sub=False, direct="t")
```
guinneak198, line 327: for preprocessing SC noise captures we don't want to shift the data by any tau value and the direct axis is t not t2

pyspecProcScripts/load_data.py
```diff
@@ -682,9 +691,13 @@ def proc_spincore_generalproc_v1(s, include_tau_sub=True, fl=None):
...
     # {{{ apply the receiver response
-    s /= s.fromaxis("t2").run(
-        lambda x: np.sinc(x / (s.get_prop("acq_params")["SW_kHz"] * 1e3))
+    s.set_prop(
+        "dig_filter",
+        s.fromaxis(direct).run(
+            lambda x: np.sinc(x / (s.get_prop("acq_params")["SW_kHz"] * 1e3))
+        ),
     )
+    s /= s.get_prop("dig_filter")
```
guinneak198, line 700: save digital filter as property in case we want to mess with it or access it in later steps

pyspecProcScripts/load_data.py
```diff
@@ -825,6 +838,7 @@ def hack_field_sweep_v4(s, fl=None):
     )
     return proc_spincore_generalproc_v1(s, fl=fl)
 
+
```
guinneak198, line 841: black

pyspecProcScripts/load_data.py
```diff
@@ -838,6 +852,7 @@ def hack_field_sweep_v4(s, fl=None):
     "proc_Hahn_echoph": proc_Hahn_echoph,
     "spincore_FID_nutation_v1": proc_FID_v1,
     "spincore_FID_nutation_v2": proc_FID_v1,
+    "spincore_general": proc_spincore_general_v1,
```
guinneak198, line 855: processing for SC noise PSD

pyspecProcScripts/load_data.py
jmfranck: The naming here is confusing... These are almost indistinguishable names at a glance 

Also, you could just use a lambda function in the look up table dictionary

pyspecProcScripts/load_data.py
```diff
@@ -656,11 +661,15 @@ def proc_spincore_ODNP_v4(s, fl=None):
     return proc_spincore_generalproc_v1(s, fl=fl)
 
 
-def proc_spincore_generalproc_v1(s, include_tau_sub=True, fl=None):
+def proc_spincore_generalproc_v1(
+    s, direct="t2", include_tau_sub=True, fl=None
```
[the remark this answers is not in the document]
jmfranck: Yes, this is the way to handle that aspect

pyspecProcScripts/load_data.py
```diff
@@ -656,11 +661,15 @@ def proc_spincore_ODNP_v4(s, fl=None):
...
+):
+    s.fromaxis(direct).run(np.conj)  # SC flips data in a weird way, this
```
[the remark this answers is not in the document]
jmfranck: this doesn't make sense -- why are are you using `fromaxis` to build a new thing, and then taking the complex conjugate of that?

pyspecProcScripts/load_data.py
jmfranck: again, just use a lambda function here rather than very similarly named function

jmfranck (changes requested):
missed the mark.  you need to apply complex conjugate to the data.

I think on slack you are pointing out that it changes the ordering of Δp when you do that.  Yes, it does, because the phases are labeled incorrectly.  They are now negative of what they were before -- so just change the phase coordinate axis to be negative of what it was before, and things should work out.

pyspecProcScripts/load_data.py
guinneak198: using lambda function in table dictionary now

pyspecProcScripts/load_data.py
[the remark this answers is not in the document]
guinneak198: has since been addressed

pyspecProcScripts/load_data.py
```diff
@@ -656,16 +656,26 @@ def proc_spincore_ODNP_v4(s, fl=None):
...
+    s.ft(direct, shift=True)
     for j in [k for k in s.dimlabels if k.startswith("ph")]:
         s.ft([j])  # if we have used cycles for the axis
         #            coordinates, signal in the coherence dimension will match
         #            the amplitude of signal in a single transient if we do
         #            this
+        s[j] = (
+            -s[j] / 4 + 1
+        )  # invert (b/c of conj), convert to cyc and add 1 full cyc
+        s[j] %= 1
```
guinneak198, line 676: modulo to set first ph to 0 so indexing of ph domain works

pyspecProcScripts/load_data.py
```diff
@@ -660,22 +660,39 @@ def proc_spincore_generalproc_v1(
...
+        s[j] = ((-s[j] + 4) % 4) / 4 # when we take the complex conjugate,
+        #                              that changes the phase of the
+        #                              phase cycle, as well, so we have
+        #                              to re-label the axis coordinates
+        #                              for the phase cycle to the
+        #                              negative of what they were
+        #                              before.  To keep things sane, we
+        #                              also apply phase wrapping to get
+        #                              positive numbers.
+        s.sort(j)
```
jmfranck, line 680: please note this change.  I think this should work, but I haven't tested it.  Probably makes sense to read the message below, then this one.

pyspecProcScripts/load_data.py
```diff
@@ -660,22 +660,39 @@ def proc_spincore_generalproc_v1(
...
+        # equivalent to what you have before.
+        # *HOWEVER* you seem to think that this is in radians.  It is not
+        # -- it's in units of quarter cycles, which is weird.
+        # If you look at the DCCT paper, you will see that we say that the
+        # units of phase here should be in units of cycles (which is this
+        # divided by four).
+        # Also, if you look at other postproc methods here, we are always
+        # doing that.  So I don't understand why you are doing this this
+        # way.
+        # s[j] = (-s[j] + 4) % 4
```
jmfranck, line 695: this is all part of the change above.

pyspecProcScripts/load_data.py
```diff
@@ -690,10 +707,6 @@ def proc_spincore_generalproc_v1(
     #     should be nothing outside that
     if "ph_overall" in s.dimlabels:
         s.reorder("ph_overall")
-        s["ph_overall"] = -s["ph_overall"] / 4 + 1
-        s["ph_overall"] %= 1
-        s["ph_overall"] *= 4
-        s.sort("ph_overall")
```
jmfranck: this conditional has nothing to do with the axis coordinates, and what you're doing here would be handled above.  (here's we're just saying that if there is a dimension called "overall', we want it on the outside)

jmfranck (changes requested):

pyspecProcScripts/load_data.py
guinneak198: no need to divide by 4 because we are already in cyc

PR #131, opened 2025-01-04
example to process gain taken from manuscript

guinneak198:


jmfranck:
I edited the first block.  Carefully review and note the mistakes that were made, and use this to fix the rest.  Also note the TODO comments.

guinneak198:
Incorporated comments from the receiver response PR and reviewed again

jmfranck:
there were still several things, so I ended up just editing myself.  Please review in detail, and make any tweak changes.  Note that this is basically a chunk of the code that ends up getting used in the other PR as well.

guinneak198:
reviewed and like the addition of the fit function (kicking myself for not doing this) I just moved the nodename and freq_kHz before the function since for this script the files do have the same frequencies (unlike the receiver response) so no point in doing that twice if it's not necessary

examples/process_gain.py
```diff
@@ -66,6 +53,19 @@ def determine_power_from_fit(filename, guessamp, guessph):
...
+    all_node_names = sorted(
+        psd.find_file(
+            re.escape(filename),
+            exp_type=data_dir,
+            return_list=True,
+        ),
+        key=lambda x: float(x.split("_")[1]),
+    )
+    frq_kHz = np.array([float(j.split("_")[1]) for j in all_node_names])
+    # }}}
```
guinneak198: this is part of the function and adds robustness incase the files DON'T have the same nodenames

examples/process_gain.py
```diff
@@ -82,9 +82,7 @@ def determine_power_from_fit(filename, guessamp, guessph):
         )
         # {{{ Fit to complex
         d = psd.lmfitdata(d)
-        d.functional_form = A * sp.exp(
-            1j * 2 * pi * omega * t + 1j * phi
-        )
+        d.functional_form = A * sp.exp(1j * 2 * pi * omega * t + 1j * phi)
```
guinneak198: seems to just be a black issue (probably forgot to change line length to 79 after editing the scripts in the manuscript to be 58)

examples/process_gain.py
```diff
@@ -105,9 +103,7 @@ def determine_power_from_fit(filename, guessamp, guessph):
 
 input_power = determine_power_from_fit(file1, 5e-2, 0.75)
 input_power.name("Input Power").set_plot_color("r")
-output_power = determine_power_from_fit(
-    file2, 15e-2, 0.75
-)
+output_power = determine_power_from_fit(file2, 15e-2, 0.75)
```
guinneak198: most likely same black issue

examples/process_gain.py
```diff
@@ -127,14 +123,6 @@ def determine_power_from_fit(filename, guessamp, guessph):
...
-    gain_dB.name("Gain").set_units("dB").set_plot_color(
-        "purple"
-    )
-    gain_dB.human_units(scale_data=True)
-    gain_spline = gain_dB.spline_lambda()
-    fl.plot(gain_spline(nu_fine))
-    fl.plot(gain_dB, "o")
+    gain_dB = 10 * np.log10(output_spline(nu_fine) / input_spline(nu_fine)) + attenuator_dB
+    gain_dB.name("Gain").set_units("dB").set_plot_color("purple")
+    fl.plot(gain_dB)
```
guinneak198: avoids creating a new spline just divide the two premade splines and call it a day

guinneak198: this also avoids the human_units thing I think

examples/process_gain.py
```diff
@@ -31,29 +31,16 @@
...
-frq_kHz = np.array(
-    [float(j.split("_")[1]) for j in all_node_names]
-)
 
 
 def determine_power_from_fit(filename, guessamp, guessph):
-    """Fit time capture to extract the amplitude ($V_{p}$) for each node within
-    the fed HDF5 file.
+    """Fit time-domain capture to extract the amplitude ($V_{p}$) for each node
+    within the HDF5 file.
```
guinneak198: "fed file" is weird wording. and time-domain is more accurate/exact

examples/process_gain.py
```diff
@@ -31,29 +31,16 @@
...
 
     Parameters
     ==========
     filename: str
-        Name of HDF5 file
+        Name of HDF5 file --- contains multiple nodes, named according to frequency
```
guinneak198: more descriptive since a big portion deals with iterating through the nodes

jmfranck: the main issue here was the RST formatting

examples/process_gain.py
jmfranck: yes, this is mostly what I'm saying in the comment above

examples/process_gain.py
jmfranck: yes, but I had thought this was b/c of omega/nu again -- I'm going to fix and push that.

examples/process_gain.py
jmfranck: yes, I'm running black here with standard line length --  we want to re-run when copying file to manuscript

examples/process_gain.py
jmfranck: yes -- but what motivated this is b/c I changed above to allow for different spacing, this also permits different spacing.

PR #132, opened 2025-01-06
Rec resp

guinneak198:


examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+control = psd.ndshape([len(control_frqs_kHz)], ["nu_test"]).alloc()
+control.setaxis("nu_test", control_frqs_kHz).set_units("nu_test", "Hz")
+# }}}
+for j, nodename in enumerate(file1_nodes):
+    rf_frq = control["nu_test"][j]
+    d = psd.find_file(
+        file1,
+        expno=nodename,
+        exp_type="ODNP_NMR_comp/noise_tests",
+    )
```
guinneak198, line 71: note I'm not using the re which you had started to use. This way is a. more consistent with the process_gain example and b. uses less lines of code also I find it's easier to follow

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+    carrier = (
+        d.get_prop("acq_params")["carrierFreq_MHz"] * 1e6
+    )  # $\nu_{RX,LO}$
+    if j == 0:
+        # Allocate and array that's shaped like the data acquired
+        # for one frequency but add axis to store the frequency of
+        # the test signal frequency
+        rec_data1 = (d.shape + ("nu_test", len(resp_frq_kHz))).alloc()
+        rec_data1.setaxis("t", d.getaxis("t")).set_units("t", "s")
+        rec_data1.setaxis("nu_test", resp_frq_kHz).set_units("nu_test", "Hz")
```
guinneak198, line 111: need to allocate shape in here to have the nScans axis (inspired by how we allocate for saving the psd data in the acquisition files

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+    different frequency (in kHz) which are fit to a complex function to extract
+    the power (in :math:`V^{2}`).
+    File2 contains the quadrature signal acquired on the receiver when the same
+    signal of File1 is injected into it. Each node pertains to signal with a
+    different frequency (in kHz) which are converted to a PSD. The power (in
+    :math:`dg^{2}`) is calculated from the peak of the convolved PSD.
+
+    The receiver response is then the ratio of :math:`dg^{2}(\\nu)` to
+    :math:`V^{2}(\\nu)` which is fit to a sinc function. The final plot shows
+    the square of the data with the fit.
```
guinneak198, line 17: willing to discuss all the math we have in the paper but figure this is simpler? let me know if I should include more details on why we do the PSD for the receiver data but not the oscilloscope

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+        rec_data = (d.shape + ("nu_test", len(nu_test_kHz))).alloc()
+        rec_data.setaxis("t", d.getaxis("t")).set_units("t", "s")
+        rec_data.setaxis("nu_test", nu_test_kHz).set_units("nu_test", "Hz")
+    rec_data["nu_test", j] = d
+    SW = str(d.get_prop("acq_params")["SW_kHz"]) # For labeling final plot
+rec_data.sort("nu_test")
+rec_data.rename("nScans", "capture")  # To be more consistent with the
+#                                        oscilloscope data rename the nScans
+#                                        dimension
+acq_time = np.diff(rec_data.getaxis("t")[np.r_[0, -1]]).item()
```
jmfranck, line 117: import `r_` from numpy.  Don't do `np.r_` or `np.pi` which is silly (these are intended as shorthand)

use `rec_data['t']`

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+rec_data.ft("t", shift=True)  # $dg\sqrt{s/Hz}$
+rec_data = abs(rec_data) ** 2  # $dg^{2}*s/Hz$
+rec_data.mean("capture")
+rec_data /= acq_time  # $dg^{2}/Hz$
+# Convolve using $\lambda_{G}$ specified above
+rec_data.convolve("t", lambda_G, enforce_causality=False)
+rec_data.ift("t")
+rec_data.run(np.conj)  # Empirically needed to give offset
+#                        that increases with field
+rec_data.ft("t")
```
jmfranck, line 129: this is super weird -- why would you not just do this before the ft, rather than moving to fourier domain and then back again.

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+# }}}
+# {{{ Calculate power of test signal (Eq. S3)
+rec_data["t"] += carrier
+rec_data.run(np.max, "t")
+rec_data *= (lambda_G / (2 * np.sqrt(np.log(2)))) * np.sqrt(2 * np.pi)
```
jmfranck, line 134: this is incorrect.  it should match the paper.  What I had in the paper before was equivalent to what you had, but I was simplifying my math when you were not, but this is now wrong (not equivalent to before).

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+# Convolve using $\lambda_{G}$ specified above
+rec_data.convolve("t", lambda_G, enforce_causality=False)
+rec_data.ift("t")
+rec_data.run(np.conj)  # Empirically needed to give offset
+#                        that increases with field
+rec_data.ft("t")
+# }}}
+# {{{ Calculate power of test signal (Eq. S3)
+rec_data["t"] += carrier
+rec_data.run(np.max, "t")
```
jmfranck, line 133: explain that this is finding the maximum.

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+rec_data *= (lambda_G / (2 * np.sqrt(np.log(2)))) * np.sqrt(2 * np.pi)
+# }}}
+# }}}
+# {{{ Calculate receiver response as function of frequencies
+# Make axis of finely spaced frequencies to feed to spline
+rec_data["nu_test"] -= carrier # center data at 0 MHz
```
jmfranck, line 139: edit comment -- not just centering -- you are actually converting the indirect axis to offset rather than $\nu_{test}$.  You should probably rename it at the same time.

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+from sympy import symbols
+import sympy as sp
+import re
+
+
+def for_plot(thisdata):
+    """Converts :math:`dg^{2}/V^{2}` to :math:`dg/\\mu V`
+    and sets labels accordingly for plotting"""
+    # Take sqrt for units of $dg/V$
+    thisdata.data = np.sqrt(thisdata.data)
```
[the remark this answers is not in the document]
jmfranck: rather add a line for the response where you `.run(sqrt)` to convert from dg^2 to dg

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+# }}}
+# {{{ Calculate receiver response as function of frequencies
+# Make axis of finely spaced frequencies to feed to spline
+rec_data["nu_test"] -= carrier # center data at 0 MHz
+Dnu = np.linspace(
+    (carrier) - (rec_data.getaxis("nu_test")[-1] / 2),
+    (carrier) + (rec_data.getaxis("nu_test")[-1] / 2),
+    len(rec_data.getaxis("nu_test")),
+)
+P_in = Pin_spline(Dnu)  # Generate spline of input powers
```
jmfranck, line 145: this is very confusing.  why not:

- generate the spline for the control (call it control or call it Pin -- don't mix and match) above at line 91
- also above, convert the nu_test axis to a `$\Delta\nu$` axis

ALSO

you seem to be acting like $\Delta\nu$ vs. $\nu_{test}$ is a difference in spacing (fine vs. coarse)

That's incorrect --> it's about whether it's a relative or absolute frequency.

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def for_plot(thisdata):
+    """Converts :math:`dg^{2}/V^{2}` to :math:`dg/\\mu V`
+    and sets labels accordingly for plotting"""
+    # Take sqrt for units of $dg/V$
+    thisdata.data = np.sqrt(thisdata.data)
+    # Units of $dg/\mu V$
+    thisdata /= 1e6
+    thisdata.rename("nu_test", r"$\Delta \nu$")
```
jmfranck, line 33: this should not happen here -- it should happen elsewhere when you subtract out the carrrier.

Btw, I usually do something like `Dnu_str = r'$\Delta \nu$'` to make my life easier (then just use `Dnu_str` (variable) as the dim name)

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+    omega=dict(value=1, min=-100, max=100),
+    delta_nu=dict(value=9e3, min=3, max=1e4),
+)
+f.fit()
+fit = f.eval()
+# }}}
+with psd.figlist_var() as fl:
+    fl.next("Receiver Response")
+    fl.plot(for_plot(dig_filter), "o")
+    fl.plot(for_plot(fit), color="red", alpha=0.5)
```
jmfranck, line 164: why are you not showing the image of the 2D response?

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+    len(rec_data.getaxis("nu_test")),
+)
+P_in = Pin_spline(Dnu)  # Generate spline of input powers
+dig_filter = rec_data / P_in
+# }}}
+# {{{ Fit receiver response
+A, omega, delta_nu, nu_test = symbols("A omega delta_nu nu_test", real=True)
+func_form = A * sp.sinc((2 * np.pi * (nu_test - omega)) / (delta_nu)) ** 2
+f = psd.lmfitdata(dig_filter)
+f.functional_form = func_form
```
jmfranck, line 152: this is extremely silly -- combine line 150 and 152

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+data_dir = "ODNP_NMR_comp/noise_tests"
+file1 = "240123_10mV_AFG_GDS_5mV_100MSPS_analytic.h5"
+file2 = "240117_afg_sc_10mV_3p9kHz_zoom.h5"
+# There are less nodes acquired for the control case (since we assume it's
+# relatively flat) so I need to separately define them
+file1_nodes = psd.find_file(
+    re.escape(file1),
+    exp_type=data_dir,
+    return_list=True,
+)
```
jmfranck, line 48: for after 48, here is an example that sorts node names so that they come in frequency order.   The int is key.
```
In [17]: a = ['bla_1','bla_10','bla_3','bla_2']

In [18]: sorted(a, key=lambda x: int(x.split('_')[1]))
Out[18]: ['bla_1', 'bla_2', 'bla_3', 'bla_10']
```
use this example to write code for here, and then do NOT use the `.sort()` elsewhere, since it's redundant.

jmfranck (changes requested):

examples/Receiver_Response.py
jmfranck: this will not be exactly consistent after implementing comments below

examples/Receiver_Response.py
guinneak198: if you want the simplified it should be 4 *np.sqrt... check convolve.py if you don't believe me I will fix this in the paper as well

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+# $\nu_{test}$
+nu_test_kHz = np.array([float(j.split("_")[1]) for j in file2_nodes])
+# {{{ Calculate input power (acquired on Oscilloscope)
+# {{{ Make empty nddata to drop the calculated $V^{2}$ into with corresponding
+#     frequency ($\nu$) output by AFG source, and set the frequencies based on
+#     the node names
+control = psd.ndshape([len(control_frqs_kHz)], ["nu_test"]).alloc()
+control.setaxis("nu_test", control_frqs_kHz).set_units("nu_test", "Hz")
```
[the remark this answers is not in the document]
guinneak198: I comment on this when I define nodenames

examples/Receiver_Response.py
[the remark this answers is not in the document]
guinneak198: fixed

examples/Receiver_Response.py
```diff
@@ -0,0 +1,245 @@
...
+            * 1e3
+        )
+    else:
+        these_frqs = np.array(
+            [int(j.split("_")[1]) for j in nodenames]
+        )
+    # Determine if nScans is a dimension in the acquired data
+    # If nScans is a dimension this indicates the data was acquired on the
+    # receiver with the intention to convert to a PSD in which case we will
+    # just pull the shape of one of the datasets and copy its shape first
```
guinneak198, line 87: this can be edited if you find a better way of wording this. basically for the receiver data theres the nScans dimension whereas the control is a single capture so for the control we can just make an empty nddata. but the receiver data requires this extra nscans so that we can eventually make the psd to get dg

examples/Receiver_Response.py
guinneak198: see function above

examples/Receiver_Response.py
```diff
@@ -0,0 +1,164 @@
...
+    return_list=True,
+)
+control_frqs_kHz = np.array([float(j.split("_")[1]) for j in file1_nodes])
+file2_nodes = psd.find_file(
+    re.escape(file2),
+    exp_type=data_dir,
+    return_list=True,
+)
+# $\nu_{test}$
+nu_test_kHz = np.array([float(j.split("_")[1]) for j in file2_nodes])
```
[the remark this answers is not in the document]
jmfranck: I didn't realize the node names had a decimal place in them.

examples/Receiver_Response.py
```diff
@@ -0,0 +1,245 @@
...
+    og_data = psd.find_file(
+        thisfile, expno=nodenames[0], exp_type=data_dir
+    )
+    if "nScans" in og_data.dimlabels:
+        ret_data = (
+            og_data.shape + ("nu_test", len(these_frqs))
+        ).alloc()
+        ret_data.setaxis(
+            "t", og_data.getaxis("t")
+        ).set_units("t", "s")
```
guinneak198, line 97: decided to keep this as t rather than label nu direct so it makes more sense when we calculate the psd first (acq time)

examples/Receiver_Response.py
```diff
@@ -0,0 +1,245 @@
...
+    # $\nu_{test}$
+    rec_data["nu_test"] = rec_data["nu_test"] - carrier
+    rec_data.rename("nu_test", Dnu_name)
+    rec_data.rename("t", nu_direct)
+    # {{{ Plot 2D pcolor
+    fig = plt.figure()
+    fl.next("2D plot", fig=fig)
+    the_2D = rec_data.C.run(np.real).pcolor(
+        cmap="BuPu", fig=fig, scale_independently=True
+    )
```
guinneak198, line 200: I manually played with the function pcolor to get the bar and labels on top for the paper fig. But here the positioning doesn't seem important for the color bar

examples/Receiver_Response.py
```diff
@@ -0,0 +1,198 @@
...
+        return_list=True,
+    ),
+    key=lambda x: int(x.split("_")[1]),
+)
+control_frqs = (
+    np.array(
+        [int(j.split("_")[1]) for j in control_nodes]
+    )
+    * 1e3
+)
```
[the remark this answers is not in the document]
jmfranck: discussed on slack

examples/Receiver_Response.py
jmfranck: I am not following this

examples/Receiver_Response.py
jmfranck: why do you have something called og_data -- that does not seem right at al.

examples/Receiver_Response.py
jmfranck: I agree

examples/Receiver_Response.py
```diff
@@ -0,0 +1,247 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+from numpy import r_, pi
+import pyspecdata as psd
+from sympy import symbols
+import sympy as sp
+import re
+
+Dnu_name = r"$\Delta\nu$"
+nu_direct = r"$\nu_{direct}$"
```
jmfranck, line 27: `nu_direct_name` or `nu_direct_str`

examples/Receiver_Response.py
```diff
@@ -0,0 +1,247 @@
...
+        Empty nddata in the appropriate shape with an extra 'nu_test'
+        dimension that contains the signal frequencies
+    """
+    nodenames = sorted(
+        psd.find_file(
+            re.escape(thisfile),
+            exp_type=data_dir,
+            return_list=True,
+        ),
+        key=lambda x: int(x.split("_")[1]),
```
jmfranck, line 71: you talk on slack about how you need to use `float` not `int` -- in that case, all these need to be `float`

examples/Receiver_Response.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+# {{{ Function to return the nodenames, test signal frequencies, and an
+#     empty nddata to drop calculated dg and V into
+def make_ndData(thisfile, nodes_in_kHz=True):
+    """Determine the nodenames of the file and sort them according to the
+    frequency of the test signal. Based on these nodenames a list of the
+    frequencies as integers are made. Anempty NDData in the shape of the
+    acquired data is created with an extra dimension 'nu_test' that
+    the sorted frequencies are placed inside that dimension.
```
jmfranck, line 41: this is just too wordy.  Don't overdramatize.  You look at the nodes in the hdf5 file, and you put them in an ordered list and use it to label the axis coordinates of an empty nddata that you also return.

examples/Receiver_Response.py
jmfranck: OK, in context, I see

examples/Receiver_Response.py
```diff
@@ -0,0 +1,247 @@
...
+            [int(j.split("_")[1]) for j in nodenames]
+        )
+    # Determine if nScans is a dimension in the acquired data
+    # If nScans is a dimension this indicates the data was acquired on the
+    # receiver with the intention to convert to a PSD in which case we will
+    # just pull the shape of one of the datasets and copy its shape first
+    og_data = psd.find_file(
+        thisfile, expno=nodenames[0], exp_type=data_dir
+    )
+    if "nScans" in og_data.dimlabels:
```
jmfranck, line 91: what you are doing here doesn't make sense -- you are always trying to return the nddata where you put the response -- and in both cases, that should be a 1D function that only has dimension `nu_test`

examples/Receiver_Response.py
```diff
@@ -0,0 +1,247 @@
...
+            "t", og_data.getaxis("t")
+        ).set_units("t", "s")
+    else:
+        ret_data = psd.ndshape(
+            [len(these_frqs)], ["nu_test"]
+        ).alloc()
+    ret_data.setaxis("nu_test", these_frqs).set_units(
+        "nu_test", "Hz"
+    )
+    return nodenames, these_frqs, ret_data
```
jmfranck, line 105: there should not be three arguments -- only two.  The list of node names, and the nddata for storing the response.

examples/Receiver_Response.py
```diff
@@ -0,0 +1,247 @@
...
+        these_frqs = (
+            np.array(
+                [int(j.split("_")[1]) for j in nodenames]
+            )
+            * 1e3
+        )
+    else:
+        these_frqs = np.array(
+            [int(j.split("_")[1]) for j in nodenames]
+        )
```
jmfranck, line 83: this is weirdly repetitive.  just do:
```
if nodes_in_kHz: these_frqs *= 1e3
```

jmfranck (changes requested):

examples/Receiver_Response.py
guinneak198: fixed

examples/Receiver_Response.py
guinneak198: no longer included in function

examples/Receiver_Response.py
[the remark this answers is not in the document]
guinneak198: okay using float now not int

jmfranck (commented):
see slack

examples/Receiver_Response.py
```diff
@@ -50,13 +57,14 @@ def get_freqs(thisfile):
     these_frqs: array
         Array of sorted signal frequencies as integers
     """
+    frq_extract_fn = lambda x: float(x.split("_")[1])
```
guinneak198, line 60: since this is repeated we can make this a variable that we call upon

jmfranck: it's not a variable

examples/Receiver_Response.py
```diff
@@ -81,20 +87,18 @@ def get_freqs(thisfile):
         exp_type=data_dir,
     )
     # {{{ fit signal in t domain to complex exponential
-    A, omega, phi, t = symbols("A omega phi t", real=True)
     f = psd.lmfitdata(d)
-    f.functional_form = A * sp.exp(1j * 2 * pi * omega * t + 1j * phi)
+    f.functional_form = A * sp.exp(1j * 2 * pi * nu * t + 1j * phi)
```
guinneak198, line 91: we are using Hz not rad/s so we should use nu not omega

guinneak198: and for consistency since we are calling it nu_test

examples/Receiver_Response.py
```diff
@@ -81,20 +87,18 @@ def get_freqs(thisfile):
...
+        nu=dict(
             value=control["nu_test"][j],
             min=control["nu_test"][j] - 1e4,
             max=control["nu_test"][j] + 1e4,
         ),
         phi=dict(value=0.75, min=-pi, max=pi),
     )
-    f.fit(use_jacobian=False)
-    f.eval()
+    f.fit()
```
guinneak198, line 101: we don't want eval because we aren't plotting we literally just want the fit parameters that come from the fit (which are a result of f.fit()) we don't need to run eval which runs the fit (using the parameters calculated from fit) along an axis.

examples/Receiver_Response.py
```diff
@@ -105,6 +109,8 @@ def get_freqs(thisfile):
 # }}}
 # }}}
 # {{{ Calculate dg
+# Get nodenames and frequencies recorded on receiver
+rec_nodenames, rec_frqs = get_freqs(file2)
```
guinneak198, line 113: this was moved down for organization so that the stuff that accommodate file2 are where we use file2

examples/Receiver_Response.py
```diff
@@ -173,22 +179,16 @@ def get_freqs(thisfile):
...
+        nu=dict(value=1, min=-100, max=100),
+        W=dict(value=9e3, min=3, max=1e4),
     )
     f.fit()
-    fit = f.eval()
     # }}}
     fl.next("Receiver Response")
     fl.plot(dig_filter, "o")
-    fl.plot(fit, color="red", alpha=0.5)
+    fl.plot(f.eval(), color="red", alpha=0.5)
```
guinneak198, line 194: no need to make a variable when we can just plot the thing we want

examples/Receiver_Response.py
guinneak198: we do want to use the jacobian because it is more efficient and accurate... somehow... the minimization using the jacobian is still pretty hand wavey in my head and I think its a matrix containing the partial derivatives (with respect to the parameters?) any references or good intro reads appreciated

jmfranck: yes, jacobian shows how the shape of the residual varies with changes in parameters -- so it can use the shape to figure out what to change.

examples/Receiver_Response.py
```diff
@@ -3,18 +3,22 @@
...
-    the :math:`V_{p}` (in μV).
-    File2 contains the quadrature signal acquired on the receiver when the same
+file1
+    contains the analytic signal acquired on the (GW Instek GDS) oscilloscope
+    directly captured from the rf source (GW Instek AFG). Each node pertains to
+    signal with a different frequency (in kHz) which are fit to a complex
+    function to extract the :math:`V_{p}` (in μV).
+
+file2
+    contains the quadrature signal acquired on the receiver when the same
```
guinneak198, line 13: since we are laying this out like a list (the ":") make each file arranged like we would the list of parameters

examples/Receiver_Response.py
```diff
@@ -3,18 +3,22 @@
...
     signal of File1 is injected into it. Each node pertains to signal with a
     different frequency (in Hz) which are converted to a PSD. The amplitude (in
     dg) is calculated from the peak of the convolved PSD.
 
-    The receiver response is then the ratio of :math:`dg(\\nu)` to
-    :math:`μV(\\nu)` which is fit to an absolute sinc function.
+The receiver response is then the ratio of :math:`dg(\\nu)` to
+:math:`μV(\\nu)` which is fit to an absolute sinc function.
```
guinneak198, line 19: this doesn't pertain to the files themselves so indent back out

examples/Receiver_Response.py
```diff
@@ -3,18 +3,22 @@
...
 """
+
```
guinneak198, line 21: not sure why black didn't catch this but there should always be a space between the bottom of a docstring and the code

examples/Receiver_Response.py
```diff
@@ -173,22 +179,16 @@ def get_freqs(thisfile):
...
     # {{{ Fit receiver response
-    A, omega, delta_nu, Dnu_name = symbols(
-        r"A omega delta_nu $\Delta\nu$", real=True
-    )
     f = psd.lmfitdata(dig_filter)
     # Fit to absolute sinc function
-    f.functional_form = A * abs(
-        sp.sinc((2 * pi * (Dnu_name - omega)) / (delta_nu))
-    )
+    f.functional_form = A * abs(sp.sinc((2 * pi * (Dnu_symb - nu)) / W))
```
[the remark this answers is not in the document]
guinneak198: omega to nu for the same reason as above

examples/Receiver_Response.py
jmfranck: this is an RST formatting issue

examples/Receiver_Response.py
guinneak198: what would you call it then? I get you are pulling this out for clarity (labeling gives a purpose) but if not a variable then what?

jmfranck: what would you call it?

guinneak198: so I chatted with chatgpt and it's saying this would be a named function, function reference, or callable (I like the latter term). The lambda function is a more concise way of defining a function because you don't need the def...: yada yada.

jmfranck: But, committing to an answer, what would you call it?  I really am very not interested in random junk that chat gpt spits out

guinneak198: lambda is the function. we are assigning it to what we are calling frq_extract_fn so that anytime we call frq_extract_fn we run the lambda function. so because we are assigning a variable frq_extract_fn a function then frq_extract_fn is now a function itself

jmfranck: lambda is not a function

guinneak198: moved to[ discussion on slack](https://jmfrancklab.slack.com/archives/CP9TAE89L/p1736441322125089)

guinneak198: frq_extract_fn = identifier

PR #133, opened 2025-01-10
edit 2D for receiver response

guinneak198:


guinneak198:
These were edits you made in the manuscript repo and said to add a PR for (I think just plotting the 2D of the PSDs vs sqrt or whatever was there (I forget). I don't want tot use the dataset from figure 12 since the data is so big it takes several minutes to run the example so I keep the dataset the same

examples/Receiver_Response.py
```diff
@@ -147,6 +147,11 @@ def get_freqs(thisfile):
     # Convolve using $\lambda_{G}$ specified above
     rec_data.convolve("t", lambda_G, enforce_causality=False)
     # }}}
+    # {{{ center data at 0 MHz thus converting to
+    #     $\Delta\nu$ rather than $\nu_{test}$
+    rec_data["nu_test"] -= carrier
+    rec_data.rename("nu_test", Dnu_name)
+    # }}}
```
guinneak198: This way the x axis for the 2D and 1D are both relative/consistent and make more sense together

guinneak198: also helps that the y axis (direct) is relative to the carrier so that when this axis is 0 it means it is outputting at the carrier

guinneak198 (commented):
renamed file to match other examples and show it is a processing script

Issue #44, opened 2021-06-11
auto bounds doesn't work correctly


jmfranck:

jmfranck:
- a rough manual choice outperforms the automatically chosen integration limits
- automatic routine doesn't sum over all coherence pathways
- explicitly propagated error doesn't match result of manual error exactly --> could be a good reason, but should be explained if so.

jmfranck:
Here's a figure showing the problem with the auto-integration -- notice how the auto bounds give a much larger error!

![Figure_2](https://user-images.githubusercontent.com/8870541/121747740-dc6e6800-cad5-11eb-8972-90f387d1862d.png)

aabeaton:
![210823_Gaussian_method](https://user-images.githubusercontent.com/31140790/130446817-e68c311a-fda1-4653-9f6f-752cbc2a5d66.png)
Here I have rerun this example using the new edits to integrate limits with a Gaussian convolution

aabeaton:
![210823_Lorentzian_method](https://user-images.githubusercontent.com/31140790/130446919-f700e12c-579a-4b67-b1bf-29059e8e49cb.png)
Here I have run with a Lorentzian convolution

aabeaton:
Here I have run the example with the new integrate_limits code using a [Gaussian convolution](https://jmfrancklab.slack.com/archives/CLMMYDD98/p1629721793000100?thread_ts=1629385912.002400&cid=CLMMYDD98) and here with a [Lorentzian convolution](https://jmfrancklab.slack.com/files/UKGAYDEKC/F02C1336J5R/210823_lorentzian_method.png). I think the Gaussian convolution looks better, but both are improvements from the plot above.

PR #137, opened 2025-02-13
Update int w error f

guinneak198:
updating to use a mask for calculating noise

### DIFF
--- a/examples/check_integration_error.py
+++ b/examples/check_integration_error.py
@@ -20,30 +20,42 @@ import numpy as np
 from numpy import r_
 from pyspecProcScripts import integral_w_errors, select_pathway
 from pylab import rcParams
+import sympy as s
+from collections import OrderedDict
 
 rcParams["image.aspect"] = "auto"  # for sphinx gallery
 # sphinx_gallery_thumbnail_number = 1
 
-t2 = psd.nddata(r_[0:1:1024j], "t2")
-vd = psd.nddata(r_[0:1:40j], "vd")
-ph1 = psd.nddata(r_[0, 2] / 4.0, "ph1")
-ph2 = psd.nddata(r_[0:4] / 4.0, "ph2")
+t2, vd, ph1, ph2 = s.symbols("t2 vd ph1 ph2")
 n_repeats = 10
 noise_bounds = (0, 200)  # seem reasonable to me
 signal_pathway = {"ph1": 0, "ph2": 1}
 # {{{ Generate fake IR dataset with multiple repeats
 # This generates fake clean_data w/ a T₂ of 0.2s amplitude of 21, just to pick
 # a random amplitude offset of 300 Hz, FWHM 10 Hz
-clean_data = (
+expression = (
     21
-    * (1 - 2 * np.exp(-vd / 0.2))
-    * np.exp(+1j * 2 * np.pi * 100 * t2 - t2 * 10 * np.pi)
+    * (1 - 2 * s.exp(-vd / 0.2))
+    * s.exp(+1j * 2 * s.pi * 100 * t2 - t2 * 10 * s.pi)
 )
-clean_data *= np.exp(signal_pathway["ph1"] * 1j * 2 * np.pi * ph1)
-clean_data *= np.exp(signal_pathway["ph2"] * 1j * 2 * np.pi * ph2)
+orderedDict = [
+    ("vd", psd.nddata(r_[0:1:40j], "vd")),
+    ("ph1", psd.nddata(r_[0, 2] / 4.0, "ph1")),
+    ("ph2", psd.nddata(r_[0:4] / 4.0, "ph2")),
+    ("t2", psd.nddata(r_[0:1:1024j], "t2")),
+]
+clean_data = psd.fake_data(
+    expression, OrderedDict(orderedDict), signal_pathway, scale=0
+)
+clean_data.reorder(["vd", "t2"], first=False)
 clean_data["t2":0] *= 0.5
 fake_data_noise_std = 2.0
 clean_data.reorder(["ph1", "ph2", "vd"])
+# Unitary FT concentrates signal into a single index/position along coherence
+# transfer pathway while preserve the vector norm so divide by sqrt(N_ph)
+clean_data /= np.sqrt(
+    psd.ndshape(clean_data)["ph1"] * psd.ndshape(clean_data)["ph2"]
+)
 # }}}
 # Allocate nddata to place the calculated integrals with error into
 all_results = psd.ndshape(clean_data) + (n_repeats, "repeats")
@@ -56,21 +68,20 @@ with psd.figlist_var() as fl:
         #                                     repeat
         # at this point, the fake data has been generated
         # {{{ Normalization
-        data.ft(["ph1", "ph2"])
-        data /= 0.5 * 0.25
-        data /= np.sqrt(psd.ndshape(data)["ph1"] * psd.ndshape(data)["ph2"])
+        # data /= np.sqrt(0.5 * 0.25)
         # }}}
-        data.ft("t2", shift=True)
+        data.ft("t2")  # , shift=True)
         dt = data.get_ft_prop("t2", "dt")
         # {{{ Vector-normalize the FT
         data /= np.sqrt(psd.ndshape(data)["t2"]) * dt
         # }}}
-        s_int, frq_slice = integral_w_errors(
-            data,
-            signal_pathway,
-            fl=fl,
-            return_frq_slice=True,
-        )
+        if j == 0:
+            s_int, frq_slice = integral_w_errors(
+                data,
+                signal_pathway,
+                fl=fl,
+                return_frq_slice=True,
+            )
         manual_bounds = select_pathway(data["t2":frq_slice], signal_pathway)
         N = psd.ndshape(manual_bounds)["t2"]
         df = manual_bounds.get_ft_prop("t2", "df")
--- a/examples/check_integration_error.py
+++ b/examples/check_integration_error.py
@@ -6,123 +6,120 @@ Generate a fake dataset of an inversion recovery with multiple repeats (φ
 × t2 × vd × repeats) w/ normally distributed random noise.
 Check that the following match:
 
-- integral w/ error (the canned routine :func:`~pyspecProcScripts.integral_w_errors`)
+- integral w/ error (the canned routine
+  :func:`~pyspecProcScripts.integral_w_errors`)
 - propagate error based off the programmed σ of the normal distribution
 - set the error bars based on the standard deviation (along the repeats
   dimension) of the *real* part of the integral
-- propagate error based off the variance of the noise in the inactive
-  coherence channels (do this manually inside this script -- should mimic
-  what :func:`~pyspecProcScripts.integral_w_errors` does)
+- propagate error based off the variance of the noise in the inactive coherence
+  channels (do this manually inside this script -- should mimic what
+  :func:`~pyspecProcScripts.integral_w_errors` does)
 """
-from pylab import *
-from pyspecdata import *
-from pyspecProcScripts import integrate_limits, integral_w_errors
+import pyspecdata as psd
+import numpy as np
+from numpy import r_
+from pyspecProcScripts import integral_w_errors, select_pathway
+from pylab import rcParams
 
+rcParams["image.aspect"] = "auto"  # for sphinx gallery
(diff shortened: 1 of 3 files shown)


examples/check_integration_error.py
```diff
@@ -6,123 +6,131 @@
...
+orderedDict = [
+    ("vd", psd.nddata(r_[0:1:40j], "vd")),
+    ("ph1", psd.nddata(r_[0, 2] / 4.0, "ph1")),
+    ("ph2", psd.nddata(r_[0:4] / 4.0, "ph2")),
+    ("t2", psd.nddata(r_[0:1:1024j], "t2")),
+]
+clean_data = psd.fake_data(
+    expression, OrderedDict(orderedDict), signal_pathway, scale=0
+)
+clean_data.reorder(["vd", "t2"], first=False)
```
guinneak198, line 50: no longer using manual generation of fake data since we have a function that does that for us

examples/check_integration_error.py
```diff
@@ -6,123 +6,131 @@
...
-fl = figlist_var()
-t2 = nddata(r_[0:1:1024j], "t2")
-vd = nddata(r_[0:1:40j], "vd")
-ph1 = nddata(r_[0, 2] / 4.0, "ph1")
-ph2 = nddata(r_[0:4] / 4.0, "ph2")
+t2, vd, ph1, ph2 = s.symbols("t2 vd ph1 ph2")
+n_repeats = 10
+noise_bounds = (0, 200)  # seem reasonable to me
 signal_pathway = {"ph1": 0, "ph2": 1}
-excluded_pathways = [(0, 0), (0, 3)]
```
guinneak198, line 30: no longer needed since the "excluded pathways" for determining noise is determined with a mask

examples/check_integration_error.py
```diff
@@ -6,123 +6,131 @@
...
-bounds = (0, 200)  # seem reasonable to me
-result = 0
-n_repeats = 100
-all_results = ndshape(clean_data) + (n_repeats, "repeats")
+# Unitary FT concentrates signal into a single index/position along coherence
+# transfer pathway while preserve the vector norm so divide by sqrt(N_ph)
+clean_data /= np.sqrt(
+    psd.ndshape(clean_data)["ph1"] * psd.ndshape(clean_data)["ph2"]
+)
+# }}}
```
guinneak198, line 59: pulling this from noise Verify.py, generate_fake_data unitarily FTs the data

examples/check_integration_error.py
```diff
@@ -6,123 +6,135 @@
...
+                signal_pathway,
+                fl=fl,
+                return_frq_slice=True,
+            )
+        else:
+            s_int, frq_slice = integral_w_errors(
+                data,
+                signal_pathway,
+                fl=None,
+                return_frq_slice=True,
```
guinneak198, line 87: prevents overlays of diagnostic plots for every repeat (we only need to see them once

examples/check_integration_error.py
```diff
@@ -6,123 +6,135 @@
...
             )
-        )
-        - set(excluded_pathways)
-        - set([(signal_pathway["ph1"], signal_pathway["ph2"])])
+        manual_bounds = select_pathway(data["t2":frq_slice], signal_pathway)
+        N = psd.ndshape(manual_bounds)["t2"]
+        df = manual_bounds.get_ft_prop("t2", "df")
+        manual_bounds.integrate("t2")
+        # N terms that have variance given by fake_data_noise_std**2 each
+        # multiplied by df
```
guinneak198, line 94: this was a comment carried over - if you think unnecessary or doesn't apply I can easily delete

examples/check_integration_error.py
```diff
@@ -6,123 +6,135 @@
...
+    fl.plot(
+        manual_bounds, ".", capsize=6, label=r"std from repeats", alpha=0.5
+    )
+    manual_bounds.set_error(np.sqrt(propagated_variance_from_inactive.data))
+    fl.plot(
+        manual_bounds,
+        ".",
+        capsize=6,
+        label=r"propagated from inactive std",
+        alpha=0.5,
```
guinneak198, line 139: this was all from original script

examples/integration_with_error.py
```diff
@@ -1,151 +1,124 @@
 """
-Check Integration
-=================
+Check Integration_w_error Function
+==================================
 
-Makes sure that automatically chosen integral bounds perform similar to or
+Generates a fake dataset of an inversion recovery w/ normally distributed
+random noise.
+Then checks that automatically chosen integral bounds perform similar to or
```
guinneak198, line 7: honestly, this seems to be doing a similar job to check_integration.py but figured we could discuss about deleting later and so I just updated this one

examples/integration_with_error.py
```diff
@@ -1,151 +1,124 @@
...
-data *= exp(signal_pathway["ph2"] * 1j * 2 * pi * ph2)
+orderedDict = [
+    ("vd", psd.nddata(r_[0:1:40j], "vd")),
+    ("ph1", psd.nddata(r_[0, 2] / 4.0, "ph1")),
+    ("ph2", psd.nddata(r_[0:4] / 4.0, "ph2")),
+    ("t2", psd.nddata(r_[0:1:1024j], "t2")),
+]
+data = psd.fake_data(
+    data_expression, OrderedDict(orderedDict), signal_pathway, scale=0
+)
```
guinneak198, line 43: again update to use pyspecdatas function to automatically generate fake data

examples/integration_with_error.py
```diff
@@ -1,151 +1,124 @@
...
+        manual_bounds.set_error(
+            np.sqrt(propagated_variance_from_inactive.data)
+        )
+        fl.plot(
+            manual_bounds,
+            ".",
+            capsize=6,
+            label="%s (inactive CT σ)\n$%4g\\rightarrow%4g$"
+            % ((thislabel,) + bounds),
+            alpha=0.5,
```
guinneak198, line 123: again from og script and looks okay

pyspecProcScripts/integral_w_error.py
```diff
@@ -1,46 +1,63 @@
...
+
+    def masked_var(x):
+        if var_has_imag:
+            return np.var(x[np.isfinite(x)], ddof=1) / 2
+        else:
+            return np.var(x[np.isfinite(x)], ddof=1)
+
+    return np.apply_along_axis(masked_var, axis, x)
+
+
```
guinneak198, line 30: In JF words: "I know how to write a masked mean or std only along 1 dimensions, so use numpy apply_along_axis to make it a function that works along 1 dimension of multidimensional data" (pulled from our scripts in t_domain)

pyspecProcScripts/integral_w_error.py
```diff
@@ -1,46 +1,63 @@
...
     s,
     sig_path,
-    error_path,
     cutoff=0.1,
     convolve_method="Gaussian",
     indirect="vd",
     direct="t2",
     fl=None,
     return_frq_slice=False,
+    has_imag=True,
```
guinneak198, line 40: see def below in parameter description (this is a kwarg for masked_var_multi)

pyspecProcScripts/integral_w_error.py
```diff
@@ -49,60 +66,39 @@ def integral_w_errors(
              not included in the signal pathway.
     """
     assert s.get_ft_prop(direct), "need to be in frequency domain!"
-    if convolve_method is not None:
-        kwargs = {"convolve_method": convolve_method}
-    else:
-        kwargs = {}
```
guinneak198, line 55: flake said these were never used anywhere so deleted

pyspecProcScripts/integral_w_error.py
```diff
@@ -72,37 +83,23 @@ def integral_w_errors(
...
+    noise = s.C
+    noise[direct:frq_slice] = np.nan
+    temp = psdpr.select_pathway(noise, sig_path)
+    temp.data[:] = np.nan
+    if fl is not None:
+        fl.next("frequency noise masked")
+        fl.image(noise)
+    noise.run(masked_var_multi, direct, has_imag)
+    for j in [k for k in s.dimlabels if k.startswith("ph")]:
+        noise.run(masked_mean_multi, j)
```
guinneak198, line 95: since function is only along 1 dimension I use a loop for the phase cycling dimension since some datasets have both ph1 and ph2.

pyspecProcScripts/integral_w_error.py
```diff
@@ -72,37 +83,23 @@ def integral_w_errors(
...
-    for j in range(len(error_path)):
-        # calculate N₂ Δf² σ², which is the variance of the integral (by error propagation)
-        # where N₂ is the number of points in the indirect dimension
-        s_forerror = select_pathway(s, error_path[j])
-        # previous line wipes everything out and starts over -- why not use
-        # collected_variance above, as I had originally set up --> part of
-        # issue #44
-        if j == 0:
-            N2 = psp.ndshape(s_forerror)[direct]
-        # mean divides by N₁ (indirect), integrate multiplies by Δf, and the
```
guinneak198, line 88: no longer needed since I am using a mask!

examples/check_integration_error.py
```diff
@@ -6,123 +6,135 @@
...
+  :func:`~pyspecProcScripts.integral_w_errors`)
 - propagate error based off the programmed σ of the normal distribution
 - set the error bars based on the standard deviation (along the repeats
   dimension) of the *real* part of the integral
-- propagate error based off the variance of the noise in the inactive
-  coherence channels (do this manually inside this script -- should mimic
-  what :func:`~pyspecProcScripts.integral_w_errors` does)
+- propagate error based off the variance of the noise in the inactive coherence
+  channels (do this manually inside this script -- should mimic what
+  :func:`~pyspecProcScripts.integral_w_errors` does)
```
guinneak198, line 16: for flake

examples/check_integration_error.py
```diff
@@ -6,123 +6,135 @@
...
-from pyspecProcScripts import integrate_limits, integral_w_errors
+import pyspecdata as psd
+import numpy as np
+from numpy import r_
+from pyspecProcScripts import integral_w_errors, select_pathway
+from pylab import rcParams
+import sympy as s
+from collections import OrderedDict
 
+rcParams["image.aspect"] = "auto"  # for sphinx gallery
```
guinneak198, line 26: pulled from other examples. This seemed to be missing

examples/check_integration_error.py
```diff
@@ -6,123 +6,135 @@
...
+    * s.exp(+1j * 2 * s.pi * 100 * t2 - t2 * 10 * s.pi)
+)
+orderedDict = [
+    ("vd", psd.nddata(r_[0:1:40j], "vd")),
+    ("ph1", psd.nddata(r_[0, 2] / 4.0, "ph1")),
+    ("ph2", psd.nddata(r_[0:4] / 4.0, "ph2")),
+    ("t2", psd.nddata(r_[0:1:1024j], "t2")),
+]
+clean_data = psd.fake_data(
+    expression, OrderedDict(orderedDict), signal_pathway, scale=0
```
guinneak198, line 48: no longer manually making fake data since we have a function to do this

examples/check_integration_error.py
```diff
@@ -6,123 +6,128 @@
...
+        # at this point, the fake data has been generated
+        data.ft("t2")
+        dt = data.get_ft_prop("t2", "dt")
+        # {{{ Vector-normalize the FT
+        data /= np.sqrt(psd.ndshape(data)["t2"]) * dt
+        # }}}
+        s_int, frq_slice = integral_w_errors(
+            data,
+            signal_pathway,
+            fl=fl if j == 0 else None,
```
guinneak198, line 79: No need to show diagnostics for every repeat so just show them for the first one

examples/check_integration_error.py
jmfranck: no -- the set logic is good and should be left in place.

jmfranck:
This is too complicated, and it's not clear what the benefits are.  Also, there is an issue #44 that would seem to be something we would want to take care of.  You really need to break what you want to do into reasonable PRs with a single purpose.  For example:

- replace the manual code generation with `fake_signal` (although it's not clear to me that this is something you want to do b/c the hard part of fake_signal is the frequency drift, which we are not using)
- replace the integration bound choosing with the same function we are using in `rough_table_of_integrals`

Note that I think part of the issue here is that the functions you are touching were not flake-compliant to begin with, so I have edited them on master to do so.  In the future, doing a PR for flake-compliance only (in keeping with above) is probably the way to go.

PR #140, opened 2025-03-05
Mod correl

guinneak198:
cleaned up code and reference DCCT paper according to [this](https://jmfrancklab.slack.com/files/UK6AJJPS8/F08FP99GYJK/image_from_ios.jpg)


pyspecProcScripts/correlation_alignment.py
```diff
@@ -81,20 +82,19 @@ def correl_align(
 
     logging.debug(psd.strm("Applying the correlation routine"))
     if avg_dim:
-        phcycdims = [j for j in s_orig.dimlabels if j.startswith("ph")]
-        indirect = set(s_orig.dimlabels) - set(phcycdims) - set([direct])
-        indirect = [j for j in s_orig.dimlabels if j in indirect]
-        avg_dim_len = len(s_orig.getaxis(avg_dim))
```
guinneak198, line 87: flake said this wasn't called anywhere so deleted

pyspecProcScripts/correlation_alignment.py
```diff
@@ -81,20 +82,19 @@ def correl_align(
...
+    for phnames in signal_pathway.keys():
+        assert not s_jk.get_ft_prop(phnames), (
+            str(phnames) + " must not be in the coherence domain"
         )
     signal_keys = list(signal_pathway)
     signal_values = list(signal_pathway.values())
-    ph_len = {j: psd.ndshape(s_orig)[j] for j in signal_pathway.keys()}
-    N = psd.ndshape(s_orig)[indirect_dim]
-    sig_energy = (abs(s_orig) ** 2).data.sum().item() / N
+    ph_len_dict = {j: psd.ndshape(s_jk)[j] for j in signal_pathway.keys()}
```
guinneak198, line 95: specify this is a dictionary containing the lengths of ph cycling dims

pyspecProcScripts/correlation_alignment.py
```diff
@@ -117,47 +117,55 @@ def correl_align(
...
     for my_iter in range(100):
         i += 1
         logging.debug(psd.strm("*** *** ***"))
         logging.debug(psd.strm("CORRELATION ALIGNMENT ITERATION NO. ", i))
         logging.debug(psd.strm("*** *** ***"))
-        s_orig.ift(direct)
-        s_copy = s_orig.C
+        s_jk.ift(direct)
+        # s_mn will become the signal with the additional Delta ph# dimension.
+        s_mn = s_jk.C
```
guinneak198, line 138: made a comment noting that we are now making s_mn. I make a copy though because later we will want to use s_jk still as is

pyspecProcScripts/correlation_alignment.py
```diff
@@ -117,47 +117,55 @@ def correl_align(
...
     nu_center = for_nu_center.mean(indirect_dim).C.argmax(direct)
+    # }}}
     logging.debug(psd.strm("Center frequency", nu_center))
     for my_iter in range(100):
         i += 1
         logging.debug(psd.strm("*** *** ***"))
         logging.debug(psd.strm("CORRELATION ALIGNMENT ITERATION NO. ", i))
         logging.debug(psd.strm("*** *** ***"))
-        s_orig.ift(direct)
-        s_copy = s_orig.C
```
guinneak198, line 135: dumb name

pyspecProcScripts/correlation_alignment.py
```diff
@@ -117,47 +117,55 @@ def correl_align(
...
+        for ph_name, ph_len in ph_len_dict.items():
+            for ph_index in range(ph_len):
+                s_mn["Delta%s" % ph_name.capitalize(), ph_index] = s_mn[
+                    "Delta%s" % ph_name.capitalize(), ph_index
+                ].run(
+                    lambda x, axis=None: np.roll(x, ph_index, axis=axis),
+                    ph_name,
+                )
+        # }}}
+        correl = s_mn.mean(indirect_dim).run(np.conj) * s_jk
```
guinneak198, line 168: Doc F's edit so no more looping (pushes sum inside)

pyspecProcScripts/correlation_alignment.py
```diff
@@ -170,15 +178,21 @@ def correl_align(
                     ax=ax_list[1],
                 )
                 ax_list[1].set_title("correlation function (t), \nafter apod")
+        # we want Δnu centered at 0 and don't care about the previous window
+        # selection
```
guinneak198, line 182: unsure if comment wnated but this is one of the things we discussed in meeting

pyspecProcScripts/correlation_alignment.py
```diff
@@ -170,15 +178,21 @@ def correl_align(
...
         correl.ft_new_startpoint(direct, "freq")
         correl.ft_new_startpoint(direct, "time")
+        # Part of convolution is multiplying in t domain and then zero-filling
+        # before re-FT
         correl.ft(direct, shift=True, pad=2**14)
-        for k, v in signal_pathway.items():
-            correl.ft(["Delta%s" % k.capitalize()])
+        # {{{ Isolate the correlation function for the signal pathway
```
guinneak198, line 188: I think this is what we are doing?

pyspecProcScripts/correlation_alignment.py
```diff
@@ -219,12 +236,13 @@ def correl_align(
             psd.strm(
                 "signal energy per transient (recalc to check that it stays"
                 " the same):",
-                (abs(s_copy**2).data.sum().item() / N),
+                (abs(for_Ecalc**2).data.sum().item() / N),
             )
         )
-
+        # {{{ Calculate energy difference from last shift to see if
+        #     there is any further gain to keep reiterating
```
guinneak198, line 243: again I think this is what we are doing so that we don't need to keep reiterating if we don't need to

pyspecProcScripts/correlation_alignment.py
```diff
@@ -199,16 +214,17 @@ def correl_align(
             )
         else:
             f_shift = correl.run(np.real).argmax(direct)
-        s_copy = s_orig.C
-        s_copy *= np.exp(-1j * 2 * np.pi * f_shift * s_copy.fromaxis(direct))
-        s_orig.ft(direct)
```
guinneak198, line 204: goes with above moving ift out of loop because this ft doesn't do anything

pyspecProcScripts/correlation_alignment.py
```diff
@@ -251,7 +269,4 @@ def correl_align(
             "after correlation\nph0 restored \nsig. energy=%g" % sig_energy
         )
         fl.pop_marker()
-    if avg_dim:
-        s_orig.chunk(avg_dim, [avg_dim, "power"], [avg_dim_len, -1])
-        s_orig.reorder(["ph1", avg_dim, "power", direct])
```
guinneak198, line 256: we couldn't figure out what this was doing here since s_orig was never returned and this outside of the loop??

guinneak198 (commented):
tested and passed examples/correlation_alignment_example.py

pyspecProcScripts/correlation_alignment.py
```diff
@@ -36,9 +36,10 @@ def correl_align(
 
     Parameters
     ==========
-    s_orig:  psd.nddata
+    s_jk:  psd.nddata
```
jmfranck, line 39: this should still be called s_orig, I think.  I think the strategy here was to make a copy (once!) so the original data is unmodified.

pyspecProcScripts/correlation_alignment.py
```diff
@@ -81,20 +82,19 @@ def correl_align(
...
-        avg_dim_len = len(s_orig.getaxis(avg_dim))
-        s_orig.smoosh(indirect)
-    for j in signal_pathway.keys():
-        assert not s_orig.get_ft_prop(j), (
-            str(j) + " must not be in the coherence domain"
+        phcycdims = [j for j in s_jk.dimlabels if j.startswith("ph")]
+        indirect = set(s_jk.dimlabels) - set(phcycdims) - set([direct])
+        indirect = [j for j in s_jk.dimlabels if j in indirect]
+        s_jk.smoosh(indirect)
+    for phnames in signal_pathway.keys():
```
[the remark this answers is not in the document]
jmfranck: I think leave this as `s_orig` --> see above

pyspecProcScripts/correlation_alignment.py
jmfranck: again, I do not see the benefit to renaming  at this point --> your naming of `s_jk` **absolutely** makes sense later, when you have what I have been calling the "left" and the "right".  But I see this as just complicating the diff for these earlier/setup functions.

I know that I do discourage copies where not needed, but I also really see the benefit here of an `s_orig`, which is *not to be modified* since that's the way the code is written.

pyspecProcScripts/correlation_alignment.py
```diff
@@ -117,47 +117,55 @@ def correl_align(
...
-            ph = np.ones(v)
-            s_copy *= psd.nddata(ph, "Delta" + k.capitalize())
-            s_copy.setaxis("Delta" + k.capitalize(), "#")
-        correl = s_copy * 0
-        for k, v in ph_len.items():
-            for ph_index in range(v):
-                s_copy["Delta%s" % k.capitalize(), ph_index] = s_copy[
-                    "Delta%s" % k.capitalize(), ph_index
-                ].run(lambda x, axis=None: np.roll(x, ph_index, axis=axis), k)
-        for j in range(1, N):
```
jmfranck, line 157: before making the other modifications to the code, did you check that the examples still ran with the mod that we have made together in the office? (removing the `range(1,N)` loop here?

pyspecProcScripts/correlation_alignment.py
jmfranck: do not make a copy here.  One of them is replicated across the Δφ dimension.  The multiplication that is used to generate the replicas creates a copy for you.

pyspecProcScripts/correlation_alignment.py
jmfranck: I don't know what this comment means out of context.  I am reading your comments in the conversation.  If you have more detailed comments, you can add to the code.

jmfranck: leave this comment.

pyspecProcScripts/correlation_alignment.py
jmfranck: I don't like the choice of loop variables here -- I would call these `ph_name` and `ph_val` or something like that.

pyspecProcScripts/correlation_alignment.py
```diff
@@ -170,15 +178,21 @@ def correl_align(
...
-        for k, v in signal_pathway.items():
-            correl.ft(["Delta%s" % k.capitalize()])
+        # {{{ Isolate the correlation function for the signal pathway
+        for sig_pathname, sig_pathval in signal_pathway.items():
+            correl.ft(["Delta%s" % sig_pathname.capitalize()])
             correl = (
-                correl["Delta" + k.capitalize(), v]
-                + correl["Delta" + k.capitalize(), 0]
+                correl["Delta" + sig_pathname.capitalize(), sig_pathval]
+                + correl["Delta" + sig_pathname.capitalize(), 0]
```
jmfranck, line 193: line 192 to 193 here are attempting to apply a mask and sum along the Δp_l dimension.

You should comment that the ft on line 190 transforms from Δφ to Δp_l

pyspecProcScripts/correlation_alignment.py
```diff
@@ -199,16 +214,17 @@ def correl_align(
...
+        # Calculate energy with shift applied
+        for_Ecalc = s_jk * np.exp(
+            -1j * 2 * np.pi * f_shift * s_jk.fromaxis(direct)
+        )
+        for_Ecalc.ft(direct)
         if my_iter == 0:
             logging.debug(psd.strm("holder"))
             if fl:
-                s_copy.reorder([direct], first=False)
+                for_Ecalc.reorder([direct], first=False)
```
jmfranck, line 225: I do not like this `for_Ecalc` .  It's actually just the aligned signal --> maybe `s_aligned`.  Also, you should be starting form `s_orig` here, in keeping with comments above.  It's possible that at the beginning `s_orig` gets an ift, to be ft'd again before the end *without being modified*.

pyspecProcScripts/correlation_alignment.py
jmfranck: yes!

pyspecProcScripts/correlation_alignment.py
jmfranck: hmm, it's possible that it's undoing things that were done to `s_orig` earlier.  Even if `s_orig` isn't returned, if you mess with its contents, they will change, so this could be undoing stuff that was done at the top.  Specifically, it looks like it's trying to undo a smoosh and reorder.

pyspecProcScripts/correlation_alignment.py
guinneak198: I'm assuming you mean changing `for_Ecalc = s_jk * np.exp(-1j * 2 * np.pi * f_shift * s_jk.fromaxis(direct)`) to `for_Ecalc = s_orig * np.exp(-1j * 2 * np.pi * f_shift * s_jk.fromaxis(direct))`. But:

1.  see the strm below where it says "signal energy per transient (recalc to check that it stays the same):"..." which means I don't think we want to use s_orig here. 
2. doing so causes the example to fail / result isn't aligned

guinneak198: but if you are talking about the dignostic, that runs still so I am going to go with that?

pyspecProcScripts/correlation_alignment.py
guinneak198: I made the smooshed version the copy so s_orig no longer gets modified so these are not necessary

pyspecProcScripts/correlation_alignment.py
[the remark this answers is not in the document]
guinneak198: now using s_orig for assert and only changing variable name

examples/correlation_alignment_example.py
```diff
@@ -64,15 +64,15 @@
...
+        fig.suptitle("Data Processing " + fl.basename)
         fl.next("Data Processing", fig=fig)
         data = psd.fake_data(
             expression, OrderedDict(orderedDict), signal_pathway
         )
         data.reorder([indirect, "t2"], first=False)
         data.ft("t2")
         data /= np.sqrt(psd.ndshape(data)["t2"]) * data.get_ft_prop("t2", "dt")
-        fl.DCCT(data, bbox=gs[0], title="Raw Data")
+        psd.DCCT(data, bbox=gs[0], fig=fig, title="Raw Data")
```
guinneak198, line 75: use updated DCCT format

examples/correlation_alignment_example.py
```diff
@@ -64,15 +64,15 @@
         fig = plt.figure(figsize=(11, 7))
         gs = plt.GridSpec(1, 4, figure=fig, wspace=0.4)
         # }}}
-        fig.suptitle(fl.basename)
+        fig.suptitle("Data Processing " + fl.basename)
```
jmfranck, line 67: what is the point of this?

examples/correlation_alignment_example.py
jmfranck: why is this better?

pyspecProcScripts/correlation_alignment.py
```diff
@@ -199,16 +222,17 @@ def correl_align(
...
+        # Calculate energy with shift applied
+        s_aligned = s_jk * np.exp(
+            -1j * 2 * np.pi * f_shift * s_jk.fromaxis(direct)
+        )
+        s_aligned.ft(direct)
         if my_iter == 0:
             logging.debug(psd.strm("holder"))
             if fl:
-                s_copy.reorder([direct], first=False)
+                s_orig.reorder([direct], first=False)
```
jmfranck, line 233: here, it does make sense to not reorder the original -- so I don't understand this.

pyspecProcScripts/correlation_alignment.py
jmfranck: I think that I'm very confused by this and we should sit down and review this together in the code

jmfranck (changes requested):
we need to review the code for this together.

examples/correlation_alignment_example.py
guinneak198: using fl doesn't seem to properly title the dcct (instead it uses the title used in the fl.next)

pyspecProcScripts/correlation_alignment.py
guinneak198: nixed this

jmfranck:
we have renamed and extensively commented to follow the paper, and marked up with TODOs for subsequent small PRs

PR #141, opened 2025-03-13
Mod correl

guinneak198:
first PR for changes nix avg_dim and indirect_dim


pyspecProcScripts/correlation_alignment.py
```diff
@@ -17,7 +17,9 @@ def to_percent(y, position):
 def correl_align(
     s_orig,
     tol=1e-4,
-    indirect_dim="indirect",
+    indirect_dim=None,
```
guinneak198, line 20: you said to keep these kwargs and just throw an error so I did

pyspecProcScripts/correlation_alignment.py
```diff
@@ -17,7 +17,9 @@ def to_percent(y, position):
...
+    repeat_dims=["indirect"],
+    non_repeat_dims=[""],
```
guinneak198, line 22: unless specified this is empty but length is 1

pyspecProcScripts/correlation_alignment.py
```diff
@@ -77,64 +79,29 @@ def correl_align(
...
-        indirect = [j for j in s_orig.dimlabels if j in indirect]
-        s_jk = s_orig.C.smoosh(indirect)  # this version ends up with
+    assert indirect_dim is None, (
+        "We updated the correlation function to no longer take indirect_dim as"
+        " a kwarg!"
+    )
+    assert avg_dim is None, (
+        "We updated the correlation function to no longer take avg_dim as a"
+        " kwarg!"
+    )
```
guinneak198, line 89: these are the errors thrown for the old kwargs

pyspecProcScripts/correlation_alignment.py
```diff
@@ -77,64 +79,29 @@ def correl_align(
...
+    phcycdims = [j for j in s_orig.dimlabels if j.startswith("ph")]
+    indirect = set(s_orig.dimlabels) - set(phcycdims) - set([direct])
+    assert len(indirect - set(repeat_dims) - set(non_repeat_dims)) == 0
+    indirect = [j for j in s_orig.dimlabels if j in indirect]
+    if len(indirect) > 1:
```
guinneak198, line 94: for example when indirect has 'vd' and 'nScans'

pyspecProcScripts/correlation_alignment.py
```diff
@@ -77,64 +79,29 @@ def correl_align(
...
+            indirect, "repeats"
+        )  # this version ends up with
         #                                   three dimensions
         #                                   (j=align_dim, k=phcyc, and
         #                                   direct nu) and is NOT conj
     else:
-        s_jk = s_orig.C  # even if there isn't an indirect to smoosh we will
+        s_jk = s_orig.C.rename(
+            indirect[0], "repeats"
+        )  # even if there isn't an indirect to smoosh we will
```
guinneak198, line 104: to be consistent with the above (when indirect length is more than 1) I just rename so I can keep the name throughout the rest of the code

pyspecProcScripts/correlation_alignment.py
```diff
@@ -147,9 +114,9 @@ def correl_align(
     signal_keys = list(signal_pathway)
     signal_values = list(signal_pathway.values())
     ph_len = {j: psd.ndshape(s_orig)[j] for j in signal_pathway.keys()}
-    N = psd.ndshape(s_orig)[indirect_dim]
+    N = psd.ndshape(s_jk)["repeats"]
```
guinneak198, line 117: N is the length of the indirect which now might be smooshed and include nScans - s_jk has smooshed dim

pyspecProcScripts/correlation_alignment.py
```diff
@@ -147,9 +114,9 @@ def correl_align(
...
     # TODO ☐: as noted below, this doesn't include the mask!
-    sig_energy = (abs(s_orig) ** 2).data.sum().item() / N
+    sig_energy = (abs(s_jk) ** 2).data.sum().item() / N
```
guinneak198, line 119: let me know if this is wrong

examples/correlation_alignment_example.py
```diff
@@ -105,7 +105,7 @@
         data.ift(list(signal_pathway.keys()))
         opt_shift, sigma, mask_func = psdpr.correl_align(
             data * mysgn,
-            indirect_dim=indirect,
+            repeat_dims=[indirect],
```
jmfranck, line 108: this doesn't make sense.  This should default to an empty list.

pyspecProcScripts/correlation_alignment.py
jmfranck: no -- default to an empty list

pyspecProcScripts/correlation_alignment.py
```diff
@@ -44,16 +46,16 @@ def correl_align(
         This data is not modified.
     tol:            float
                     Sets the tolerance limit for the alignment procedure.
-    indirect_dim:   str
-                    Name of the indirect dimension along which you seek to
-                    align
-                    the transients.
+    repeat_dims:   list
+                    List of names of the dimensions along which you seek to
+                    align.
```
jmfranck, line 51: you need to point out that these are dimensions along which signal is essentially **repeated** and therefore can be aligned.  You should also point out that `nScans` is included by default (and do this below), but that a common strategy is to flip the sign .... as I said in the comments.

pyspecProcScripts/correlation_alignment.py
```diff
@@ -44,16 +46,16 @@ def correl_align(
...
+    nonrepeat_dims: list
+                    These are indirect dimensions that we are not OK to align
+                    (e.g. the indirect dimension of a 2D COSY experiment).
```
jmfranck, line 54: this is fine -- add `list` --> `list (default [])`

pyspecProcScripts/correlation_alignment.py
```diff
@@ -77,64 +79,29 @@ def correl_align(
...
-    #         With both `repeat_dims` and
-    #         `non_repeat_dims` defined, we can use a set
-    #         expression like below to find the
-    #         indirect dimension (not phase
-    #         cycling and not direct) and to make
-    #         sure that they are all either (1)
-    #         nScans (2) specified as safe or (3)
-    #         listed in `non_repeat_dims`
-    #         Currently, I don't think you have
-    #         any `non_repeat_dims`
```
jmfranck, line 122: you are not using the explanations in the docstrings above.  By all means, put into your own words and make more concise, but there is a reason why I took the time to write these out.

pyspecProcScripts/correlation_alignment.py
```diff
@@ -77,64 +79,29 @@ def correl_align(
...
-        indirect = set(s_orig.dimlabels) - set(phcycdims) - set([direct])
-        # TODO ☐: see todo above -- you need to
-        #         first make repeat_dims and
-        #         non_repeat_dims lists of length
-        #         1 if they are given as strings
-        # assert len(indirect - set(repeat_dims) - set(non_repeat_dims)) = 0
-        indirect = [j for j in s_orig.dimlabels if j in indirect]
-        s_jk = s_orig.C.smoosh(indirect)  # this version ends up with
+    assert indirect_dim is None, (
+        "We updated the correlation function to no longer take indirect_dim as"
```
jmfranck, line 83: I *think* that `indirect_dim` roughly corresponds to `repeated_dims`, so you should say that

pyspecProcScripts/correlation_alignment.py
```diff
@@ -77,64 +79,29 @@ def correl_align(
...
+        " a kwarg!"
+    )
+    assert avg_dim is None, (
+        "We updated the correlation function to no longer take avg_dim as a"
+        " kwarg!"
+    )
+    phcycdims = [j for j in s_orig.dimlabels if j.startswith("ph")]
+    indirect = set(s_orig.dimlabels) - set(phcycdims) - set([direct])
```
jmfranck, line 91: include logic in case they pass the new arguments as a single string, rather than a list of strings.

pyspecProcScripts/correlation_alignment.py
```diff
@@ -77,64 +79,29 @@ def correl_align(
...
+    assert len(indirect - set(repeat_dims) - set(non_repeat_dims)) == 0
+    indirect = [j for j in s_orig.dimlabels if j in indirect]
```
jmfranck, line 93: there is no longer any `indirect` --> so this is not right

pyspecProcScripts/correlation_alignment.py
jmfranck: I think what you're looking for here is `repeat_dims`

jmfranck: and if you have `nScans` that needs to be added to the `repeat_dims` list, always.

pyspecProcScripts/correlation_alignment.py
jmfranck: this is good, but again:

1.  you mean `repeat_dims` not `indirect`
2. have a check above if `type(repeat_dims) is str` (then make a list of length 1)
3. add `nScans` (if there is an nScans dimension!!!)
4. after all that, but before the chunk, throw an error if there is not `repeat_dims`

pyspecProcScripts/correlation_alignment.py
jmfranck: should be fine

pyspecProcScripts/correlation_alignment.py
```diff
@@ -172,7 +139,7 @@ def correl_align(
     energy_diff = 1.0
     i = 0
     energy_vals = []
-    this_E = (abs(s_orig.C.sum(indirect_dim)) ** 2).data.sum().item() / N**2
+    this_E = (abs(s_jk.C.sum("repeats")) ** 2).data.sum().item() / N**2
```
jmfranck, line 142: add a comment -- this divides by an extre N compared to above, because if the signals along repeats are the same, then the energy of the resulting sum should increase by N (vs. taking the square and then summing, which is what we do above)

pyspecProcScripts/correlation_alignment.py
```diff
@@ -401,4 +368,12 @@ def correl_align(
...
     if fl is not None:
         fl.pop_marker()
+    # {{{ Make sure returned f_shift is the same ndshape that the
+    #     data was originally fed in - meaning we must chunk it back
+    #     into the original repeat_dims or rename back to the original
+    #     indirect name
+    if len(indirect) > 1:
+        f_shift.chunk("repeats", indirect)
+    else:
+        f_shift.rename("repeats", indirect[0])
```
jmfranck, line 378: this is very good, but is missing a `# }}}`

same comment about the change of name (indirect vs. repeat_dims) as above

jmfranck (changes requested):
done reviewing -- made indiv comments

pyspecProcScripts/correlation_alignment.py
guinneak198: what is the point of defining repeat_dims as a kwarg if we just overwrite it here (`repeat_dims = set(s_orig.dimlabels)-set(phcycdims)-set([direct])`?

examples/correlation_alignment_example.py
guinneak198: are you saying don't feed a repeat_dims argument? - mind, this is the example

pyspecProcScripts/correlation_alignment.py
guinneak198: renamed to repeats

pyspecProcScripts/correlation_alignment.py
guinneak198: checked

pyspecProcScripts/correlation_alignment.py
```diff
@@ -44,16 +46,16 @@ def correl_align(
         This data is not modified.
     tol:            float
                     Sets the tolerance limit for the alignment procedure.
-    indirect_dim:   str
-                    Name of the indirect dimension along which you seek to
-                    align
-                    the transients.
+    repeat_dims:   list
```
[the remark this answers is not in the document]
guinneak198: checked

pyspecProcScripts/correlation_alignment.py
guinneak198: checked

pyspecProcScripts/correlation_alignment.py
guinneak198: checked

pyspecProcScripts/correlation_alignment.py
guinneak198: copied comments into docstring

pyspecProcScripts/correlation_alignment.py
guinneak198: check

pyspecProcScripts/correlation_alignment.py
guinneak198: checked - now have assert to check that repeat_dims is a list

pyspecProcScripts/correlation_alignment.py
guinneak198: okay now using safe_repeat_dims

pyspecProcScripts/correlation_alignment.py
```diff
@@ -172,7 +158,11 @@ def correl_align(
...
     energy_vals = []
-    this_E = (abs(s_orig.C.sum(indirect_dim)) ** 2).data.sum().item() / N**2
+    # below we divide by an extra N because if the signals along the repeats
+    # are the same, then the energy of the resulting sum should increase by N
+    # (vs taking the square and summing which is what we do for calculating the
+    # sig_energy above)
+    this_E = (abs(s_jk.C.sum("repeats")) ** 2).data.sum().item() / N**2
     energy_vals.append(this_E / sig_energy)
     last_E = None
     # TODO ☐: the mask needs to be separated into
```
guinneak198, line 168: this is the comment you were talking about earlier in our conversation where you basically lay the ground work for the next PR which makes the mask a function

pyspecProcScripts/correlation_alignment.py
guinneak198: - 2 is simplified to just checking it's a list and if not it says it needs to be a list of strings

- addressed 3 around lines 112

- 4 is now incorporated

pyspecProcScripts/correlation_alignment.py
jmfranck: that's not the correct solution -- you should also allow for it to be a single string (same for nonrepeat...)

pyspecProcScripts/correlation_alignment.py
jmfranck: hmm -- yes, I was noticing that as I read as well as I was editing.  It's better to eliminate the `indirect` variable -- it's involved in set math, but you can combine the set math.  Though you might need to add a comment to explain, that is still a bit cleaner than spinning up a variable we use once.

examples/broken/correlation_alignment_example.py
```diff
@@ -16,6 +18,71 @@
...
+    for_mask.ift(list(signal_pathway))
+    # }}}
+    return for_mask * frq_mask
+
+
+def Delta_p_mask(s, signal_pathway):
+    """ Filters out all but the signal pathway and the "ph1":0 or
+    {'ph1':0,'ph2':0} pathways (depending on which experiment below is used).
+    Note this serves as an example function and other filter functions could
+    alternatively be used"""
```
guinneak198, line 77: Even though I'm using exactly what was in the correl align I note this is an example and the function could really filter any pathway

jmfranck: makes sense

examples/broken/correlation_alignment_example.py
```diff
@@ -16,6 +18,71 @@
...
+    assert s.get_ft_prop(direct), "You must be in the frequency domain!"
+    for phnames in signal_pathway.keys():
+        assert not s.get_ft_prop(phnames), (
+            str(phnames) + "must NOT be in coherence domain!"
+        )
+    signal_keys = list(signal_pathway)
+    signal_values = list(signal_pathway.values())
+    # we want to leave the original s unchanged and return a copy
+    for_mask = s.C
+    for_mask.ft(list(signal_pathway))
```
jmfranck, line 54: you should not FT.  You should just assume the data will be provided in frequency (and coherence) domain

examples/broken/correlation_alignment_example.py
```diff
@@ -16,6 +18,71 @@
 from collections import OrderedDict
 from numpy.random import seed
 
+
+# {{{ Define the frequency mask function and the ph cyc mask
+def frq_mask(s, signal_pathway, direct="t2", indirect="repeats"):
```
jmfranck, line 23: remember the intention is that this would be included *every* time you use the correlation function so:
- it's too long/complicated
- it should not have kwargs -- it's OK to assume you know the name of the direct dimension.
I don't understand why you can't so something like this:
```
def frq_mask(s):
    s *= # use s.fromaxis('t2') to construct the mask **which has only one dimension**
    return s
```

examples/broken/correlation_alignment_example.py
```diff
@@ -16,6 +18,71 @@
...
+    frq_mask = np.exp(
+        -((for_mask.fromaxis(direct) - nu_center) ** 2) / (2 * 20.0**2)
+    )
+    # }}}
+    for_mask.ift(list(signal_pathway))
+    # }}}
+    return for_mask * frq_mask
+
+
+def Delta_p_mask(s, signal_pathway):
```
jmfranck, line 67: same comment as above -- assume it's in the coherence domain (and add the code to correlation so that it is, in both cases)

examples/broken/correlation_alignment_example.py
guinneak198: image
it is not always 1 dimension

guinneak198: but did simplify

examples/broken/correlation_alignment_example.py
```diff
@@ -16,6 +18,71 @@
...
+    # }}}
+    # {{{ Make mask using the center frequency and sigma (whose estimate here
+    #     is 20)
+    frq_mask = np.exp(
+        -((for_mask.fromaxis(direct) - nu_center) ** 2) / (2 * 20.0**2)
+    )
+    # }}}
+    for_mask.ift(list(signal_pathway))
+    # }}}
+    return for_mask * frq_mask
```
jmfranck, line 64: OK, I've thought about this a bit, and the square root of the frequency domain mask should multiply s_j,k (the one that does not have the extra dimension, and that does not have the complex conjugate)

examples/broken/correlation_alignment_example.py
jmfranck: I've reviewed in the code, so you should look at the TODOs there

examples/broken/correlation_alignment_example.py
```diff
@@ -16,6 +18,71 @@
...
+    #     function
+    # {{{ find center frequency
+    for j in range(len(signal_keys)):
+        signal = for_mask[signal_keys[j], signal_values[j]].C
+    nu_center = signal.mean(indirect).C.argmax(direct)
+    # }}}
+    # {{{ Make mask using the center frequency and sigma (whose estimate here
+    #     is 20)
+    frq_mask = np.exp(
+        -((for_mask.fromaxis(direct) - nu_center) ** 2) / (2 * 20.0**2)
```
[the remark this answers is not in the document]
jmfranck: it seems better to focus on the comments in the code

examples/broken/correlation_alignment_example.py
jmfranck: these comments seem to be indicating the same misunderstanding of how pyspecdata works that we discussed recently.

jmfranck: (when there is a dimension mismatch, it deals naturally with the extra dimensions)

guinneak198: I think this is resolved

examples/broken/correlation_alignment_example.py
guinneak198: I think we've resolved this

pyspecProcScripts/correlation_alignment.py
```diff
@@ -44,97 +55,108 @@ def correl_align(
...
+    )
+    assert avg_dim is None, (
+        "We updated the correlation function to no longer take avg_dim as a"
+        " kwarg!"
+    )
+    # }}}
+    if isinstance(non_repeat_dims, str):
+        non_repeat_dims = [non_repeat_dims]
+    if isinstance(repeat_dims, str):
+        repeat_dims = [repeat_dims]
```
guinneak198, line 112: in case fed as str instead of lists

jmfranck:
I reviewed in the code and added new todos.

examples/correlation_alignment_example.py
```diff
@@ -16,6 +16,44 @@
 from collections import OrderedDict
 from numpy.random import seed
 
+
+# {{{ Define the frequency mask function and the ph cyc mask
+def frq_mask(s, sigma=20.0):
+    """Note that we assume that our mask is a product of a frequency-domain and
+    a coherence-domain function.  This multiplies by the square root of the
+    frequency-domain part"""
+    s.ft(list(s.get_prop("coherence_pathway")))
```
guinneak198, line 25: note in f domain and ct domain

guinneak198: previously I was in ph cyc domain which was causing a lot of issues

pyspecProcScripts/correlation_alignment.py
```diff
@@ -44,161 +55,157 @@ def correl_align(
...
         direct
     ), "direct dimension must be in the frequency domain"
-    signal_keys = list(signal_pathway)
-    signal_values = list(signal_pathway.values())
     ph_len = {j: psd.ndshape(s_orig)[j] for j in signal_pathway.keys()}
-    N = psd.ndshape(s_orig)[indirect_dim]
-    # TODO ☐: as noted below, this doesn't include the mask!
-    sig_energy = (abs(s_orig) ** 2).data.sum().item() / N
+    N = s_jk.shape["repeats"]
+    sig_energy = (abs(frq_mask_fn(s_jk)) ** 2).data.sum().item() / N
```
guinneak198, line 177: s_jk has the mask applied now!

pyspecProcScripts/correlation_alignment.py
```diff
@@ -44,161 +55,157 @@ def correl_align(
...
-    energy_vals.append(this_E / sig_energy)
+    # E_of_avg is the energy calculated from the averaged signal
+    # (vs. sig_energy above, which is the energy of the *un*averaged
+    # signal.
+    # We divide by an extra N because if the signals along the repeats
+    # are the same, then the energy of the resulting sum should increase by N
+    # (vs taking the square and summing which is what we do for calculating the
+    # sig_energy above)
+    E_of_avg = (
+        abs(s_jk.C.sum("repeats")) ** 2
```
guinneak198, line 205: remember mask was applied above

pyspecProcScripts/correlation_alignment.py
```diff
@@ -343,62 +322,49 @@ def correl_align(
...
-                human_units=False,
-            )
-            ax_list[3].set_title("after correlation")
+            psd.DCCT(s_jk, fig, title="After correlation", bbox=gs[0, 3])
         logging.debug(
             psd.strm(
                 "signal energy per transient (recalc to check that it stays"
                 " the same):",
-                (abs(s_aligned**2).data.sum().item() / N),
+                (abs(s_jk ** 2).data.sum().item() / N),
```
guinneak198, line 336: remember mask is applied above

pyspecProcScripts/correlation_alignment.py
```diff
@@ -44,161 +55,158 @@ def correl_align(
...
         direct
     ), "direct dimension must be in the frequency domain"
-    signal_keys = list(signal_pathway)
-    signal_values = list(signal_pathway.values())
     ph_len = {j: psd.ndshape(s_orig)[j] for j in signal_pathway.keys()}
-    N = psd.ndshape(s_orig)[indirect_dim]
-    # TODO ☐: as noted below, this doesn't include the mask!
-    sig_energy = (abs(s_orig) ** 2).data.sum().item() / N
+    N = s_jk.shape["repeats"]
+    s_jk = frq_mask_fn(s_jk)
```
guinneak198, line 177: only need to apply mask once to s_jk

pyspecProcScripts/correlation_alignment.py
```diff
@@ -265,20 +260,33 @@ def correl_align(
...
+        # {{{ this applies the Fourier transform from Δφ to Δpₗ
+        #     that is found inside the left square bracket of eq. 29.
+        #     The paper implies a sum along Δpₗ terms as in eq. 28, but
+        #     doesn't actually show them. (Simultaneously maximize all
+        #     coherence pathways that are not masked)
+        #     Note that because only the left square bracket depends on
+        #     Δpₗ, we can apply the coherence mask here, before
+        #     multiplication, in order to decrease the dimensionality of
+        #     the correlation function.
+        s_leftbracket.run(np.conj)
```
guinneak198, line 272: I'm in the t domain here!

pyspecProcScripts/correlation_alignment.py
jmfranck: note how I changed things around a bit.

pyspecProcScripts/correlation_alignment.py
jmfranck: again -- note my changes.

jmfranck (changes requested):
reviewed in code.  Please note that the comments you are adding on github are mostly just slowing down the review, so I wouldn't do that.

jmfranck (commented):
getting rid of self-requested review -- ignore this comment.

guinneak198:
You said to make a comment that there is one remaining to do that you wanted to save

jmfranck (changes requested):
I reviewed.  I added two TODOs that should be easy to implement.

Note that the alignment actually doesn't care (or know about) the signal pathway.  It only cares about the CT-domain mask.

As far as the alignment is concerned, ANY unmasked CT pathway is the same as any other.

PR #143, opened 2025-03-17
New err fun

guinneak198:
pertaining to[ this](https://jmfrancklab.slack.com/archives/C0811PEGU5B/p1741805260965559)

### DIFF
--- a/pyspecProcScripts/calc_error.py
+++ b/pyspecProcScripts/calc_error.py
@@ -1,7 +1,34 @@
 import pyspecdata as psd
 from .simple_functions import select_pathway
+import numpy as np
 
 
+# {{{ Functions to calculate var and mean of nan-masked data
+#     Pulled from time_domain_noise.py example
+def masked_mean_multi(x, axis=None):
+    "Calculate the mean of nan-masked data on a 1D axis"
+    assert axis is not None
+
+    def masked_mean(x):
+        return np.mean(x[np.isfinite(x)])
+
+    return np.apply_along_axis(masked_mean, axis, x)
+
+
+def masked_var_multi(x, var_has_imag=False, axis=None):
+    "Calculates the variance along a 1D axis"
+    assert axis is not None
+
+    def masked_var(x):
+        if var_has_imag:  # take average of variance along real and imag
+            return np.var(x[np.isfinite(x)], ddov=1) / 2
+        else:
+            return np.var(x[np.isfinite(x)], ddof=1)
+
+    return np.apply_along_axis(masked_var, axis, x)
+
+
+# }}}
 def calc_error(
     s,
     frq_slice,
@@ -55,6 +82,7 @@ def calc_error(
 
     s = s[direct:frq_slice]
     df = s.get_ft_prop(direct, "df")
+    N = len(s[direct])
     collected_variance = (
         psd.ndshape(
             [psd.ndshape(s)[indirect], len(error_path)], [indirect, "pathways"]
@@ -72,11 +100,8 @@ def calc_error(
             # (by error propagation) where N₂ is the number of points in
             # the indirect dimension
             s_forerror = select_pathway(s, error_path[j])
-            # previous line wipes everything out and starts over -- why not use
-            # collected_variance above, as I had originally set up --> part of
-            # issue #44
             if j == 0:
-                N2 = psd.ndshape(s_forerror)[direct]
+                Nshape = psd.ndshape(s_forerror)[direct]
             # mean divides by N₁ (indirect), integrate multiplies by Δf, and
             # the mean sums all elements (there are N₁N₂ elements)
             s_forerror -= s_forerror.C.mean_all_but([indirect, direct]).mean(
@@ -86,9 +111,21 @@ def calc_error(
                 [direct, indirect]
             ).mean(direct)
             s_forerror *= df**2  # Δf
-            s_forerror *= N2
+            s_forerror *= Nshape
             collected_variance["pathways", j] = s_forerror
         collected_variance = psd.sqrt(
             collected_variance.sum("pathways") / len(error_path)
         )
-        return collected_variance
+    else:
+        collected_variance = s.C  # so we don't alter s
+        collected_variance[direct:frq_slice] = np.nan
+        temp = select_pathway(collected_variance, signal_pathway)
+        temp.data[:] = np.nan
+        if fl is not None:
+            fl.next("Frequency Noise")
+            fl.image(collected_variance)
+        collected_variance.run(masked_var_multi, direct)
+        for j in [k for k in s.dimlabels if k.startswith("ph")]:
+            collected_variance.run(masked_mean_multi, j)
+        collected_variance = psd.sqrt(collected_variance * df**2 * N)
+    return collected_variance
--- a/pyspecProcScripts/calc_error.py
+++ b/pyspecProcScripts/calc_error.py
@@ -1,5 +1,4 @@
 import pyspecdata as psd
-import numpy as np
 from .simple_functions import select_pathway
 
 
@@ -11,7 +10,7 @@ def calc_error(
     direct="t2",
     indirect="nScans",
     use_mask=False,
-    fl = None,
+    fl=None,
 ):
     """Calculates the propagation of error for the given signal.
 
@@ -55,7 +54,7 @@ def calc_error(
     """
 
     s = s[direct:frq_slice]
-    df = s.get_ft_prop(direct,"df")
+    df = s.get_ft_prop(direct, "df")
     collected_variance = (
         psd.ndshape(
             [psd.ndshape(s)[indirect], len(error_path)], [indirect, "pathways"]
@@ -63,7 +62,7 @@ def calc_error(
         .alloc()
         .setaxis("pathways", error_path)
     )
-    if use_mask == False:
+    if use_mask is False:
         assert error_path is not None, (
             "If you don't want to use the mask you need to tell me what"
             " pathways to use for calculating the noise!"
@@ -78,8 +77,8 @@ def calc_error(
             # issue #44
             if j == 0:
                 N2 = psd.ndshape(s_forerror)[direct]
-            # mean divides by N₁ (indirect), integrate multiplies by Δf, and the
-            # mean sums all elements (there are N₁N₂ elements)
+            # mean divides by N₁ (indirect), integrate multiplies by Δf, and
+            # the mean sums all elements (there are N₁N₂ elements)
             s_forerror -= s_forerror.C.mean_all_but([indirect, direct]).mean(
                 direct
             )
--- a/pyspecProcScripts/calc_error.py
+++ b/pyspecProcScripts/calc_error.py
@@ -1,0 +1,35 @@
+import pyspecdata as psd
+import numpy as np
+from .simple_functions import select_pathway
+
+def calc_error(s,int_bounds=(-500,500), signal_pathway = {"ph1":1}, direct= "t2",indirect="nScans",use_mask = False, error_path = {}):
+    f = s.getaxis(direct)
+    df = f[1]-f[0]
+    collected_variance = psd.ndshape(
+        [psd.ndshape(s)[indirect], len(error_path)], [indirect, "pathways"]
+    ).alloc().setaxis("pathways",error_path)
+    if use_mask == False:
+        assert error_path is not None, "If you don't want to use the mask you need to tell me what pathways to use for calculating the noise!"
+        for j in range(len(error_path)):
+            # calculate N₂ Δf² σ², which is the variance of the integral
+            # (by error propagation) where N₂ is the number of points in
+            # the indirect dimension
+            s_forerror = select_pathway(s, error_path[j])
+            # previous line wipes everything out and starts over -- why not use
+            # collected_variance above, as I had originally set up --> part of
+            # issue #44
+            if j == 0:
+                N2 = psd.ndshape(s_forerror)[direct]
+            # mean divides by N₁ (indirect), integrate multiplies by Δf, and the
+            # mean sums all elements (there are N₁N₂ elements)
+            s_forerror -= s_forerror.C.mean_all_but([indirect, direct]).mean(
+                direct
+            )
+            s_forerror.run(lambda x: abs(x) ** 2 / 2).mean_all_but(
+                [direct, indirect]
+            ).mean(direct)
+            s_forerror *= df**2  # Δf
+            s_forerror *= N2
+            collected_variance["pathways",j] = s_forerror
+        collected_variance = psd.sqrt(collected_variance.sum("pathways")/len(error_path))
+        return collected_variance
--- a/pyspecProcScripts/calc_error.py
+++ b/pyspecProcScripts/calc_error.py
@@ -2,14 +2,72 @@ import pyspecdata as psd
 import numpy as np
 from .simple_functions import select_pathway
 
-def calc_error(s,int_bounds=(-500,500), signal_pathway = {"ph1":1}, direct= "t2",indirect="nScans",use_mask = False, error_path = {}):
-    f = s.getaxis(direct)
-    df = f[1]-f[0]
-    collected_variance = psd.ndshape(
-        [psd.ndshape(s)[indirect], len(error_path)], [indirect, "pathways"]
-    ).alloc().setaxis("pathways",error_path)
+
+def calc_error(
+    s,
+    frq_slice,
+    signal_pathway,
+    error_path,
+    direct="t2",
+    indirect="nScans",
+    use_mask=False,
+    fl = None,
+):
+    """Calculates the propagation of error for the given signal.
+
+    Before declaring the error_path,
+    look at an examples such as integration_w_error.py to see how to
+    decide which excluded pathways to take the error over.
+
+    Parameters
+    ==========
+    s: nddata
+        Full nddata set in the frequency domain for which the error is being
+        propagated.
+    frq_slice: tuple
+        Frequency range over which the signal resides.
+    signal_pathway:   dict
+                Dictionary of the path of the desired signal.
+    error_path: dict
+                Dictionary of all coherence pathways that are
+                not the signal pathway.
+
+                Before declaring the error_path,
+                look at an examples such as integration_w_error.py to
+                see how to decide which excluded pathways to take the
+                error over.
+    direct:     str
+                Direct axis.
+    indirect:   str
+                Indirect axis.
+    use_mask: boolean
+        If you would rather calculate the error excluding all the phase cycling
+        noise this should be set to true. Specifically this will apply a mask
+        over the signal pathway as well as over the integration slice in all
+        other pathways. When False, the error will only be calculated for the
+        pathways in error_path.
+
+    Returns
+    =======
+    collected_variance:       nddata
+             The error associated with coherence pathways not included in the
+             signal pathway.
+    """
+
+    s = s[direct:frq_slice]
+    df = s.get_ft_prop(direct,"df")
+    collected_variance = (
+        psd.ndshape(
+            [psd.ndshape(s)[indirect], len(error_path)], [indirect, "pathways"]
+        )
+        .alloc()
+        .setaxis("pathways", error_path)
+    )
     if use_mask == False:
-        assert error_path is not None, "If you don't want to use the mask you need to tell me what pathways to use for calculating the noise!"
+        assert error_path is not None, (
+            "If you don't want to use the mask you need to tell me what"
+            " pathways to use for calculating the noise!"
+        )
         for j in range(len(error_path)):
             # calculate N₂ Δf² σ², which is the variance of the integral
             # (by error propagation) where N₂ is the number of points in
@@ -30,6 +88,8 @@ def calc_error(s,int_bounds=(-500,500), signal_pathway = {"ph1":1}, direct= "t2"
             ).mean(direct)
             s_forerror *= df**2  # Δf
             s_forerror *= N2
-            collected_variance["pathways",j] = s_forerror
-        collected_variance = psd.sqrt(collected_variance.sum("pathways")/len(error_path))
+            collected_variance["pathways", j] = s_forerror
+        collected_variance = psd.sqrt(
+            collected_variance.sum("pathways") / len(error_path)
+        )
         return collected_variance
--- a/pyspecProcScripts/integral_w_error.py
+++ b/pyspecProcScripts/integral_w_error.py
@@ -14,7 +14,7 @@ def integral_w_errors(
     convolve_method="Gaussian",
     indirect="vd",
     direct="t2",
-    use_mask = False,
+    use_mask=False,
     fl=None,
     return_frq_slice=False,
 ):
@@ -72,19 +72,18 @@ def integral_w_errors(
             "You have extra (non-phase cycling, non-indirect) dimensions: "
             + str(extra_dims)
         )
-        
-    collected_variance = (
-        psp.ndshape(
-            [psp.ndshape(s)[indirect], len(error_path)], [indirect, "pathways"]
-        )
-        .alloc()
-        .setaxis("pathways", error_path)
-    )
     variance = calc_error(
-        s, frq_slice, sig_path, indirect=indirect, error_path=error_path, use_mask = use_mask,
-        fl = fl
+        s,
+        frq_slice,
+        sig_path,
+        indirect=indirect,
+        error_path=error_path,
+        use_mask=use_mask,
+        fl=fl,
     )
-    retval = select_pathway(s, sig_path).integrate(direct).set_error(variance.data)
+    retval = (
+        select_pathway(s, sig_path).integrate(direct).set_error(variance.data)
+    )
     if not return_frq_slice:
         return retval
     elif return_frq_slice:
@@ -123,6 +122,7 @@ def active_propagation(
     s = s[direct : ((frq_slice[-1] + offset), None)]  # grab all data more than
     #                                             offset to the right of the
     #                                             peak
+    df = s.get_ft_prop(direct, "df")
     all_labels = set(s.dimlabels)
     all_labels -= set([indirect, direct])
     extra_dims = [j for j in all_labels if not j.startswith("ph")]
--- a/pyspecProcScripts/integral_w_error.py
+++ b/pyspecProcScripts/integral_w_error.py
@@ -1,6 +1,7 @@
 import pyspecdata as psp
 from .integrate_limits import integrate_limits
 from .simple_functions import select_pathway
+from .calc_error import calc_error
 import logging
 import numpy as np
 
@@ -62,6 +63,7 @@ def integral_w_errors(
     )
     logging.debug(psp.strm("frq_slice is", frq_slice))
     s = s[direct:frq_slice]
+    variance = calc_error(s,frq_slice, sig_path, indirect = indirect, error_path = error_path)
     f = s.getaxis(direct)
     df = f[1] - f[0]
     all_labels = set(s.dimlabels)
@@ -111,7 +113,7 @@ def integral_w_errors(
         )
     # }}}
     s = select_pathway(s, sig_path)
-    retval = s.integrate(direct).set_error(psp.sqrt(s_forerror.data))
+    retval = s.integrate(direct).set_error(variance.data)
     if not return_frq_slice:
         return retval
     elif return_frq_slice:
--- a/pyspecProcScripts/integral_w_error.py
+++ b/pyspecProcScripts/integral_w_error.py
@@ -14,6 +14,7 @@ def integral_w_errors(
     convolve_method="Gaussian",
     indirect="vd",
     direct="t2",
+    use_mask = False,
     fl=None,
     return_frq_slice=False,
 ):
@@ -47,7 +48,7 @@ def integral_w_errors(
     Returns
     =======
     s:       nddata
-             Data with error associated with coherence pathways
+             Integrated data with error associated with coherence pathways
              not included in the signal pathway.
     """
     assert s.get_ft_prop(direct), "need to be in frequency domain!"
@@ -63,9 +64,6 @@ def integral_w_errors(
     )
     logging.debug(psp.strm("frq_slice is", frq_slice))
     s = s[direct:frq_slice]
-    variance = calc_error(s,frq_slice, sig_path, indirect = indirect, error_path = error_path)
-    f = s.getaxis(direct)
-    df = f[1] - f[0]
     all_labels = set(s.dimlabels)
     all_labels -= set([indirect, direct])
     extra_dims = [j for j in all_labels if not j.startswith("ph")]
@@ -74,46 +72,19 @@ def integral_w_errors(
             "You have extra (non-phase cycling, non-indirect) dimensions: "
             + str(extra_dims)
         )
-    collected_variance = psp.ndshape(
-        [psp.ndshape(s)[indirect], len(error_path)], [indirect, "pathways"]
-    ).alloc()
-    avg_error = []
-    for j in range(len(error_path)):
-        # calculate N₂ Δf² σ², which is the variance of the integral
-        # (by error propagation) where N₂ is the number of points in
-        # the indirect dimension
-        s_forerror = select_pathway(s, error_path[j])
-        # previous line wipes everything out and starts over -- why not use
-        # collected_variance above, as I had originally set up --> part of
-        # issue #44
-        if j == 0:
-            N2 = psp.ndshape(s_forerror)[direct]
-        # mean divides by N₁ (indirect), integrate multiplies by Δf, and the
-        # mean sums all elements (there are N₁N₂ elements)
-        s_forerror -= s_forerror.C.mean_all_but([indirect, direct]).mean(
-            direct
+        
+    collected_variance = (
+        psp.ndshape(
+            [psp.ndshape(s)[indirect], len(error_path)], [indirect, "pathways"]
         )
-        s_forerror.run(lambda x: abs(x) ** 2 / 2).mean_all_but(
-            [direct, indirect]
-        ).mean(direct)
-        s_forerror *= df**2  # Δf
-        s_forerror *= N2
-        avg_error.append(s_forerror)
-    avg_error = sum(avg_error) / len(avg_error)
-    # {{{ variance calculation for debug
-    variance_debug = False
-    if variance_debug:
-        print(
-            "(inside automatic routine) the stdev seems to be",
-            np.sqrt(collected_variance / (df * N2)),
-        )
-        print(
-            "automatically calculated integral error:",
-            np.sqrt(collected_variance.data),
-        )
-    # }}}
-    s = select_pathway(s, sig_path)
-    retval = s.integrate(direct).set_error(variance.data)
+        .alloc()
+        .setaxis("pathways", error_path)
+    )
+    variance = calc_error(
+        s, frq_slice, sig_path, indirect=indirect, error_path=error_path, use_mask = use_mask,
+        fl = fl
+    )
+    retval = select_pathway(s, sig_path).integrate(direct).set_error(variance.data)
     if not return_frq_slice:
         return retval
     elif return_frq_slice:
@@ -152,7 +123,6 @@ def active_propagation(
     s = s[direct : ((frq_slice[-1] + offset), None)]  # grab all data more than
     #                                             offset to the right of the
     #                                             peak
-    df = s.get_ft_prop(direct, "df")
     all_labels = set(s.dimlabels)
     all_labels -= set([indirect, direct])
     extra_dims = [j for j in all_labels if not j.startswith("ph")]


pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+    def masked_var(x):
+        if var_has_imag:  # take average of variance along real and imag
+            return np.var(x[np.isfinite(x)], ddov=1) / 2
+        else:
+            return np.var(x[np.isfinite(x)], ddof=1)
+
+    return np.apply_along_axis(masked_var, axis, x)
+
+
+# }}}
```
guinneak198, line 31: this is used in the time_domain_noise.py - yes I know we aren't calculating t domain noise this is just to give an option for a nan mask vs declaring error pathways when calculating the variance!

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+def calc_error(
+    s,
+    frq_slice,
+    signal_pathway,
+    error_path,
+    direct="t2",
+    indirect="nScans",
+    use_mask=False,
```
guinneak198, line 39: the default is to use the error_path as we've been using it but if True this is going to use the nan mask

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+    s = s[direct:frq_slice]
+    df = s.get_ft_prop(direct, "df")
+    N = len(s[direct])
+    collected_variance = (
+        psd.ndshape(
+            [psd.ndshape(s)[indirect], len(error_path)], [indirect, "pathways"]
+        )
+        .alloc()
+        .setaxis("pathways", error_path)
+    )
```
guinneak198, line 92: allocates an array in the shape of the indirect with the error pathways as an extra dim (used below)

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+                direct
+            )
+            s_forerror.run(lambda x: abs(x) ** 2 / 2).mean_all_but(
+                [direct, indirect]
+            ).mean(direct)
+            s_forerror *= df**2  # Δf
+            s_forerror *= Nshape
+            collected_variance["pathways", j] = s_forerror
+        collected_variance = psd.sqrt(
+            collected_variance.sum("pathways") / len(error_path)
```
guinneak198, line 117: pulled from and cleaned up from original integral_w_errors function but now we are storing the error for each error pathway in their designated spot in the allocated array

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+        collected_variance[direct:frq_slice] = np.nan
+        temp = select_pathway(collected_variance, signal_pathway)
+        temp.data[:] = np.nan
+        if fl is not None:
+            fl.next("Frequency Noise")
+            fl.image(collected_variance)
+        collected_variance.run(masked_var_multi, direct)
+        for j in [k for k in s.dimlabels if k.startswith("ph")]:
+            collected_variance.run(masked_mean_multi, j)
+        collected_variance = psd.sqrt(collected_variance * df**2 * N)
```
guinneak198, line 130: uses nan mask

pyspecProcScripts/integral_w_error.py
```diff
@@ -72,46 +72,18 @@ def integral_w_errors(
...
-            direct
-        )
-        s_forerror.run(lambda x: abs(x) ** 2 / 2).mean_all_but(
-            [direct, indirect]
-        ).mean(direct)
-        s_forerror *= df**2  # Δf
-        s_forerror *= N2
-        avg_error.append(s_forerror)
-    avg_error = sum(avg_error) / len(avg_error)
-    # {{{ variance calculation for debug
```
guinneak198, line 101: all of this now in the calc_error fn

pyspecProcScripts/integral_w_error.py
```diff
@@ -72,46 +72,18 @@ def integral_w_errors(
...
-    if variance_debug:
-        print(
-            "(inside automatic routine) the stdev seems to be",
-            np.sqrt(collected_variance / (df * N2)),
-        )
-        print(
-            "automatically calculated integral error:",
-            np.sqrt(collected_variance.data),
-        )
-    # }}}
```
guinneak198, line 112: I deleted this debug thing since it seemed hard coded False anyway

guinneak198:
tested and ran with examples/integration_w_error.py and check_integration_error.py

pyspecProcScripts/integral_w_error.py
```diff
@@ -72,46 +72,18 @@ def integral_w_errors(
...
+        s,
+        frq_slice,
+        sig_path,
+        indirect=indirect,
+        error_path=error_path,
+        use_mask=use_mask,
+        fl=fl,
+    )
+    retval = (
+        select_pathway(s, sig_path).integrate(direct).set_error(variance.data)
```
jmfranck, line 85: the error should not be  set to the variance -- it should be set to the square root -- not sure  if this is a naming issue or what.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+    assert axis is not None
+
+    def masked_mean(x):
+        return np.mean(x[np.isfinite(x)])
+
+    return np.apply_along_axis(masked_mean, axis, x)
+
+
+def masked_var_multi(x, var_has_imag=False, axis=None):
+    "Calculates the variance along a 1D axis"
```
jmfranck, line 19: it's also important that when you  feed this complex data, it gives you the variance of the real data (or variance of the imaginary data)

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+    assert axis is not None
+
+    def masked_var(x):
+        if var_has_imag:  # take average of variance along real and imag
+            return np.var(x[np.isfinite(x)], ddov=1) / 2
```
jmfranck, line 24: this is where it does what I'm describing above

pyspecProcScripts/calc_error.py
jmfranck, line 130: you do not want to do this here -- this determines the error of the integral.  The point of this function is to give the time-domain variance (or standard deviation -- whichever makes more sense)

jmfranck: (otherwise, it's not going to be useful for your time-domain integration)

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+    Returns
+    =======
+    collected_variance:       nddata
+             The error associated with coherence pathways not included in the
+             signal pathway.
+    """
+
+    s = s[direct:frq_slice]
+    df = s.get_ft_prop(direct, "df")
+    N = len(s[direct])
```
jmfranck, line 85: going with my previous comment, these two should also be moved out of this function.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+    if use_mask is False:
+        assert error_path is not None, (
+            "If you don't want to use the mask you need to tell me what"
+            " pathways to use for calculating the noise!"
+        )
+        for j in range(len(error_path)):
+            # calculate N₂ Δf² σ², which is the variance of the integral
+            # (by error propagation) where N₂ is the number of points in
+            # the indirect dimension
```
jmfranck, line 101: this is not relevant here anymore, I think

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+            # mean divides by N₁ (indirect), integrate multiplies by Δf, and
+            # the mean sums all elements (there are N₁N₂ elements)
+            s_forerror -= s_forerror.C.mean_all_but([indirect, direct]).mean(
+                direct
+            )
+            s_forerror.run(lambda x: abs(x) ** 2 / 2).mean_all_but(
+                [direct, indirect]
+            ).mean(direct)
+            s_forerror *= df**2  # Δf
+            s_forerror *= Nshape
```
jmfranck, line 114: we don't want to do this here -- this is for propagating the error in the frequency domain

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+            collected_variance["pathways", j] = s_forerror
+        collected_variance = psd.sqrt(
+            collected_variance.sum("pathways") / len(error_path)
+        )
```
jmfranck, line 118: let's just get rid of the part where we don't use a mask.  Let's just assume that there is some type of mask.

jmfranck: that way, you can get rid of the need to allocate this `collected_variance` thing, too.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+        pathways in error_path.
+
+    Returns
+    =======
+    collected_variance:       nddata
+             The error associated with coherence pathways not included in the
+             signal pathway.
+    """
+
+    s = s[direct:frq_slice]
```
jmfranck, line 83: you don't want to frequency slice here.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+            s_forerror *= df**2  # Δf
+            s_forerror *= Nshape
+            collected_variance["pathways", j] = s_forerror
+        collected_variance = psd.sqrt(
+            collected_variance.sum("pathways") / len(error_path)
+        )
+    else:
+        collected_variance = s.C  # so we don't alter s
+        collected_variance[direct:frq_slice] = np.nan
+        temp = select_pathway(collected_variance, signal_pathway)
```
jmfranck, line 122: here, you  should rather loop over `signal_pathway` as well as `excluded_pathways`

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+    Parameters
+    ==========
+    s: nddata
+        Full nddata set in the frequency domain for which the error is being
+        propagated.
+    frq_slice: tuple
+        Frequency range over which the signal resides.
+    signal_pathway:   dict
+                Dictionary of the path of the desired signal.
+    error_path: dict
```
jmfranck, line 57: do not define this -- rather, define excluded  pathways

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+    Before declaring the error_path,
+    look at an examples such as integration_w_error.py to see how to
+    decide which excluded pathways to take the error over.
+
+    Parameters
+    ==========
+    s: nddata
+        Full nddata set in the frequency domain for which the error is being
+        propagated.
+    frq_slice: tuple
```
jmfranck, line 53: this gets eliminated, as noted below.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+    direct:     str
+                Direct axis.
+    indirect:   str
+                Indirect axis.
+    use_mask: boolean
+        If you would rather calculate the error excluding all the phase cycling
+        noise this should be set to true. Specifically this will apply a mask
+        over the signal pathway as well as over the integration slice in all
+        other pathways. When False, the error will only be calculated for the
+        pathways in error_path.
```
jmfranck, line 74: this gets removed, since we always want to use a mask

pyspecProcScripts/calc_error.py
jmfranck: to be clear, I think it makes sense to return the variance from this function, then you do the math I'm saying to remove here inside the integration function (which involves taking the sqrt and converting to a stdev)

jmfranck (changes requested):
I made series of individual comments above.  The general theme is that we want ar function that calculates the masked variance that we can re-use not only in the newer code (rough table of integrals) but also when you do time-domain integration, etc.)

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
+import pyspecdata as psd
+from .simple_functions import select_pathway
+import numpy as np
+
+
+# {{{ Functions to calculate var and mean of nan-masked data
+#     Pulled from time_domain_noise.py example
+def masked_mean_multi(x, axis=None):
```
jmfranck, line 8: both this and the following function should be renamed starting with a single underscore, indicating they are intended to be private to this module.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,131 @@
...
+        if var_has_imag:  # take average of variance along real and imag
+            return np.var(x[np.isfinite(x)], ddov=1) / 2
+        else:
+            return np.var(x[np.isfinite(x)], ddof=1)
+
+    return np.apply_along_axis(masked_var, axis, x)
+
+
+# }}}
+def calc_error(
```
jmfranck, line 32: rename this as `calc_masked_error`

pyspecProcScripts/integral_w_error.py
```diff
@@ -13,6 +14,7 @@ def integral_w_errors(
     convolve_method="Gaussian",
     indirect="vd",
     direct="t2",
+    use_mask=False,
```
jmfranck, line 17: this should not have `use_mask` defined.  Also, it should not have `error_path`, but rather `excluded_pathways`

pyspecProcScripts/calc_error.py
jmfranck: though I don't usually like short functions, this does add to the clarity.

pyspecProcScripts/calc_error.py
jmfranck: see comments about using `excluded_pathways` instead

pyspecProcScripts/calc_error.py
jmfranck: this should be obsolete now.

pyspecProcScripts/calc_error.py
jmfranck: ???

pyspecProcScripts/integral_w_error.py
jmfranck: OK, so having read more, I can clarify that your `calc_masked_variance` (as I'm suggesting you rename the function) will return *σ_ν²* (which is the **variance of your spectral datapoint** or, if you've taken a unitary FT, the **variance of your time domain datapoints**).  You then multiply by the other stuff (*Δν²* etc) and take the square root to determine the **estimated standard deviation of your integral,** which is what you attach as the error here.

pyspecProcScripts/calc_error.py
jmfranck: (I'm saying add this info to the docstring)

pyspecProcScripts/calc_error.py
jmfranck: (this is the "multiplying by other stuff" that I refer to above)

pyspecProcScripts/calc_error.py
jmfranck: but, I think this is an obsolete comment, since, as noted elsewhere, you just want to simplify your code by removing support for data that is not masked.

pyspecProcScripts/calc_error.py
jmfranck: (again, I think this is obsolete)

pyspecProcScripts/calc_error.py
jmfranck: because you can use data from outside the bounds of the frequency slice to calculate the error.

pyspecProcScripts/calc_error.py
guinneak198: the integral_w_error function that we are basing this off of was propagating error. Please specify if you want this function to only calculate the variance of the frequency domain vs propagated error

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,97 @@
...
+    Returns
+    =======
+    collected_variance:       nddata
+             The error associated with coherence pathways not included in the
+             signal pathway.
+    """
+    df = s.get_ft_prop(direct, "df")
+    N = len(s[direct])
+    # {{{ Define the pathways used for calculating the error
+    collected_variance = s.C  # so we don't alter s
```
guinneak198, line 77: so now I have a copy of the data that below I will be applying masks to based on excluded pathways

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,97 @@
...
+    temp.data[:] = np.nan
+    if excluded_pathways is not None:
+        error_paths = [{phcycdims[i]: path[i] for i in range(len(phcycdims))} for
+                path in excluded_pathways]
+        for j in range(len(error_paths)):
+            temp = select_pathway(collected_variance,error_paths[j])
+            temp.data[:] = np.nan
+    else:
+        temp = collected_variance[direct:frq_slice]
+        temp.data[:] = np.nan
```
guinneak198, line 89: have an option where a mask will be made over the signal pathway and the f slice of all pathways (assume phase cycling noise is in f slice in all other pathways) --> example still passes when excluded pathways = None

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,97 @@
...
+    if fl is not None:
+        fl.next("Frequency Noise")
+        fl.image(collected_variance)
+    collected_variance.run(_masked_var_multi, direct)
+    for j in [k for k in s.dimlabels if k.startswith("ph")]:
+        collected_variance.run(_masked_mean_multi, j)
+    collected_variance = collected_variance * df**2 * N
```
guinneak198, line 96: still confused because I thought we wanted this function to produce the propagated error that we then attach to the integrals but waiting on response/discussion for this

pyspecProcScripts/integral_w_error.py
```diff
@@ -1,14 +1,15 @@
...
+from .calc_error import calc_masked_error
 import logging
 import numpy as np
 
 
 def integral_w_errors(
     s,
     sig_path,
-    error_path,
+    excluded_pathways=None,
```
guinneak198, line 12: again the default masks out the signal pathway and the f slice in all other pathways (assumes ph cycling noise in f slice in all other pathways)

pyspecProcScripts/calc_error.py
guinneak198: the collected variance was a way to store the variance of each coherence pathway (nix the excluded_pathways and signal pathwa) and then we take the sum to return the total collected variance... still get rid of?

pyspecProcScripts/integral_w_error.py
guinneak198: okay so now the calc_masked_variance only returns sigma_f^2

pyspecProcScripts/calc_error.py
guinneak198: removed

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,97 @@
...
+        propagated.
+    frq_slice: tuple
+        Frequency range over which the signal resides.
+    signal_pathway:   dict
+                Dictionary of the path of the desired signal.
+    excluded_pathways: list
+                List of tuples containing the coherence pathways that are
+                to be masked out when calculating the error.
+                If no excluded_pathways are fed, the function will apply a mask
+                over just the signal pathway.
```
guinneak198, line 62: I like having a default

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,97 @@
...
+    Returns
+    =======
+    collected_variance:       nddata
+             The error associated with coherence pathways not included in the
+             signal pathway.
+    """
+    # {{{ Determine if the data has an imaginary component
+    var_has_imag = False
+    if np.iscomplex(s):
+        var_has_imag = True
```
guinneak198, line 76: easy way to determine whether we feed var_has_imag as true or false to the _masked_var_multi fn

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,97 @@
...
+    # }}}
+    # {{{ Define the pathways used for calculating the error
+    collected_variance = s.C  # so we don't alter s
+    phcycdims = [j for j in s.dimlabels if j.startswith("ph")]
+    temp = select_pathway(collected_variance, signal_pathway)
+    temp.data[:] = np.nan
```
guinneak198, line 82: automatically filter out signal pathway

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,97 @@
...
+    if excluded_pathways is not None:
+        error_paths = [
+            {phcycdims[i]: path[i] for i in range(len(phcycdims))}
+            for path in excluded_pathways
+        ]
+        for j in range(len(error_paths)):
+            temp = select_pathway(collected_variance, error_paths[j])
+            temp.data[:] = np.nan
```
guinneak198, line 90: loop over excluded pathways to filter out

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,97 @@
...
+    if fl is not None:
+        fl.next("Frequency Noise")
+        fl.image(collected_variance)
+    collected_variance.run(_masked_var_multi, direct, var_has_imag)
+    for j in [k for k in s.dimlabels if k.startswith("ph")]:
+        collected_variance.run(_masked_mean_multi, j, var_has_imag)
```
guinneak198, line 96: loop over each ct pathway and calc var

examples/check_integration_error.py
```diff
@@ -62,20 +62,10 @@
...
-        ))
-        - set(excluded_pathways)
-        - set([(signal_pathway["ph1"], signal_pathway["ph2"])])
-    )
-    error_pathway = [{"ph1": j, "ph2": k} for j, k in error_pathway]
     s_int, frq_slice = integral_w_errors(
         data,
         signal_pathway,
-        error_pathway,
+        excluded_pathways,
```
jmfranck, line 68: yes!

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,98 @@
...
+    return np.apply_along_axis(masked_mean, axis, x)
+
+
+def _masked_var_multi(x, var_has_imag=False, axis=None):
+    """Calculates the variance along a 1D axis.
+    If the data is complex you must assign var_has_imag as true.
+    By default it calculates the variance of the real of the data"""
+    assert axis is not None
+
+    def masked_var(x):
```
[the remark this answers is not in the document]
jmfranck: no, because you couldn't import it into another module, anyways.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,102 @@
...
+        return np.mean(x[np.isfinite(x)])
+
+    return np.apply_along_axis(masked_mean, axis, x)
+
+
+def _masked_var_multi(x, var_has_imag=False, axis=None):
+    """Calculates the variance along a 1D axis.
+    If the data is complex you must assign var_has_imag as true
+    so that the calculated variance is divided by 2.
+    By default it calculates the variance of the real of the data"""
```
jmfranck, line 21: why can't it just figure  out if it has an imaginary part?

pyspecProcScripts/calc_error.py
jmfranck: see above -- just do it in place --  this shouldn't be here.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,102 @@
...
+    collected_variance:       nddata
+             The error associated with coherence pathways not included in the
+             signal pathway.
+    """
+    # {{{ Determine if the data has an imaginary component
+    var_has_imag = False
+    if np.iscomplex(s):
+        var_has_imag = True
+    # }}}
+    phcycdims = [j for j in s.dimlabels if j.startswith("ph")]
```
jmfranck, line 80: this shouldn't be needed

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,102 @@
...
+    collected_variance = s.C  # so we don't alter s
+    # {{{ filter out signal pathway and excluded error pathways
+    temp = select_pathway(collected_variance, signal_pathway)
+    temp.data[:] = np.nan
+    if excluded_pathways is not None:
```
jmfranck, line 85: `if excluded_pathways is not None and len(excluded_pathways) > 0`

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,102 @@
...
+        error_paths = [
+            {phcycdims[i]: path[i] for i in range(len(phcycdims))}
+            for path in excluded_pathways
+        ]
```
jmfranck, line 89: why? this is what I was saying -- excluded pathways should just be a list of dicts, since that's how we specify coherence pathways now.

pyspecProcScripts/calc_error.py
jmfranck: these should not be called error pathways.  you should just directly do:
```
if type(excluded_pathways) == dict:
    raise ValueError("excluded pathways should be a list of dicts.  If you really mean to exclude only one pathway, pass a list with a single dict inside")
    
for j in excluded_pathways:
    temp = select_pathway(collect... , j)
    .... = nan
````

File: pyspecProcScripts/calc_error.py [hunk 38]
```diff
@@ -0,0 +1,102 @@
...
+        for j in range(len(error_paths)):
+            temp = select_pathway(collected_variance, error_paths[j])
+            temp.data[:] = np.nan
+    # }}}
+    # Filter out frq_slice where ph noise resides        
+    collected_variance[direct:frq_slice].data[:] = np.nan
```
jmfranck, line 95: this should be a loop over the list of frequency slices, like we discussed

jmfranck: list of frequency slices that are excluded

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,102 @@
...
+
+    Parameters
+    ==========
+    s: nddata
+        Full nddata set in the frequency domain for which the error is being
+        propagated.
+    frq_slice: tuple
+        Frequency range that will be filtered out in calculating the error - it
+        is assumed this region in all coherence transfer pathways contains some
+        amount of phase cycling noise.
```
jmfranck, line 57: this is confusing.  You also need to specify the frequency slice for the integration.  I would call that `frq_slice`.  On the other hand, you want a list tuples -- something like `excluded_frequences`, to mimic the `excluded_pathways` below.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,102 @@
...
+    signal_pathway:   dict
+                Dictionary of the path of the desired signal.
```
jmfranck, line 59: I realized you  don't want to pass this separately.  Here, there is not difference between the signal pathway and any other excluded pathway

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,102 @@
...
+        for j in range(len(error_paths)):
+            temp = select_pathway(collected_variance, error_paths[j])
+            temp.data[:] = np.nan
+    # }}}
+    # Filter out frq_slice where ph noise resides        
+    collected_variance[direct:frq_slice].data[:] = np.nan        
+    if fl is not None:
+        fl.next("Frequency Noise")
+        fl.image(collected_variance)
+    collected_variance.run(_masked_var_multi, direct, var_has_imag)
```
jmfranck, line 99: do not pass var_has_imag -- see above

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,102 @@
...
+    for j in [k for k in s.dimlabels if k.startswith("ph")]:
+        collected_variance.run(_masked_mean_multi, j, var_has_imag)
```
jmfranck, line 101: this doesn't make sense -- I'm going to comment above about how to change the function.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,102 @@
...
+
+    return np.apply_along_axis(masked_mean, axis, x)
+
+
+def _masked_var_multi(x, var_has_imag=False, axis=None):
+    """Calculates the variance along a 1D axis.
+    If the data is complex you must assign var_has_imag as true
+    so that the calculated variance is divided by 2.
+    By default it calculates the variance of the real of the data"""
+    assert axis is not None
```
jmfranck, line 22: do not do this.

pyspecProcScripts/calc_error.py
jmfranck: yea, so here, you want to just end with `return _masked_var_multi(collected_variance.data).item()`

pyspecProcScripts/integral_w_error.py
```diff
@@ -62,8 +59,10 @@ def integral_w_errors(
     )
     logging.debug(psp.strm("frq_slice is", frq_slice))
     s = s[direct:frq_slice]
-    f = s.getaxis(direct)
-    df = f[1] - f[0]
+    # {{{ variables in calculating error over slice
+    N = len(s[direct])  # number of pts within the slice
```
jmfranck, line 63: `N = s.shape[direct]`

pyspecProcScripts/integral_w_error.py
```diff
@@ -72,46 +71,25 @@ def integral_w_errors(
...
-        print(
-            "automatically calculated integral error:",
-            np.sqrt(collected_variance.data),
-        )
-    # }}}
-    s = select_pathway(s, sig_path)
-    retval = s.integrate(direct).set_error(psp.sqrt(s_forerror.data))
+    # Determine smaller slice within signal bounds where ph cycling noise most
+    # likely resides
+    frq_filter_push = (frq_slice[1] - frq_slice[0]) / 3
```
jmfranck, line 76: what is this? you were not doing this before.  Also, I would not call this `frq_filter_push`.  Rahter `frq_slice[1]-frq_slice[0]` is the width of your integration bounds

pyspecProcScripts/integral_w_error.py
```diff
@@ -72,46 +71,25 @@ def integral_w_errors(
...
+        frq_slice[-1] - frq_filter_push,
+    )
+    variance = calc_masked_error(
+        s,
+        frq_filter_bounds,
+        sig_path,
+        indirect=indirect,
+        excluded_pathways=excluded_pathways,
+        fl=fl,
+    )
```
jmfranck, line 88: this needs to be moved up, before you take `frq_slice` to calculate the integral.

pyspecProcScripts/integral_w_error.py
jmfranck: but you definitely are not going to want to do something like this.

jmfranck (changes requested):
this really didn't seem to reflect our conversation at all, but I hope that the comments help to guide it in the right direction.

pyspecProcScripts/calc_error.py
guinneak198: I like having this as a default mask choice though so even if now excluded pathways are fed it will still filter out the signal pathway... still nix it?

pyspecProcScripts/calc_error.py
guinneak198: a) masked_var_multi needs an axis kwarg (the only remaining axis is indirect)
b) when I use indirect as the axis the returned value is one float
c) this would mean the error for all datapoints is the same (err = sqrt(variance*df**2*N)))
d) UNLESS you mean move the `collected_variance.run(_masked_var_multi,direct)` down to the return (so after the `_masked_mean_multi`)? but again that would not be an item

guinneak198: [question](https://jmfrancklab.slack.com/archives/CV553TPRV/p1742390027557529)

pyspecProcScripts/calc_error.py
guinneak198: unless you mean nixing the var_has_imag which I did

examples/check_integration_error.py
```diff
@@ -62,20 +66,11 @@
...
-        ))
-        - set(excluded_pathways)
-        - set([(signal_pathway["ph1"], signal_pathway["ph2"])])
-    )
-    error_pathway = [{"ph1": j, "ph2": k} for j, k in error_pathway]
     s_int, frq_slice = integral_w_errors(
         data,
-        signal_pathway,
-        error_pathway,
+        sig_path=signal_pathway,
```
jmfranck, line 71: already discussed how we don't want to define signal pathway, and only want to define the excluded pathways.  assuming it's likely other comments haven't been addressed, and stopping here

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,105 @@
...
+    """Calculates the variance along a 1D axis.
+    If the data is complex you must assign var_has_imag as true
+    so that the calculated variance is divided by 2.
+    By default it calculates the variance of the real of the data"""
+
+    def masked_var(x):
+        if np.iscomplex(axis) and sum(abs(np.imag(axis))) > 1e-7:
+            # take average of variance along real and imag
+            return np.var(x[np.isfinite(x)], ddov=1) / 2
+        else:
```
jmfranck, line 27: here, you actually want ot make sure that it's `not np.iscomplex`  if it's alleging to be complex, but the imaginary part is small, then throw and error

pyspecProcScripts/calc_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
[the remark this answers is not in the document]
guinneak198: checked

pyspecProcScripts/calc_error.py
guinneak198: done in masked_var fn

pyspecProcScripts/calc_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
guinneak198: throws a ValueError now if this passes

pyspecProcScripts/calc_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
guinneak198: check

pyspecProcScripts/calc_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
guinneak198: calc_error no longer takes signal_pathways as a kwarg

pyspecProcScripts/calc_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,102 @@
...
+    By default it calculates the variance of the real of the data"""
+    assert axis is not None
+
+    def masked_var(x):
+        if var_has_imag:  # take average of variance along real and imag
+            return np.var(x[np.isfinite(x)], ddov=1) / 2
+        else:
+            return np.var(x[np.isfinite(x)], ddof=1)
+
+    return np.apply_along_axis(masked_var, axis, x)
```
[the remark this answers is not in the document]
guinneak198: checked

pyspecProcScripts/integral_w_error.py
guinneak198: checked

pyspecProcScripts/integral_w_error.py
guinneak198: no longer used

pyspecProcScripts/integral_w_error.py
guinneak198: moved before frequency slice

pyspecProcScripts/calc_error.py
guinneak198: this throws an error now if that's what you mean? I tried chatting with chatgpt and it seemed to come to the same conclusion (that you wanted an error thrown)

examples/check_integration_error.py
jmfranck: to be consistent with other code in the repo
, this should be called `signal_pathway`

pyspecProcScripts/calc_error.py
jmfranck: no -- the current else block should be `elif not np.iscomplex...` and then there should be an else block at the end that throws an error

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,104 @@
...
+            "excluded_frqs should be a list of tuples."
+            "If you really mean to exclude only one slice, pass a "
+            "list with a single tuple inside"
+        )
+    if excluded_frqs is not None and len(excluded_frqs) > 0:
+        for j in range(len(excluded_frqs)):
+            collected_variance[direct : excluded_frqs[j]].data[:] = np.nan
+    # }}}
+    if fl is not None:
+        fl.next("Frequency Noise")
```
jmfranck, line 98: call this masked frequency noise

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,104 @@
...
+        fl.image(collected_variance)
+    # {{{ Average over remaining ct pathways
+    for j in [k for k in s.dimlabels if k.startswith("ph")]:
+        collected_variance.run(_masked_mean_multi, j)
+    # }}}
```
jmfranck, line 103: you should not need to do this (see previous comments)

pyspecProcScripts/integral_w_error.py
```diff
@@ -25,29 +27,31 @@ def integral_w_errors(
...
+        Direct axis.
 
     Returns
     =======
-    s:       nddata
-             Data with error associated with coherence pathways
-             not included in the signal pathway.
+    s: nddata
+        Integrated data with error associated with coherence pathways not
+        included in the signal pathway.
```
jmfranck, line 54: not quite:

```
Data sampled at the indicated coherence pathway, and integrated [HOW do you determine the bounds? can you specify manual bounds??], with errors determined by error propagation in the frequency domain, with the noise associated with the spectral datapoints determined from the masked variance of the DCCT data.
```

pyspecProcScripts/integral_w_error.py
```diff
@@ -72,46 +85,10 @@ def integral_w_errors(
...
-            np.sqrt(collected_variance / (df * N2)),
-        )
-        print(
-            "automatically calculated integral error:",
-            np.sqrt(collected_variance.data),
-        )
-    # }}}
-    s = select_pathway(s, sig_path)
-    retval = s.integrate(direct).set_error(psp.sqrt(s_forerror.data))
+    std_error = psp.sqrt(variance * df**2 * N)
```
jmfranck, line 88: OK, I see now -- this is where you want to have a variance with the indirect dimension.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,104 @@
...
+    if excluded_frqs is not None and len(excluded_frqs) > 0:
+        for j in range(len(excluded_frqs)):
+            collected_variance[direct : excluded_frqs[j]].data[:] = np.nan
+    # }}}
+    if fl is not None:
+        fl.next("Frequency Noise")
+        fl.image(collected_variance)
+    # {{{ Average over remaining ct pathways
+    for j in [k for k in s.dimlabels if k.startswith("ph")]:
+        collected_variance.run(_masked_mean_multi, j)
```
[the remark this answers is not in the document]
jmfranck: but this is not quite the goal here -- the goal should be to take the variance over all dimension that are *not* indirect.  You  do that like this:
```
collected_variance.run(_masked_var_multi, direct)
for j in set(s.dimlabels) - set([direct, indirect]):
    collected_variance.run(_masked_mean_multi, j)
```

pyspecProcScripts/calc_error.py
guinneak198: I don't quite understand why this isn't needed (runs fine without it) but I _think_ it's because we are using this collected_variance thing?

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,104 @@
...
+        return masked_var(x)
+
+
+# }}}
+def calc_masked_error(
+    s,
+    excluded_frqs=None,
+    excluded_pathways=None,
+    direct="t2",
+    indirect="nScans",
```
jmfranck, line 41: it's weird that this defaults to `"nScans"`, but the integration function defaults to `"vd"`.  Pick one.

pyspecProcScripts/calc_error.py
jmfranck: this and the earlier comment I refer to are incorrect -- see my new comments about returning an nddata with dimension `indirect`

jmfranck (changes requested):
I reviewed code, added new comments, and then went through and resolved previous comments that are no longer relevant.

pyspecProcScripts/calc_error.py
jmfranck: this and the earlier comment I refer to are incorrect -- see my new comments about returning an nddata with dimension `indirect`

pyspecProcScripts/integral_w_error.py
```diff
@@ -72,46 +85,10 @@ def integral_w_errors(
...
+    retval = (
+        select_pathway(s, sig_path).integrate(direct).set_error(std_error.data)
```
jmfranck, line 90: remove line 88 and change this to
```
select_pathway(s, sig_path).integrate(direct).set_error(np.sqrt(variance.data * N) * df)
```
(which does the same thing in one line)

jmfranck: rename `variance` to `spectral_datapoint_variance`.

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,104 @@
...
+        If no excluded_pathways are fed, the function will apply a mask
+        over just the signal pathway.
+    direct: str
+        Direct axis.
+    indirect: str
+        Indirect axis.
+    Returns
+    =======
+    collected_variance: float
+        The variance of the spectral datapoints.
```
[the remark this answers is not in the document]
guinneak198: [indirect!](https://github.com/jmfrancklab/proc_scripts/pull/143#discussion_r2003810738)

examples/check_integration_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def masked_mean(x):
+        return np.mean(x[np.isfinite(x)])
+
+    return np.apply_along_axis(masked_mean, axis, x)
+
+
+def _masked_var_multi(x, axis=None):
+    """Calculates the variance along a 1D axis.
+    Note, it calculates the variance of the real of the data"""
```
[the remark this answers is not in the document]
guinneak198: checked

pyspecProcScripts/calc_error.py
```diff
@@ -0,0 +1,104 @@
...
+    if isinstance(excluded_pathways, dict):
+        raise ValueError(
+            "excluded_pathways should be a list of dicts."
+            "If you really mean to exclude only one pathway, pass a "
+            "list with a single dict inside"
+        )
+    if excluded_pathways is not None and len(excluded_pathways) > 0:
+        for j in range(len(excluded_pathways)):
+            temp = select_pathway(collected_variance, excluded_pathways[j])
+            temp.data[:] = np.nan
```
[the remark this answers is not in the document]
guinneak198: checked

pyspecProcScripts/calc_error.py
guinneak198: checked

pyspecProcScripts/integral_w_error.py
```diff
@@ -1,14 +1,16 @@
...
 from .integrate_limits import integrate_limits
 from .simple_functions import select_pathway
+from .calc_error import calc_masked_error
 import logging
 import numpy as np
 
 
 def integral_w_errors(
     s,
     sig_path,
```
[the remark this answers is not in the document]
guinneak198: checked

pyspecProcScripts/integral_w_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
guinneak198: checked -s tick with vd

pyspecProcScripts/integral_w_error.py
guinneak198: checked

pyspecProcScripts/calc_error.py
jmfranck: not resolved yet

jmfranck (changes requested):
see todos in code and the comments in the conversation that are not resolved (might just be one left?)

examples/integration_with_error.py
```diff
@@ -46,33 +51,25 @@
...
 # }}}
 data.ft("t2", shift=True)
-dt = data.get_ft_prop("t2", "dt")
-# {{{ vector-normalize the FT
-data /= sqrt(ndshape(data)["t2"]) * dt
+# {{{ usually, we don't use a unitary FT -- this makes it unitary
+#     i.e. vector-normalize the FTs above
+for thisdim in ["t2", "ph1", "ph2"]:
+    dt = data.get_ft_prop("t2", "dt")
+    data /= sqrt(ndshape(data)["t2"]) * dt
```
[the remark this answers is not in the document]
jmfranck: This was a typo. The idea was to replace t2 with thisdim. One minute though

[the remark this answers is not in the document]
jmfranck: But, I think that we should rather focus on using normal Fourier transforms, and then just converting from σ_ν to σ_t if our code requires that. (Which it does not unless it's a Time domain error propagation)

The last thing that I did was to edit the file where we demonstrate how to do this.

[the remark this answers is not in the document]
jmfranck: But, I think that we should rather focus on using normal Fourier transforms, and then just converting from σ_ν to σ_t if our code requires that. (Which it does not unless it's a Time domain error propagation)

The last thing that I did was to edit the file where we demonstrate how to do this.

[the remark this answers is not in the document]
jmfranck: (how to perform the σ_ν to σ_t conversion)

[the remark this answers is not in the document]
guinneak198: I'm confused are you saying then to replace the hard coded dim "t2" with thisdim? what do you mean by "rather focus on using the normal FTs and converting? does that pertain to these lines?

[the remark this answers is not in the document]
guinneak198: okay well replacing the hard coded t2 worked still not sure what you mean by "rather focus..."

pyspecProcScripts/calc_error.py
guinneak198: this is outdated and since been fixed

examples/integration_with_error.py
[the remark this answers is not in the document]
jmfranck: I'm going to resolve this, and make a todo in the code

jmfranck (changes requested):
see TODOs in code

examples/check_integration_error.py
```diff
@@ -53,29 +57,17 @@
...
-    data /= 0.5 * 0.25  # the dt in the integral for both dims
-    data /= sqrt(ndshape(data)["ph1"] * ndshape(data)["ph2"])  # normalization
-    # }}}
+    data.ft(["ph1", "ph2"],unitary=True)
+    # {{{ vector normalize
     dt = diff(data.getaxis("t2")[r_[0, 1]]).item()
     data.ft("t2", shift=True)
-    # {{{
+    # }}}
     data /= sqrt(ndshape(data)["t2"]) * dt
```
guinneak198, line 65: this is the line I include in integration_with_error.py as well but unsure if this is problematic/contradicts what we are saying

examples/integration_with_error.py
```diff
@@ -39,115 +43,113 @@
...
-fl.plot(just_noise, ".", capsize=6)
-# {{{ usually, we don't use a unitary FT -- this makes it unitary
-data /= 0.5 * 0.25  # the dt in the integral for both dims
-data /= sqrt(ndshape(data)["ph1"] * ndshape(data)["ph2"])  # normalization
-# }}}
 data.ft("t2", shift=True)
+data.ft(["ph1", "ph2"], unitary=True)
 dt = data.get_ft_prop("t2", "dt")
 # {{{ vector-normalize the FT
 data /= sqrt(ndshape(data)["t2"]) * dt
```
guinneak198, line 50: without this line code fails

jmfranck: I'm going to assume this comment is made obsolete by my recent push

examples/check_integration_error.py
jmfranck: I assume this is also obsolete with my recent push?

jmfranck: I assume this is also obsolete with my recent push?

examples/check_integration_error.py
```diff
@@ -41,90 +45,87 @@
 clean_data["t2":0] *= 0.5
 fake_data_noise_std = 2.0
 clean_data.reorder(["ph1", "ph2", "vd"])
-bounds = (0, 200)  # seem reasonable to me
```
guinneak198, line 44: I am going to use the bounds returned by the function so that the integrals are the same but we are truly testing the error as the name implies

examples/check_integration_error.py
```diff
@@ -41,90 +45,87 @@
...
+        data /= sqrt(ndshape(data)["t2"]) * dt
+        s_int, frq_slice = frequency_domain_integral(
+            data,
+            signal_pathway=signal_pathway,
+            excluded_pathways=excluded_pathways,
+            indirect="vd",
+            fl=fl,
+            return_frq_slice=True,
+        )
+        manual_integration = select_pathway(data, signal_pathway)[
```
guinneak198, line 72: here I am integrating over the same range as the function but manually calculating the error for comparison of the error as the name implies

jmfranck: But the point of the function, and what I think the legend indicates, is to compare the automatically determined bounds versus bounds that you have yourself manually chosen

jmfranck: But ine of the things that we are testing in the example, and what I think the legend indicates, is to compare the automatically determined bounds versus bounds that you have yourself manually chosen

jmfranck: But one of the things that we are testing in the example, and what I think the legend indicates, is to compare the automatically determined bounds versus bounds that you have yourself manually chosen

examples/check_integration_error.py
jmfranck: Not a good change

jmfranck (changes requested):
you just need to figure out why it's not passing the checks.

jmfranck (approved):

PR #146, opened 2025-05-08
HH for weighted integral function

guinneak198:
need a simple function to generate a HH with a width equal to integration bounds that is IFTed to make the weighted integral function when integrating and propagating error in the t domain

### DIFF
--- a/examples/sinc_weighted_integral_fn.py
+++ b/examples/sinc_weighted_integral_fn.py
@@ -1,0 +1,52 @@
+"""
+Weighted Integral Function Example
+==================================
+
+The weighted integral function is utilized in integrating a signal.
+Specifically here, we generate a sinc function whose FT is a heaviside hat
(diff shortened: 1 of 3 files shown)


examples/sinc_weighted_integral_fn.py
```diff
@@ -0,0 +1,53 @@
...
+    fl.next("Weighted Integral Functions", fig=fig)
+    fl.skip_units_check()
+    thisax[0].set_title("Frequency Domain")
+    thisax[1].set_title("Time Domain")
+    fig.suptitle("Weighted Integral Functions")
+    # }}}
+    data.set_units("t2", "s")
+    data.ft("t2", shift=True)
+    # {{{ make weighted integral function
+    wt_fn = psdpr.HH_weighted_integral(data, f_range)
```
jmfranck, line 43: can we just call this HH_time_domain?

PR #152, opened 2026-04-16
adding synthetic test proc type

atahangarip:
adding synthetic test proc type for save_data_example.py which was not present in load_data.py

### DIFF
--- a/pyspecProcScripts/load_data.py
+++ b/pyspecProcScripts/load_data.py
@@ -749,6 +749,14 @@ def proc_capture(s):
     return s
 
 
+def proc_save_data_example(s, fl=None):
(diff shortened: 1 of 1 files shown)


jmfranck:
I disagree with this.  I don't think that the save_data_example in FLinst has a postproc_type set.  Which means that it should not try to post-process at all.  OR, you could set postproc_type to something sensible.  OR, if that's the issue you might need to use [this](https://jmfrancklab.github.io/pyspecdata/auto_api/pyspecdata.load_files.html#pyspecdata.load_files.register_proc_lookup) to set the postproc dict to zero -- though post on slack if that's the case, b/c it really shouldn't be needed.

PR #158, opened 2026-07-01
Optimizing Correlation Alignment

atahangarip:
We get 2x improvement in the computation time with these changes.


pyspecProcScripts/correlation_alignment.py
```diff
@@ -285,15 +285,33 @@ def smoosh_or_rename(temp_phcycdims, x):
...
-                ].run(
-                    lambda x, axis=None: np.roll(x, ph_index, axis=axis),
-                    phname,
+            delta_name = "Delta%s" % phname.capitalize()
+            delta_axis = s_leftbracket.dimlabels.index(delta_name)
+            if s_leftbracket.shape[delta_name] != phlen:
+                raise ValueError(
+                    f"{phname} and {delta_name} must have the same length"
                 )
+            phase_index = np.arange(phlen)[:, None]
```
jmfranck: can we call this ph_idx_array (and the other delta_idx_array)?

It's an array of all the indeces along the given dimension -- it's not jsut a single index

pyspecProcScripts/correlation_alignment.py
```diff
@@ -285,15 +285,33 @@ def smoosh_or_rename(temp_phcycdims, x):
...
-            for ph_index in range(phlen):
-                s_leftbracket[
-                    "Delta%s" % phname.capitalize(), ph_index
-                ] = s_leftbracket[
-                    "Delta%s" % phname.capitalize(), ph_index
-                ].run(
-                    lambda x, axis=None: np.roll(x, ph_index, axis=axis),
-                    phname,
+            delta_name = "Delta%s" % phname.capitalize()
+            delta_axis = s_leftbracket.dimlabels.index(delta_name)
```
jmfranck: so, I would call this delta_axn.  Also, use the .axn( rather than .dimlabels.index(

(axn stands for axis number)

pyspecProcScripts/correlation_alignment.py
```diff
@@ -285,15 +285,33 @@ def smoosh_or_rename(temp_phcycdims, x):
...
+            # copy with (phase-delta) modulo the phase-cycle length constructs
+            # every rolled column at once and gives the same Delta-phi table.
+            # Keeping the NumPy operation inside nddata.run preserves the
+            # nddata axes, properties, and errors. This removes repeated
+            # nddata slicing/assignment overhead; it improves computation,
+            # not the alignment physics or phase convention.
+            s_leftbracket.run(
+                lambda data, axis: np.moveaxis(
+                    np.moveaxis(data, [axis, delta_axis], [0, 1])[:, 0, ...][
+                        (phase_index - delta_index) % phlen, ...
```
jmfranck: so if I understand correctly, you are saying that this difference modulo thing has the effect of rolling all the dimensions at once?  I am a bit confused about this indexing method here -- is the thing that comes before the comma a 2D array, then?  What, in this instance does `...` mean?

(can you use this feedback to update your comment -- and leave it as an explanatory code comment, not addressed to me).

atahangarip: Yes, the modulo expression creates a 2D array of source indices with shape (phase, delta-phase), which constructs all rolled columns at once. The ... retained the remaining dimensions. I replaced it with a named roll_idx_array and omitted trailing indices, which NumPy retains automatically. This is equivalent, clearer, and has the same vectorized performance.

